PR #44, opened 2022-12-13
Add Test Infrastructure

### LINKED ISSUE #2: Test Infrastructure with PHPUnit
## Description

Setup test infrastructure with PHPUnit

## Acceptance Criteria

- Install the following packages with composer:
    - `"phpunit/phpunit"`
    - `"squizlabs/php_codesniffer"`
    - `"wp-coding-standards/wpcs"`
- Create a `phpunit.xml.dist` file to run tests from the `/tests` folder
- Create a `phpcs.xml.dist` file to run PHPCS tests
- Update `composer.json` with the following `scripts`:
   - `"lint": "phpcs --standard=phpcs.xml.dist"` - to run PHPCS
   - `"test": "phpunit -c phpunit.xml.dist --verbose"` - to run PHPUnit tests
- Install the following packages as dev dependencies with npm
   - `@wordpress/env`
   - `@wordpress/scripts`
- Update `package.json` with the following `scripts`:
   - `"wp-env": "wp-env"` 
   - `"lint-php": "wp-env run composer run-script lint"`
   - `"test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/performance/phpunit.xml.dist --verbose'"`
   - `"lint-js": "wp-scripts lint-js ./bin"`
- Create the following GitHub workflows in the `.github/workflows/` folder. (See [Performance Lab Plugin](https://github.com/WordPress/performance/tree/trunk/.github/workflows) for examples)
   - `js-lint.yml`
   - `php-lint.yml`
   - `php-test.yml`

jjgrainger:
Add initial test infrastructure and workflows.

- One difference compared to the Performance Lab plugin is I had to include a [`composer install` step in the `php-test.yml`](https://github.com/10up/plugin-check/pull/44/files#diff-acf3e0c6c881e3230ee20ea7499891845048d94362e17e496f06dafa7757f110R54-R55) workflow file. When not included, there is a `Error: The PHPUnit Polyfills library is a requirement for running the WP test suite.` ([see action logs](https://github.com/10up/plugin-check/actions/runs/3687471339/jobs/6241269167#step:7:22)) 

Closes #2


.github/workflows/php-test.yml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: shivammathur/setup-php@v2
+      - name: Setup Node.js (.nvmrc)
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+      - name: npm install
+        run: npm ci
+      - name: Composer Install
+        run: composer install --no-interaction --no-progress
```
mukeshpanchal27:
```suggestion
```

Remove this additional call and add `"wp-env run composer 'install --no-interaction'"` in package.json

package.json
```diff
@@ -5,7 +5,13 @@
...
   },
+  "devDependencies": {
+    "@wordpress/env": "^5.7.0",
+    "@wordpress/scripts": "^24.6.0"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "wp-env": "wp-env",
+    "lint-php": "wp-env run composer run-script lint",
+    "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/plugin-check/phpunit.xml.dist --verbose'"
```
mukeshpanchal27:
```suggestion
    "lint-php": "wp-env run composer run-script lint",
    "pretest-php": "wp-env run composer 'install --no-interaction'",
    "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/plugin-check/phpunit.xml.dist --verbose'"
```

mukeshpanchal27 (commented):
Thanks @jjgrainger Great start. The PR feedbacks address below mention issue.

> * One difference compared to the Performance Lab plugin is I had to include a [`composer install` step in the `php-test.yml`](https://github.com/10up/plugin-check/pull/44/files#diff-acf3e0c6c881e3230ee20ea7499891845048d94362e17e496f06dafa7757f110R54-R55) workflow file. When not included, there is a `Error: The PHPUnit Polyfills library is a requirement for running the WP test suite.` ([see action logs](https://github.com/10up/plugin-check/actions/runs/3687471339/jobs/6241269167#step:7:22)

In Performance Lab, It was introduce on https://github.com/WordPress/performance/pull/29/commits/5a8a5c220a3b66a5029939fe0e011e7e87e304bd commit.

.wp-env.json
```diff
@@ -0,0 +1,9 @@
+{
+	"core": null,
+	"plugins": [ "." ],
+	"env": {
+		"tests": {
+			"phpVersion": "7.4"
+		}
+	}
+}
```
felixarntz: I think the indentation here is incorrect, it needs to follow https://github.com/10up/plugin-check/blob/trunk/.editorconfig#L23.

package.json
```diff
@@ -5,7 +5,14 @@
...
+  "devDependencies": {
+    "@wordpress/env": "^5.7.0",
+    "@wordpress/scripts": "^24.6.0"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "wp-env": "wp-env",
+    "lint-php": "wp-env run composer run-script lint",
+    "pretest-php": "wp-env run composer 'install --no-interaction'",
+    "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/plugin-check/phpunit.xml.dist --verbose'"
```
felixarntz: Should we add a multisite configuration as well (see e.g. https://github.com/WordPress/performance/blob/trunk/package.json#L30)? I would think this is just as relevant here.

tests/basic-tests.php
```diff
@@ -0,0 +1,17 @@
...
+ * @package plugin-check
+ */
+
+class Basic_Tests extends WP_UnitTestCase {
+	/**
+	 * Basic test that asserts true.
+	 *
+	 * @test
+	 */
+	public function it_should_assert_true() {
```
felixarntz: Nit-pick to keep in mind for the future:
* WordPress core uses the `test_*` method prefix as a convention.
* At some point in Performance Lab we started doing this `it_should_*` convention, and in that plugin it is for now mixed between the two conventions. However I would advise we stick with the core approach.
* Sticking with that convention also means we can avoid the `@test` annotation which is redudant. Usually test methods don't need a doc-block, unless they use a `@dataProvider` or benefit from any extra explanation. Descriptive method names plus inline comments within the test method are usually sufficient to explain what's going on.

This isn't a big issue here since this is obviously just a dummy method, but wanted to point out for future reference. Since there are a few minor other changes to make, maybe let's also rename this to e.g. `test_true()` and remove the doc block?

felixarntz (commented):
@jjgrainger Left a few minor comments, mostly nit-picks. It would be good to add a multisite test setup though.

tests/basic-tests.php
mukeshpanchal27: Agreed. I have raise the issue in PL GH about it https://github.com/WordPress/performance/issues/514

.github/workflows/php-lint.yml
```diff
@@ -0,0 +1,58 @@
+name: PHP Code Linting
+
+on:
+  push:
+    branches:
+      - trunk
+      - 'release/**'
+    # Only run if PHP-related files changed.
```
mukeshpanchal27, line 8: Should be change as it run workflow when the other files (.yml, .json, .lock) also changes.

package.json
```diff
@@ -5,7 +5,14 @@
...
     "url": "git+https://github.com/10up/plugin-check.git"
   },
+  "devDependencies": {
+    "@wordpress/env": "^5.7.0",
+    "@wordpress/scripts": "^24.6.0"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "wp-env": "wp-env",
+    "lint-php": "wp-env run composer run-script lint",
```
mukeshpanchal27, line 15: Do we needs to add `prelint-php` similar to PL (See https://github.com/WordPress/performance/blob/trunk/package.json#L25)?

mukeshpanchal27 (commented):
Thanks @jjgrainger Left one nit-pick feedback and question.

.github/workflows/php-lint.yml
jjgrainger: @mukeshpanchal27 are you referring to the comment here?

If so, this matches the Performance Lab plugin and the .yml, .json, .lock files are related to PHP changes. We would wan't PHP lint/tests to run if those files change.

mukeshpanchal27: @jjgrainger My question is whether the workflow also runs when we change any files listed on `paths:`.

Check WordPress workflow comment https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/coding-standards.yml#L20-L33

If this makes sense, we will apply the same changes to the Performance Lab plugin.

mukeshpanchal27: It's not blocked at the moment. Will add comments for each file in the future if this change makes sense.

package.json
jjgrainger: @mukeshpanchal27 I've added this

mukeshpanchal27 (approved):
Great work @jjgrainger, LGTM.

felixarntz (approved):
@jjgrainger Thanks, looks great!

Issue #30, opened 2022-11-21
Create the WP-CLI subcommand wp plugin check


mehulkaklotar:
## Description

The plugin should extend the current `wp plugin` command, which will do single or multiple checks. The CLI command class file should be placed in the CLI directory under the includes directory. The class name should be `WordPress\Plugin_Check\CLI\Plugin_Check_Command`.

This issue is dependent on:
- #63 
- #14 

### Sub-Command
`check`
### Command arguments
`<plugin>`
The plugin to check. Plugin name.

`--checks=<checks>` 
Only runs checks provided as an argument in comma-separated values, e.g. enqueued-scripts, escaping. Otherwise runs all checks.

`--flag=<flags>` 
Limit the checks being executed according to their flags, e.g. stable, beta or experimental. Default is stable.

`--format=<format>` 
Format to display the results. Options are table, csv, and json. The default will be a table.

`--fields=<fields>`
Limit displayed results to a subset of fields provided.

`--ignore-warnings`
Limit displayed results to exclude warnings.

`--ignore-errors`
Limit displayed results to exclude errors.

### Examples
`wp plugin check akismet`
`wp plugin check akismet --check=escaping`
`wp plugin check akismet --format=json`

## Acceptance Criteria

- `Plugin_Check_Command` command class should be created and exists at `includes/CLI`
- Constructor should receive `Plugin_Context $plugin_context`
- Command should be registered using `WP_CLI::add_command( 'plugin check', __CLASS_NAMESPACE__ . '\register_command' );` in the `Plugin_Main` file.
- Provided `--checks` argument, command should process only these checks
- Provided `--flag` argument, command should process only flagged checks, default should be `stable`
- Provided `--format` argument, command should display results in that format
- Provided `--fields` argument, command should display results with these fields only
- Provided `--ignore-warnings` argument, command should display results with errors only
- Provided `--ignore-errors` argument, command should display results with warnings only
- The `Plugin_Check_Command` has a public `check( array $args, array $assoc_args )` method.
    - The method will attempt to resolve the plugin name passed in the `$args` array. If the plugin does not exist and a `WP_CLI::error()` is printed.
    - The `CLI_Runner` classes `run()` method is called. The `CLI_Runner` is instantiated and prepared earlier in the process so no parameters are needed to be passed.
    - The `run()` method is called in a try/catch block. If an exception is thrown by the method it is caught and printed as a `WP_CLI:error()`.
    - The `run()` method will return an instance of the `Check_Result` class.
[...]

Issue #5, opened 2022-11-21
Create Plugin_Context class 


mehulkaklotar:
## Description

The Plugin Context class provides helper methods to return additional information about the Plugin Check plugin. This includes information such as the basename, paths and URLs of the Plugin Check plugin.

This allows checks to use this information when preparing or performing tests. For example, a check may want to create test content before running tests. This test content may be stored as a file within the Plugin Check plugin, which can be easily referenced using the methods and data stored in the Plugin_Context class.

Plugin_Context class will be instantiated in the Plugin_Main class constructor and use the main file passed which will be the absolute path of the Plugin Check plugin’s main file.

## Acceptance Criteria

- The class file `Plugin_Context.php` exists within `includes`
- `$main_file` as the plugin's main file should be a protected property
- Constructor should receive the plugin's main file as a parameter and assign it to `$main_file`
- `basename()` method should return the plugin's name using `plugin_basename`
- `path()` method should return the absolute path of the plugin dir using `plugin_dir_path`
- `url()` method should return the plugin's URL using `plugin_dir_url`
- PHPUnit tests covering
    - `basename()` method returns string and name of the plugin
    - `path()` method returns string and path of the plugin
    - `url()` method returns string and URL of the plugin

PR #46, opened 2022-12-19
Create Plugin_Main class

### LINKED ISSUE #4: Create Plugin_Main class
## Description

The Plugin Main class will be used as the plugin’s most central class and will be used to bootstrap the Plugin Check plugin. Using a class allows all bootstrapping logic for the Plugin Check plugin to be consolidated in a single class, separating it from check logic and making it easier to maintain. This class will be responsible for setting up the plugin checker by registering hooks in WordPress and any other logic required, for example:

- Adding hooks and filters to register the plugin check WP-CLI command
- Enqueuing any assets required for the Plugin Check admin screen
- Localization support

`Plugin_Main` class will be instantiated in the `plugin-check.php` file of the Plugin Check plugin. It will accept a single `$main_file` parameter in its constructor which is used to create the plugin context. The context of the plugin will be handled by the protected property `$context` (`Plugin_Context`).

## Acceptance Criteria

- A new `Plugin_Main` class is created
- The class contains a single protected property named `$context`
- The class `__construct()` method accepts a single parameter
    - `(string) $main_file` - A string representing the absolute path to the plugin main file
    - The `$main_file` parameter is passed to a new instances of the `Plugin_Context` class
    - The new `Plugin_Context` instances is assigned to the protected `$context` property
- An empty `add_hooks` method exists in the `Plugin_Main` class to be updated later to register hooks to WordPress
- A new `Plugin_Context` class is created. This class will be empty and setup as a basic scaffold for #5 
- The `plugin-check.php` file is updated to include the following:
    - Constant `WP_PLUGIN_CHECK_VERSION ` defined with the plugin version
    - Constant `WP_PLUGIN_CHECK_MINIMUM_PHP ` defined with the minimum PHP version required for the plugin check plugin to work
    - A `wp_plugin_check_load` function exists which:
        - Checks the current PHP version is greater than the `WP_PLUGIN_CHECK_MINIMUM_PHP` constant
        - Checks the `vendor/autoload.php` file exists in the plugin
        - If the above checks pass
           - The `vendor/autoload.php` file is required in
           - The `Plugin_Main` class is instantiated passing `__FILE__` into the constructor
           - The `add_hooks()` method is called on the instantiated `Plugin_Main` object

jjgrainger:
Adds the `Plugin_Main` class.

- This PR contains commits from #44 which should be reviewed/merged first
- The `Plugin_Main` class accepts the plugin main file in the constructor
- The `Plugin_Main` class creates a `Plugin_Context` object passing the main file parameter
- The `Plugin_Context` class is a basic scaffold and will be addressed as part of #5 
- The `Plugin_Main` class has an empty `add_hooks` method that will be addressed as part of #30 

Closes #4


includes/Plugin_Context.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check;
+
+/**
+ * Class representing the context in which the plugin is running.
+ *
+ * @since 0.1.0
```
felixarntz: Maybe use `n.e.x.t` throughout, like we do in Performance Lab? Since we don't know what the first version will be.

plugin-check.php
```diff
@@ -14,3 +14,33 @@
  *
  * @package plugin-check
  */
+
+define( 'WP_PLUGIN_CHECK_VERSION', '0.1.0' );
```
felixarntz: See above.

plugin-check.php
```diff
@@ -14,3 +14,33 @@
...
+function wp_plugin_check_load() {
+	// Check for supported PHP version.
+	if ( version_compare( phpversion(), WP_PLUGIN_CHECK_MINIMUM_PHP, '<' ) ) {
+		return;
+	}
+
+	// Check Composer autoloader exists.
+	if ( ! file_exists( plugin_dir_path( __FILE__ ) . 'vendor/autoload.php' ) ) {
+		return;
+	}
```
felixarntz: Instead of silently cancelling here, maybe add an action to trigger an admin notice, to indicate to the end user what went wrong? See the POC https://github.com/felixarntz/plugin-check/blob/main/plugin-check.php#L26 for reference.

includes/Plugin_Main.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * Constructor. Set the plugin main file.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $main_file Absolute path of the plugin main file.
+	 * @return void
+	 */
+	public function __construct( $main_file ) {
+		$this->context = new Plugin_Context( $main_file );
```
felixarntz: Should we maybe add a public method `context()` or `get_context()` to be able to access this instance in a read-only way? That would also give us something a bit more appropriate (yet still simple) to unit test.

includes/Plugin_Main.php
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	protected $context;
+
+	/**
+	 * Constructor. Set the plugin main file.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $main_file Absolute path of the plugin main file.
+	 * @return void
```
felixarntz: `@return void` should be omitted, per WP doc standards.

includes/Plugin_Main.php
```diff
@@ -0,0 +1,48 @@
...
+	public function __construct( $main_file ) {
+		$this->context = new Plugin_Context( $main_file );
+	}
+
+	/**
+	 * Register WordPress hooks for the plugin.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
```
felixarntz: See above.

tests/Plugin_Main_Tests.php
```diff
@@ -0,0 +1,22 @@
...
+	 * Test class can be instantiated.
+	 *
+	 * @test
+	 */
+	public function it_can_be_instantiated() {
+		$main = new Plugin_Main( __DIR__ . '/../plugin-check.php' );
+
+		$this->assertIsObject( $main );
+		$this->assertInstanceOf( Plugin_Main::class, $main );
+	}
```
felixarntz, line 23: I know there's nothing else to test really, but this test is a bit pointless; I suggested to maybe add a method to get the context here and test that?

felixarntz (commented):
@jjgrainger I realize this include #44, but already left a few comments on the things that are new compared to the other PR.

tests/Plugin_Main_Tests.php
```diff
@@ -0,0 +1,22 @@
...
+
+use WordPress\Plugin_Check\Plugin_Main;
+
+class Plugin_Main_Tests extends WP_UnitTestCase {
+	/**
+	 * Test class can be instantiated.
+	 *
+	 * @test
+	 */
+	public function it_can_be_instantiated() {
```
felixarntz: See my comment on #44 for using the `test_` method prefix and avoiding doc-blocks (or more specifically the `@test` annotation) where not necessary.

includes/Plugin_Main.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Class WordPress\Plugin_Check\Plugin_Main
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check;
+
+/**
+ * Main class for the plugin
```
mukeshpanchal27:
```suggestion
 * Main class for the plugin.
```
As per [Inline document CS](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#2-classes), Use a period at the end.

includes/Plugin_Main.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return Plugin_Context
+	 */
+	public function context() {
+		return $this->context;
+	}
+
+	/**
+	 * Register WordPress hooks for the plugin.
```
mukeshpanchal27:
```suggestion
	 * Registers WordPress hooks for the plugin.
```

Use third-person singular verb.  See [Language](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#language)

plugin-check.php
```diff
@@ -14,3 +14,68 @@
...
+define( 'WP_PLUGIN_CHECK_MINIMUM_PHP', '5.6' );
+
+/**
+ * Checks basic requirements and loads the plugin.
+ *
+ * @since n.e.x.t
+ */
+function wp_plugin_check_load() {
+	// Check for supported PHP version.
+	if ( version_compare( phpversion(), WP_PLUGIN_CHECK_MINIMUM_PHP, '<' ) ) {
```
mukeshpanchal27, line 29: Why `phpversion()`? Is there any specific reason? The constant `PHP_VERSION` is used by WordPress.

plugin-check.php
```diff
@@ -14,3 +14,68 @@
...
+	}
+
+	// Check Composer autoloader exists.
+	if ( ! file_exists( plugin_dir_path( __FILE__ ) . 'vendor/autoload.php' ) ) {
+		add_action( 'admin_notices', 'wp_plugin_check_display_composer_autoload_notice' );
+		return;
+	}
+
+	// Load the Composer autoloader.
+	require_once plugin_dir_path( __FILE__ ) . 'vendor/autoload.php';
```
mukeshpanchal27: Similar to Performance Lab lets define some constant that we can use in other places.

https://github.com/WordPress/performance/blob/trunk/load.php#L19-L20

mukeshpanchal27 (commented):
Thanks @jjgrainger Left some feedback.

Question: Is there any reason to use the first capital in the class file name? If not, then we can use lowercase.

plugin-check.php
jjgrainger: @mukeshpanchal27 I think they both work the same, so I have left it as using `phpversion()`

tests/Plugin_Main_Tests.php
jjgrainger: Thanks @felixarntz I've added a `context()` method to test here

tests/Plugin_Main_Tests.php
```diff
@@ -0,0 +1,24 @@
...
+	public function set_up() {
+		parent::set_up();
+
+		$this->plugin_main = new Plugin_Main( 'plugin-check/plugin-check.php' );
+	}
+
+	public function test_context() {
+		$context = $this->plugin_main->context();
+
+		$this->assertIsObject( $context );
```
felixarntz, line 21: Nit-pick, this is unnecessary since the assertion below by definition covers that already.

felixarntz (approved):
@jjgrainger LGTM, just one nit-pick.

mukeshpanchal27 (approved):
@jjgrainger looks great.

PR #47, opened 2022-12-19
Create Plugin_Context class

### LINKED ISSUE #5: Create Plugin_Context class

jjgrainger:
Adds the `Plugin_Context` class

- Addresses acceptance criteria in #5 

Closes #5

### DIFF
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -10,14 +10,14 @@ namespace WordPress\Plugin_Check;
 /**
  * Class representing the context in which the plugin is running.
  *
- * @since 0.1.0
+ * @since n.e.x.t
  */
 class Plugin_Context {
 
 	/**
 	 * Absolute path of the plugin main file.
 	 *
-	 * @since 0.1.0
+	 * @since n.e.x.t
 	 * @var string
 	 */
 	protected $main_file;
@@ -25,7 +25,7 @@ class Plugin_Context {
 	/**
 	 * Constructor.
 	 *
-	 * @since 0.1.0
+	 * @since n.e.x.t
 	 *
 	 * @param string $main_file The absolute path to the plugin main file.
 	 */
@@ -36,7 +36,7 @@ class Plugin_Context {
 	/**
 	 * Returns the plugin basename.
 	 *
-	 * @since 0.1.0
+	 * @since n.e.x.t
 	 *
 	 * @return string Plugin basename.
 	 */
@@ -47,7 +47,7 @@ class Plugin_Context {
 	/**
 	 * Returns the absolute path for a relative path to the plugin directory.
 	 *
-	 * @since 0.1.0
+	 * @since n.e.x.t
 	 *
 	 * @param string $relative_path Optional. Relative path. Default '/'.
 	 * @return string Absolute path.
@@ -59,7 +59,7 @@ class Plugin_Context {
 	/**
 	 * Returns the full URL for a pth relative to the plugin directory.
 	 *
-	 * @since 0.1.0
+	 * @since n.e.x.t
 	 *
 	 * @param string $relative_path Optional. Relative path. Default '/'.
 	 * @return string Full URL.
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -53,7 +53,7 @@ class Plugin_Context {
 	 * @return string Absolute path.
 	 */
 	public function path( $relative_path = '/' ) {
-		return plugin_dir_path( $this->main_file ) . ltrim( $relative_path . '/' );
+		return plugin_dir_path( $this->main_file ) . ltrim( $relative_path, '/' );
 	}
 
 	/**
@@ -65,6 +65,6 @@ class Plugin_Context {
 	 * @return string Full URL.
 	 */
 	public function url( $relative_path = '/' ) {
-		return plugin_dir_url( $this->main_file ) . \ltrim( $relative_path . '/' );
+		return plugin_dir_url( $this->main_file ) . ltrim( $relative_path, '/' );
 	}
 }
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -57,7 +57,7 @@ class Plugin_Context {
 	}
 
 	/**
-	 * Returns the full URL for a pth relative to the plugin directory.
+	 * Returns the full URL for a path relative to the plugin directory.
 	 *
 	 * @since n.e.x.t
 	 *
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -13,5 +13,58 @@ namespace WordPress\Plugin_Check;
  * @since 0.1.0
  */
 class Plugin_Context {
-	// @TODO: To complete as part of issue #5
+
+	/**
+	 * Absolute path of the plugin main file.
+	 *
+	 * @since 0.1.0
+	 * @var string
+	 */
+	protected $main_file;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $main_file The absolute path to the plugin main file.
+	 */
+	public function __construct( $main_file ) {
+		$this->main_file = $main_file;
+	}
+
+	/**
+	 * Returns the plugin basename.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return string Plugin basename.
+	 */
+	public function basename() {
+		return plugin_basename( $this->main_file );
+	}
+
+	/**
+	 * Returns the absolute path for a relative path to the plugin directory.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $relative_path Optional. Relative path. Default '/'.
+	 * @return string Absolute path.
+	 */
+	public function path( $relative_path = '/' ) {
+		return plugin_dir_path( $this->main_file ) . ltrim( $relative_path . '/' );
+	}
+
+	/**
+	 * Returns the full URL for a pth relative to the plugin directory.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $relative_path Optional. Relative path. Default '/'.
+	 * @return string Full URL.
+	 */
+	public function url( $relative_path = '/' ) {
+		return plugin_dir_url( $this->main_file ) . \ltrim( $relative_path . '/' );
+	}
 }
--- a/tests/Plugin_Context_Tests.php
+++ b/tests/Plugin_Context_Tests.php
@@ -15,7 +15,7 @@ class Plugin_Context_Tests extends WP_UnitTestCase {
 	}
 
 	public function test_basename() {
-		$this->assertSame('plugin-check/plugin-check.php', $this->plugin_context->basename() );
+		$this->assertSame( 'plugin-check/plugin-check.php', $this->plugin_context->basename() );
 	}
 
 	public function test_path() {
--- a/tests/Plugin_Context_Tests.php
+++ b/tests/Plugin_Context_Tests.php
@@ -8,80 +8,39 @@
 use WordPress\Plugin_Check\Plugin_Context;
 
 class Plugin_Context_Tests extends WP_UnitTestCase {
-
-	/**
-	 * Create a single Plugin_Context to test.
-	 */
 	public function set_up() {
 		parent::set_up();
 
 		$this->plugin_context = new Plugin_Context( 'plugin-check/plugin-check.php' );
 	}
 
-	/**
-	 * Test class can be instantiated.
-	 *
-	 * @test
-	 */
-	public function it_can_be_instantiated() {
-		$main = new Plugin_Context( 'plugin-check/plugin-check.php' );
-
-		$this->assertIsObject( $main );
-		$this->assertInstanceOf( Plugin_Context::class, $main );
-	}
-
-	/**
-	 * Test basename method.
-	 *
-	 * @test
-	 */
-	public function it_can_return_the_basename() {
+	public function test_basename() {
 		$basename = $this->plugin_context->basename();
 
 		$this->assertSame( $basename, 'plugin-check/plugin-check.php' );
 	}
 
-	/**
-	 * Test path method.
-	 *
-	 * @test
-	 */
-	public function it_can_return_the_path() {
-		$basename = $this->plugin_context->path();
+	public function test_path() {
+		$path = $this->plugin_context->path();
 
-		$this->assertSame( $basename, 'plugin-check/' );
+		$this->assertSame( $path, 'plugin-check/' );
 	}
 
-	/**
-	 * Test path method with parameter.
-	 *
-	 * @test
-	 */
-	public function it_can_return_the_path_with_parameter() {
-		$basename = $this->plugin_context->path( '/another/folder' );
+	public function test_path_with_parameter() {
+		$path = $this->plugin_context->path( '/another/folder' );
 
-		$this->assertSame( $basename, 'plugin-check/another/folder' );
+		$this->assertSame( $path, 'plugin-check/another/folder' );
 	}
 
-	/**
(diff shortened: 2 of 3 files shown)


includes/Plugin_Context.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check;
+
+/**
+ * Class representing the context in which the plugin is running.
+ *
+ * @since 0.1.0
```
felixarntz: See my comments on #46, let's use `n.e.x.t`.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Test class can be instantiated.
+	 *
+	 * @test
+	 */
+	public function it_can_be_instantiated() {
+		$main = new Plugin_Context( 'plugin-check/plugin-check.php' );
+
+		$this->assertIsObject( $main );
+		$this->assertInstanceOf( Plugin_Context::class, $main );
+	}
```
felixarntz, line 20: See my comment on #46, I don't think we need a test for instantiation, unless that instantiation does something extra that's worth testing.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+
+	/**
+	 * Test basename method.
+	 *
+	 * @test
+	 */
+	public function it_can_return_the_basename() {
```
felixarntz: See my comments on #44 and #46 regarding usage of `test_` prefix (applies throughout this class). The convention is to just call the method like the tested method (i.e. here `test_basename()`), unless there are multiple test methods for the same method in which case the method should be more descriptive (e.g. `test_basename_returns_basename()`, though here that is a pointless example).

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+
+		$this->assertSame( $basename, 'plugin-check/plugin-check.php' );
+	}
+
+	/**
+	 * Test path method.
+	 *
+	 * @test
+	 */
+	public function it_can_return_the_path() {
```
felixarntz: See above, this could be `test_path()`.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+
+		$this->assertSame( $basename, 'plugin-check/' );
+	}
+
+	/**
+	 * Test path method with parameter.
+	 *
+	 * @test
+	 */
+	public function it_can_return_the_path_with_parameter() {
```
felixarntz: And this `test_path_with_parameter()`.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+		$this->assertSame( $basename, 'plugin-check/plugin-check.php' );
+	}
+
+	/**
+	 * Test path method.
+	 *
+	 * @test
+	 */
+	public function it_can_return_the_path() {
+		$basename = $this->plugin_context->path();
```
felixarntz: Variable name here and below is misleading, likely copy-paste oversight :)

felixarntz (commented):
@jjgrainger Left a few comments on the code here that isn't in #44 or #46 (mostly similar feedback though).

includes/Plugin_Context.php
```diff
@@ -0,0 +1,70 @@
...
+	 *
+	 * @param string $relative_path Optional. Relative path. Default '/'.
+	 * @return string Absolute path.
+	 */
+	public function path( $relative_path = '/' ) {
+		return plugin_dir_path( $this->main_file ) . ltrim( $relative_path, '/' );
+	}
+
+	/**
+	 * Returns the full URL for a pth relative to the plugin directory.
```
```suggestion
	 * Returns the full URL for a path relative to the plugin directory.
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,46 @@
...
+	public function set_up() {
+		parent::set_up();
+
+		$this->plugin_context = new Plugin_Context( 'plugin-check/plugin-check.php' );
+	}
+
+	public function test_basename() {
+		$basename = $this->plugin_context->basename();
+
+		$this->assertSame( $basename, 'plugin-check/plugin-check.php' );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( 'plugin-check/plugin-check.php', $this->plugin_context->basename() );
```

As per the [assertSame documentation](https://phpunit.readthedocs.io/en/9.5/assertions.html#assertsame) the expected should be in first place. Also remove variable.

Do similar change for other tests.

mukeshpanchal27 (commented):
Thanks @jjgrainger, Left nit-pick feedback.

mukeshpanchal27 (approved):
Thanks @jjgrainger, LGTM!

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+		$this->assertSame( 'plugin-check/plugin-check.php', $this->plugin_context->basename() );
+	}
+
+	public function test_path() {
+		$this->assertSame( 'plugin-check/', $this->plugin_context->path() );
+	}
+
+	public function test_path_with_parameter() {
+		$this->assertSame( 'plugin-check/another/folder', $this->plugin_context->path( '/another/folder' ) );
+	}
```
felixarntz: To make these two checks more stable, I would suggest not hard-coding an assumed path here. It'd be more reliable to use `dirname( __DIR__ )` instead of `plugin-check`.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+
+	public function test_url() {
+		$this->assertSame( site_url( '/wp-content/plugins/plugin-check/' ), $this->plugin_context->url() );
+	}
+
+	public function test_url_with_parameter() {
+		$this->assertSame( site_url( '/wp-content/plugins/plugin-check/folder/file.css' ), $this->plugin_context->url( '/folder/file.css' ) );
+	}
```
felixarntz: Same here, how about using `WP_PLUGIN_DIR . '/' . basename( dirname( __DIR__ ) )` instead of `site_url( '/wp-content/plugins/plugin-check' )`?

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+class Plugin_Context_Tests extends WP_UnitTestCase {
+	public function set_up() {
+		parent::set_up();
+
+		$this->plugin_context = new Plugin_Context( 'plugin-check/plugin-check.php' );
+	}
+
+	public function test_basename() {
+		$this->assertSame( 'plugin-check/plugin-check.php', $this->plugin_context->basename() );
+	}
```
felixarntz: Technically, even here we could make it more reliable by using `basename( dirname( __DIR__ ) )` instead of the `plugin-check` directory. We know for sure the main file is called `plugin-check.php`, but the directory name depends on the WordPress setup and where the plugin is installed to.

felixarntz (approved):
@jjgrainger Overall LGTM, however I have a few suggestions for making the tests more reliable, as they currently use hard-coded strings which I suspect will fail depending on how the test environment is set up. (They work here, but they may not work if someone sets up tests in another way, e.g. without Docker.)

tests/Plugin_Context_Tests.php
mukeshpanchal27: @felixarntz Here we have to use `WP_PLUGIN_URL` instead of `WP_PLUGIN_DIR`

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+ * @package plugin-check
+ */
+
+use WordPress\Plugin_Check\Plugin_Context;
+
+class Plugin_Context_Tests extends WP_UnitTestCase {
+	public function set_up() {
+		parent::set_up();
+
+		$this->plugin_context = new Plugin_Context( 'plugin-check/plugin-check.php' );
```
mukeshpanchal27: Can we apply https://github.com/10up/plugin-check/pull/47#discussion_r1054873951 changes here also?

mukeshpanchal27:
```suggestion
		$this->plugin_name    = basename( dirname( __DIR__ ) );
		$this->plugin_context = new Plugin_Context( $this->plugin_name . '/plugin-check.php' );
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function test_basename() {
+		$this->assertSame( 'plugin-check/plugin-check.php', $this->plugin_context->basename() );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( $this->plugin_name . '/plugin-check.php', $this->plugin_context->basename() );
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function test_path() {
+		$this->assertSame( 'plugin-check/', $this->plugin_context->path() );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( $this->plugin_name . '/', $this->plugin_context->path() );
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function test_path_with_parameter() {
+		$this->assertSame( 'plugin-check/another/folder', $this->plugin_context->path( '/another/folder' ) );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( $this->plugin_name . '/another/folder', $this->plugin_context->path( '/another/folder' ) );
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function test_url() {
+		$this->assertSame( site_url( '/wp-content/plugins/plugin-check/' ), $this->plugin_context->url() );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( WP_PLUGIN_URL . '/' . $this->plugin_name . '/', $this->plugin_context->url() );
```

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function test_url_with_parameter() {
+		$this->assertSame( site_url( '/wp-content/plugins/plugin-check/folder/file.css' ), $this->plugin_context->url( '/folder/file.css' ) );
```
mukeshpanchal27:
```suggestion
		$this->assertSame( WP_PLUGIN_URL . '/' . $this->plugin_name . '/folder/file.css', $this->plugin_context->url( '/folder/file.css' ) );
```

tests/Plugin_Main_Tests.php
```diff
@@ -18,7 +18,6 @@ public function set_up() {
 	public function test_context() {
 		$context = $this->plugin_main->context();
```
mukeshpanchal27:
```suggestion
```

Remove the variable.

tests/Plugin_Main_Tests.php
```diff
@@ -18,7 +18,6 @@ public function set_up() {
...
 
-		$this->assertIsObject( $context );
 		$this->assertInstanceOf( Plugin_Context::class, $context );
```
mukeshpanchal27:
```suggestion
		$this->assertInstanceOf( Plugin_Context::class, $this->plugin_main->context() );
```

mukeshpanchal27 (approved):
@jjgrainger Over to you for review of the changes and merging.

tests/Plugin_Context_Tests.php
```diff
@@ -0,0 +1,37 @@
...
+		$this->assertSame( $this->plugin_name . '/plugin-check.php', $this->plugin_context->basename() );
+	}
+
+	public function test_path() {
+		$this->assertSame( $this->plugin_name . '/', $this->plugin_context->path() );
+	}
+
+	public function test_path_with_parameter() {
+		$this->assertSame( $this->plugin_name . '/another/folder', $this->plugin_context->path( '/another/folder' ) );
+	}
```
felixarntz, line 28: @jjgrainger Just spotting this post-merge: This still doesn't fully address my original feedback: I would think we need to use `WP_PLUGIN_DIR . '/' . $this->plugin_name` instead of just `$this->plugin_name`. Have you tried that, were there any problems? At least in theory this still looks unstable to me without that bit.

Let me know what you think. If my concerns are valid, we can fix this throughout in a quick follow up PR.

cc @mukeshpanchal27

mukeshpanchal27: The `plugin_dir_path( $this->main_file )` return plugin directory slug instead of the full relative path.

includes/Plugin_Context.php
```diff
@@ -13,5 +13,58 @@
...
+	/**
+	 * Returns the absolute path for a relative path to the plugin directory.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $relative_path Optional. Relative path. Default '/'.
+	 * @return string Absolute path.
+	 */
+	public function path( $relative_path = '/' ) {
+		return plugin_dir_path( $this->main_file ) . ltrim( $relative_path, '/' );
```
mukeshpanchal27, line 56: Should update this line with `return plugin_dir_path( WP_PLUGIN_CHECK_PLUGIN_DIR_PATH ) . $this->main_file . ltrim( $relative_path, '/' );` to get full relative path. If we use this and run the unit test it gives `Use of undefined constant WP_PLUGIN_CHECK_PLUGIN_DIR_PATH - assumed 'WP_PLUGIN_CHECK_PLUGIN_DIR_PATH' (this will throw an Error in a future version of PHP)` notice that needs to figure out.

@jjgrainger I'd like to get your feedback on this.

tests/Plugin_Context_Tests.php
jjgrainger: Thanks both, I've addressed these comments in https://github.com/10up/plugin-check/pull/51

includes/Plugin_Context.php
jjgrainger: Thanks @mukeshpanchal27 

I've addressed this in https://github.com/10up/plugin-check/pull/51. We're unable to use the `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH` as it's created in the `plugin-check.php` file which is not bootstrapped as part of the tests (nor should it be).

The line here works correctly as the full path is expected to be passed in the constructor (see the [`plugin-check.php file here](https://github.com/10up/plugin-check/blob/trunk/plugin-check.php#L45)).

I've updated the tests based on the feedback in #51 which I believe to be correct.

PR #50, opened 2022-12-22
Create Check_Result class

### LINKED ISSUE #11: Create Check_Result class
## Description

The Check_Result class stores all errors, warnings and other messages generated by Check classes. This allows all messages to be recorded to a single repository in a compatible format and consistent API. Once the run of all checks is complete, the Check_Result class will be used to retrieve all results so that they can be presented in a format best suited for the environment (WP-CLI, WP Admin, Continuous Integration Pipeline) in which the checks were run.

The Check_Result class constructor will receive an instance of the Check_Context as an argument. This is so that additional information can be added to each check message such as the file name, path, line, and violation code (a code that is unique to the check/error/warning and can be empty) where the check message occurs. The violation code will be determined by the check itself.

Additionally, this class will contain methods and properties to track the count of errors and warnings to provide an overview of the final results.

## Acceptance Criteria
- The class file `Check_Result.php` should exists at `includes/Checker`
- Protected property `$check_context` should be instance of `Check_Context` class created here https://github.com/10up/plugin-check/issues/10
- Protected property `$errors` and `$warnings` should be declared as array
- Protected property `$error_count` and `$warning_count` should be declared as int
- Constructor should receive `Check_Context` class instance as an argument
- Class should have `add_message` method which will have `$error` or `$warning` as an argument as a violation code for the check with file name, path, line as an argument 
- Class should have `get_errors` method to get list of all errors which will return `$errors`
- Class should have `get_warnings` method to get list of all errors which will return `$warnings`
- Class should have `get_error_count` method to get count of errors which will return `$error_count`
- Class should have `get_warning_count` method to get count of warnings which will return `$warning_count`
- PHPUnit tests covering
    - `add_message` method adds the message with arguments to the error or warning lists and increments the counts
    - `get_errors` method returns all errors expected
    - `get_warnings` method returns all warnings expected
    - `get_error_count` method returns expected error count
    - `get_warning_count` method returns expected warning count

jjgrainger:
Adds the Check_Result class

- Addresses acceptance criteria in #11 

Closes #11

### DIFF
--- a/includes/Checker/Check_Result.php
+++ b/includes/Checker/Check_Result.php
@@ -1,0 +1,210 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Check_Result
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+use WordPress\Plugin_Check\Plugin_Context;
+use WordPress\Plugin_Check\Checker\Check_Context;
+
+/**
+ * Result for running checks on a plugin.
+ *
+ * @since n.e.x.t
+ */
+class Check_Result {
+
+	/**
+	 * Main context instance.
+	 *
+	 * @since n.e.x.t
+	 * @var Plugin_Context
+	 */
+	protected $main_context;
+
+	/**
+	 * Context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 * @var Check_Context
+	 */
+	protected $check_context;
+
+	/**
+	 * List of errors.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	protected $errors = array();
+
+	/**
+	 * List of warnings.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	protected $warnings = array();
+
+	/**
+	 * Number of errors.
+	 *
+	 * @since n.e.x.t
+	 * @var int
+	 */
+	protected $error_count = 0;
+
+	/**
+	 * Number of warnings.
+	 *
+	 * @since n.e.x.t
+	 * @var int
+	 */
+	protected $warning_count = 0;
+
+	/**
+	 * Sets the context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Plugin_Context $main_context  Main context instance.
+	 * @param Check_Context  $check_context Check context instance for the plugin.
+	 */
+	public function __construct( Plugin_Context $main_context, Check_Context $check_context ) {
+		$this->main_context  = $main_context;
+		$this->check_context = $check_context;
+	}
+
+	/**
+	 * Returns the main context for the plugin checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Plugin_Context Main context instance.
+	 */
+	public function context() {
+		return $this->main_context;
+	}
+
+	/**
+	 * Returns the context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context Plugin context instance.
+	 */
+	public function plugin() {
+		return $this->check_context;
+	}
+
+	/**
+	 * Adds an error or warning to the respective stack.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool   $error   Whether it is an error message.
+	 * @param string $message The message.
+	 * @param array  $args    {
+	 *     Additional message arguments.
+	 *
+	 *     @type string $code   Violation code according to the message. Default empty string.
+	 *     @type string $file   The file in which the message occurred. Default empty string (unknown file).
+	 *     @type int    $line   The line on which the message occurred. Default 0 (unknown line).
+	 *     @type int    $column The column on which the message occurred. Default 0 (unknown column).
+	 * }
+	 */
+	public function add_message( $error, $message, $args = array() ) {
+		$defaults = array(
+			'code'   => '',
+			'file'   => '',
+			'line'   => 0,
+			'column' => 0,
+		);
+
+		$data = array_merge(
+			array(
+				'message' => $message,
+			),
+			$defaults,
+			array_intersect_key( $args, $defaults )
+		);
+
+		$file   = str_replace( $this->check_context->path( '/' ), '', $data['file'] );
+		$line   = $data['line'];
+		$column = $data['column'];
+		unset( $data['line'], $data['column'] );
+
+		if ( $error ) {
+			if ( ! isset( $this->errors[ $file ] ) ) {
+				$this->errors[ $file ] = array();
+			}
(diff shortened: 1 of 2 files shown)


includes/Checker/Check_Result.php
```diff
@@ -7,11 +7,204 @@
...
 class Check_Result {
-	// @TODO: Complete as part of issue #11.
+
+	/**
+	 * Main context instance.
+	 *
+	 * @since n.e.x.t
+	 * @var Plugin_Context
+	 */
+	protected $main_context;
```
felixarntz: I realize I had this property (and related `context()` method) in this class in my POC https://github.com/felixarntz/plugin-check/blob/main/includes/Checker/Check_Result.php, however that was a POC and I don't think having that main context available in here is a clean approach. Since this class represents a check result, it should only contain data relevant to the checked plugin and its check results.

The `$main_context` property isn't used by this class at all, and putting it in here only to access it from the outside doesn't look like a clean separation of concerns to me, as mentioned above.

I think we can think about how to provide access to the main context once we need it somewhere.

includes/Checker/Check_Result.php
```diff
@@ -7,11 +7,204 @@
...
+	/**
+	 * Sets the context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Plugin_Context $main_context  Main context instance.
+	 * @param Check_Context  $check_context Check context instance for the plugin.
+	 */
+	public function __construct( Plugin_Context $main_context, Check_Context $check_context ) {
+		$this->main_context  = $main_context;
```
felixarntz: See above, we don't need the first parameter here.

includes/Checker/Check_Result.php
```diff
@@ -7,11 +7,204 @@
...
+	/**
+	 * Returns the main context for the plugin checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Plugin_Context Main context instance.
+	 */
+	public function context() {
+		return $this->main_context;
+	}
```
felixarntz: See above, this shouldn't be here IMO.

includes/Checker/Check_Result.php
```diff
@@ -7,11 +7,204 @@
...
+				'message' => $message,
+			),
+			$defaults,
+			array_intersect_key( $args, $defaults )
+		);
+
+		$file   = str_replace( $this->check_context->path( '/' ), '', $data['file'] );
+		$line   = $data['line'];
+		$column = $data['column'];
+		unset( $data['line'], $data['column'] );
```
felixarntz: Should we unset `$data['file']` here too (since it's already in the structure below, like line and column are)?

includes/Checker/Check_Result.php
```diff
@@ -7,11 +7,204 @@
 
 namespace WordPress\Plugin_Check\Checker;
 
+use WordPress\Plugin_Check\Plugin_Context;
```
felixarntz: See other comments, we can remove this reference then.

felixarntz (changes requested):
@jjgrainger Left a few comments, mostly LGTM.

tests/Checker/Check_Result_Tests.php
```diff
@@ -0,0 +1,201 @@
...
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+class Check_Result_Tests extends WP_UnitTestCase {
+	public function set_up() {
+		parent::set_up();
+
+		$check_context = new Check_Context( 'test-plugin/test-plugin.php' );
+
+		$this->check_result = new Check_Result( $check_context );
```
felixarntz, line 24: I think this needs to get a class property added still, at least as a best practice.

tests/Checker/Check_Result_Tests.php
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	public function test_plugin() {
+		$this->assertInstanceOf( Check_Context::class, $this->check_result->plugin() );
```
felixarntz, line 28: Maybe also check that the context has the expected basename based on what is passed in the `set_up` method?

tests/Checker/Check_Result_Tests.php
```diff
@@ -0,0 +1,201 @@
...
+		$this->assertArrayHasKey( 40, $warnings['test-plugin.php'][12] );
+
+		// Tests the column array contains the message details.
+		$expected = array(
+			'message' => 'Warning message',
+			'code'    => 'test_warning',
+		);
+
+		$this->assertEquals( $expected, $warnings['test-plugin.php'][12][40][0] );
+	}
```
felixarntz: Isn't this test pretty much the same as the one below? In that case let's remove it.

felixarntz (commented):
@jjgrainger Production code LGTM, left some feedback on the tests.

tests/Checker/Check_Result_Tests.php
jjgrainger: Thanks @felixarntz I've added a property to the test class

tests/Checker/Check_Result_Tests.php
jjgrainger: Thanks @felixarntz I've added an additional assertion for the `Check_Context` `basename()`

tests/Checker/Check_Result_Tests.php
jjgrainger: Thanks @felixarntz this was very similar so I have removed this test

tests/Checker/Check_Result_Tests.php
```diff
@@ -0,0 +1,179 @@
...
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+class Check_Result_Tests extends WP_UnitTestCase {
+	/**
+	 * Check_Result instance.
+	 *
+	 * @var Check_Result
+	 */
+	protected $check_result;
```
felixarntz, line 17: FYI no need to do any formal doc blocks in tests. It's good to add comments when a test method does things that may not be immediately clear on why it does so, but generally tests don't need docs.

Obviously not a blocker, it's okay to have this. But not needed :)

felixarntz (approved):
@jjgrainger Thanks, LGTM!

mukeshpanchal27 (approved):
Thanks @jjgrainger The PR looks solid to me. I have only one suggestion regarding the count function.

Can we use the [count](https://www.php.net/manual/en/function.count.php) function for `errors` and `warnings`? If yes, then it removes two variables from the class and the increment logic for those variables.

Ie.
```
public function get_error_count() {
    return count( $this->errors );
}
```

cc. @felixarntz

jjgrainger:
Thanks @mukeshpanchal27 

The structure of the errors and warnings arrays mean we cannot use the `count()` function and why we track the warning/errors count manually.

PR #51, opened 2023-01-04
Fix context class test stability

jjgrainger:
Addresses comments in https://github.com/10up/plugin-check/pull/47#discussion_r1055676303

- Makes use of the `TESTS_PLUGIN_DIR` constant when creating instances of the `Plugin_Context` and `Check_Context` classes in unit tests. `TESTS_PLUGIN_DIR` is equal to `/var/www/html/wp-content/plugins/plugin-check/`.
- This is similar to how the classes are instantiated using the `__FILE__` constant (seen in [`plugin-check.php` file](https://github.com/10up/plugin-check/blob/trunk/plugin-check.php#L45) and creates the same expected full path, e.g `/var/www/html/wp-content/plugins/plugin-check/`
- Tests are then updated to use the full path by prefixing expected strings with `WP_PLUGIN_DIR . '\'`

### DIFF
--- a/tests/Checker/Check_Context_Tests.php
+++ b/tests/Checker/Check_Context_Tests.php
@@ -11,8 +11,8 @@ class Check_Context_Tests extends WP_UnitTestCase {
 	public function set_up() {
 		parent::set_up();
 
-		$this->plugin_name   = basename( dirname( __DIR__ ) );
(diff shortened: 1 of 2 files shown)


felixarntz (approved):
@jjgrainger Looks good!

mukeshpanchal27 (approved):
Thanks @jjgrainger, Great work.

PR #53, opened 2023-01-06
add Abstract_PHP_CodeSniffer_Check

### LINKED ISSUE #8: Create abstract class PHP_CodeSniffer_Check
This abstract class will be used to create concrete check classes that perform static analysis. This class will provide foundational methods that allow concrete classes to define code standards and sniffs to test the plugin code against. It will also contain an instance of `PHP_CodeSniffer\Runner` to run the tests and return the results to be processed.

## Acceptance Criteria

- An abstract `Abstract_PHP_CodeSniffer_Check` class exists
- The `Abstract_PHP_CodeSniffer_Check` implements the `Check` interface
- The class contains a `public function run( Check_Result $result )` method
- The `run()` method accepts a single `$result` parameter
- The `$result` parameter is an instance of the `Check_Result` class
- The `run()` method handles the logic to
    - Require in the PHPCS autoloader from composer vendor folder
    - Backup the original `$_SERVER['argv']` array to a `$original_cmd_args` variable
    - Generate the arguments to pass to PHPCS by passing `$result->plugin()` to `Abstract_PHP_CodeSniffer_Check::get_args()` method
    - Set the `$_SERVER['argv']` array to the arguments to pass to PHPCS
    - Run PHPCS using an instance of `\PHP_CodeSniffer\Runner` class
    - Restore the original command arguments from `$original_cmd_args` variable
    - Add the results from PHPCS to the `$results` object using the `$results->add_message()` method
- The `Abstract_PHP_CodeSniffer_Check` contains a `abstract function get_args( Check_Context $context )` method
- The `get_args()` method accepts a `$context` parameter
- The `$context` parameter is an instance of `Check_Context`
- The `get_args()` method will be written in concrete Check classes to return the arguments to pass to PHPCS

jjgrainger:
Adds the Abstract_PHP_CodeSniffer_Check class

- Addresses acceptance criteria in #8 

Closes #8

### DIFF
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Checks\PHP_CodeSniffer_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use WordPress\Plugin_Check\Checker\Check;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use Exception;
+
+/**
+ * Check for running one or more PHP CodeSniffer sniffs.
+ *
+ * @since 1.0.0
+ */
+abstract class Abstract_PHP_CodeSniffer_Check implements Check {
+
+	/**
+	 * List of allowed PHPCS arguments.
+	 *
+	 * @var array
+	 */
+	protected $allowed_args = array(
+		'extensions',
+		'standard',
+		'sniffs',
+		'exclude',
+	);
+
+	/**
+	 * Returns an associative array of arguements to pass to PHPCS.
+	 *
+	 * @return array
+	 */
+	abstract public function get_args();
+
+	/**
+	 * Creates the command arguments.
+	 *
+	 * @param Check_Result $result The check results including the plugin context to check.
+	 *
+	 * @return array
+	 */
+	private function create_args( Check_Result $result ) {
+		// Set the default arguments for PHPCS.
+		$argv = array(
+			'',
+			$result->plugin()->path( '' ),
+			'--report=Json',
+			'--report-width=9999',
+		);
+
+		// Only accept allowed PHPCS arguments from check arguments array.
+		$check_args = array_filter(
+			$this->get_args(),
+			function( $key ) {
+				return in_array( $key, $this->allowed_args, true );
+			},
+			ARRAY_FILTER_USE_KEY
+		);
+
+		// Format check arguments for PHPCS.
+		foreach ( $check_args as $key => $value ) {
+			$argv[] = "--{$key}=$value";
+		}
+
+		return $argv;
+	}
+
+	/**
+	 * Amends the given result by running the check on the associated plugin.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	public function run( Check_Result $result ) {
+
+		// Backup the original command line arguments.
+		$orig_cmd_args = $_SERVER['argv'];
+
+		// Set the check arguments for PHPCS.
+		$_SERVER['argv'] = $this->create_args( $result );
+
+		// Run PHPCS.
+		try {
+			ob_start();
+			$runner = new \PHP_CodeSniffer\Runner();
+			$runner->runPHPCS();
+			$reports = ob_get_clean();
+		} catch ( Exception $e ) {
+			$_SERVER['argv'] = $orig_cmd_args;
+			throw $e;
+		}
+
+		// Parse the reports into data to add to the overall $result.
+		$reports = json_decode( trim( $reports ), true );
+
+		if ( empty( $reports['files'] ) ) {
+			return;
+		}
+
+		foreach ( $reports['files'] as $file_name => $file_results ) {
+			if ( empty( $file_results['messages'] ) ) {
+				continue;
+			}
+
+			foreach ( $file_results['messages'] as $file_message ) {
+				$result->add_message(
+					strtoupper( $file_message['type'] ) === 'ERROR',
+					$file_message['message'],
+					array(
+						'code'   => $file_message['source'],
+						'file'   => $file_name,
+						'line'   => $file_message['line'],
+						'column' => $file_message['column'],
+					)
+				);
+			}
+		}
+	}
+}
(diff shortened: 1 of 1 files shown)


includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,134 @@
...
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use WordPress\Plugin_Check\Checker\Check;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use Exception;
+
+/**
+ * Check for running one or more PHP CodeSniffer sniffs.
+ *
+ * @since 1.0.0
```
mukeshpanchal27:
```suggestion
 * @since n.e.x.t
```

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @since 1.0.0
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	public function run( Check_Result $result ) {
+		// Include the PHPCS autoloader.
+		$autoloader = TESTS_PLUGIN_DIR . '/vendor/squizlabs/php_codesniffer/autoload.php';
```
mukeshpanchal27: Should `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH` instead of `TESTS_PLUGIN_DIR`

```suggestion
		$autoloader = WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/vendor/squizlabs/php_codesniffer/autoload.php';
```

mukeshpanchal27 (commented):
Thanks @jjgrainger, Left some nit-pick.

- `Since` missing on function documents
- Use `n.e.x.t` instead of actual version `1.0.0`

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+		'sniffs',
+		'exclude',
+	);
+
+	/**
+	 * Returns an associative array of arguements to pass to PHPCS.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array
```
felixarntz: Please include a brief description on what that array should include. Specifically since this is an abstract method, it should include details on the format of the data.

See for example https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/post.php#L1552 on how WP core formats associative array doc-blocks, we could use that here too (except that here we do it for a `@return` doc instead of `@param`.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+		);
+
+		// Only accept allowed PHPCS arguments from check arguments array.
+		$check_args = array_filter(
+			$this->get_args(),
+			function( $key ) {
+				return in_array( $key, $this->allowed_args, true );
+			},
+			ARRAY_FILTER_USE_KEY
+		);
```
felixarntz: It'd be more efficient to make `$this->allowed_args` an associative array where the keys are the actual allowed args and the values are just something like `true`. This way, you can simplify this code to use:
```suggestion
		$check_args = array_intersect_key( $this->get_args(), $this->allowed_args );
```

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+	 * Amends the given result by running the check on the associated plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	public function run( Check_Result $result ) {
```
felixarntz, line 60: Maybe move this method to the top, since it's public. It's a good practice to have public methods first, then protected, then private.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+	abstract public function get_args();
+
+	/**
+	 * Creates the command arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check results including the plugin context to check.
+	 *
+	 * @return array
```
felixarntz: This also needs a description.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	abstract public function get_args();
+
+	/**
+	 * Creates the command arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check results including the plugin context to check.
+	 *
```
felixarntz: There shouldn't be blank lines between `@param` and `@return` docs per WP core docs standards.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @return array
+	 */
+	private function create_args( Check_Result $result ) {
```
felixarntz: Instead of calling `$this->get_args()` in this method, how about we make it a bit more "standalone", so that the value of `$this->get_args()` is passed to the method as a parameter? This way it becomes less coupled to that concrete value. A bit of a nit-pick, but makes the code cleaner and more reusable.

If we do that, maybe we also want to rename this method to something like `parse_args` or `parse_argv` to clarify the purpose better?

felixarntz (commented):
@jjgrainger Overall looks solid, a bit of nit-pick feedback.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -0,0 +1,140 @@
...
+ * @since n.e.x.t
+ */
+abstract class Abstract_PHP_CodeSniffer_Check implements Check {
+
+	/**
+	 * List of allowed PHPCS arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @var array
```
mukeshpanchal27:
```suggestion
	 * @since n.e.x.t
	 * @var array
```

There shouldn't be blank lines between `@since` and `@var` docs per WP core docs standards.

felixarntz (approved):
@jjgrainger Thanks, this looks excellent now!

mukeshpanchal27 (approved):
Thanks @jjgrainger, LGTM!

PR #56, opened 2023-01-11
Create Minimal_Theme_Preparation

### LINKED ISSUE #39: Create Use_Minimal_Theme_Preparation
## Description

Minimal theme preparation will ensure that the plugin being checked is in as much isolation as possible.

## Acceptance Criteria

- Class `Use_Minimal_Theme_Preparation` should be created and exists at `includes/Checker/Preparations`
- Class `Use_Minimal_Theme_Preparation` should extend `Preparation` interface
- Protected property `$theme_slug` and `$theme_dir` should be there
- Constructor should receive `$theme_slug` and `$theme_dir` as an arguments
- `prepare()` must be implemented from `Preparation` interface
- `prepare()` method should add necessary hooks/filters to remove theme supports and return cleanups 
- `get_theme_slug()` method should be created to get theme slug `$theme_slug`
- `get_theme_name()` method should use `wp_get_theme()` to get theme name using `$theme_slug` & `$theme_dir`
- `get_theme_root()` method should use `get_raw_theme_root()` using `$theme_slug`

## Test Coverage

- `Use_Minimal_Theme_Preparation` instance expected to be implementing of `Preparation` interface
- `get_theme_slug()` method expected to return theme slug provided as argument for class
- `get_theme_name()` method expected to return theme name
- `get_theme_root()` method expected to return theme root dir
- `prepare()` method expected to remove all unnecessary theme features and templates with hooks/filters

jjgrainger:
Adds the `Minimal_Theme_Preparation` class and `wp-empty-theme` theme.

- Addresses acceptance criteria in #39 
- While writing test the `get_theme_root()` method was failing. Looking into this the [`get_raw_theme_root()`](https://developer.wordpress.org/reference/functions/get_raw_theme_root/) function calls the [`get_theme_roots()`](https://developer.wordpress.org/reference/functions/get_theme_roots/) function which uses the `theme_roots` transient to search for a valid theme root based on the theme slug. Tests were failing because the `theme_roots` transient wasn't being updated when registering the new theme directory with `register_theme_directory()` in the prepare method. I've included a [`search_theme_directories( true );`](https://developer.wordpress.org/reference/functions/search_theme_directories/) function call after registering the theme directory in the prepare method in order to force update this transient.

Closes #39

### DIFF
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -1,0 +1,139 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Preparations\Use_Minimal_Theme_Preparation
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Preparations;
+
+use WordPress\Plugin_Check\Checker\Preparation;
+use Exception;
+
+/**
+ * Class for the preparation step to force usage of a minimal theme.
+ *
+ * This ensures the plugin is checked as much in isolation as possible.
+ *
+ * @since n.e.x.t
+ */
+class Use_Minimal_Theme_Preparation implements Preparation {
+
+	/**
+	 * Theme slug / directory name.
+	 *
+	 * @since n.e.x.t
+	 * @var string
+	 */
+	protected $theme_slug;
+
+	/**
+	 * Absolute path to themes root directory.
+	 *
+	 * @since n.e.x.t
+	 * @var string
+	 */
+	protected $themes_dir;
+
+	/**
+	 * Sets the theme slug and themes root directory.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $theme_slug Slug of the theme to enforce.
+	 * @param string $themes_dir Optional. Absolute path to themes root directory, if not the regular wp-content/themes.
+	 */
+	public function __construct( $theme_slug, $themes_dir = '' ) {
+		$this->theme_slug = $theme_slug;
+		$this->themes_dir = $themes_dir;
+	}
+
+	/**
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		// Override the theme slug and name.
+		add_filter( 'template', array( $this, 'get_theme_slug' ) );
+		add_filter( 'stylesheet', array( $this, 'get_theme_slug' ) );
+		add_filter( 'pre_option_template', array( $this, 'get_theme_slug' ) );
+		add_filter( 'pre_option_stylesheet', array( $this, 'get_theme_slug' ) );
+		add_filter( 'pre_option_current_theme', array( $this, 'get_theme_name' ) );
+
+		// Override the theme directory.
+		add_filter( 'pre_option_template_root', array( $this, 'get_theme_root' ) );
+		add_filter( 'pre_option_stylesheet_root', array( $this, 'get_theme_root' ) );
+
+		// Register the custom themes directory if relevant.
+		if ( ! empty( $this->themes_dir ) ) {
+			register_theme_directory( $this->themes_dir );
+		}
+
+		// Return the cleanup function.
+		return function() {
+			global $wp_theme_directories;
+
+			remove_filter( 'template', array( $this, 'get_theme_slug' ) );
+			remove_filter( 'stylesheet', array( $this, 'get_theme_slug' ) );
+			remove_filter( 'pre_option_template', array( $this, 'get_theme_slug' ) );
+			remove_filter( 'pre_option_stylesheet', array( $this, 'get_theme_slug' ) );
+			remove_filter( 'pre_option_current_theme', array( $this, 'get_theme_name' ) );
+
+			remove_filter( 'pre_option_template_root', array( $this, 'get_theme_root' ) );
+			remove_filter( 'pre_option_stylesheet_root', array( $this, 'get_theme_root' ) );
+
+			if ( ! empty( $this->themes_dir ) ) {
+				$index = array_search( untrailingslashit( $this->themes_dir ), $wp_theme_directories, true );
+				if ( false !== $index ) {
+					array_splice( $wp_theme_directories, $index, 1 );
+					$wp_theme_directories = array_values( $wp_theme_directories );
+				}
+			}
+		};
+	}
+
+	/**
+	 * Gets the theme slug.
+	 *
+	 * Used as a filter callback.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The theme slug.
+	 */
+	public function get_theme_slug() {
+		return $this->theme_slug;
+	}
+
+	/**
+	 * Gets the theme name.
+	 *
+	 * Used as a filter callback.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The theme name.
+	 */
+	public function get_theme_name() {
+		$theme = wp_get_theme( $this->theme_slug, $this->themes_dir );
+		return $theme->display( 'Name' );
+	}
+
+	/**
+	 * Gets the theme root.
+	 *
+	 * Used as a filter callback.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The theme root.
+	 */
+	public function get_theme_root() {
+		return get_raw_theme_root( $this->theme_slug, true );
+	}
+}
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -72,6 +72,9 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 		// Register the custom themes directory if relevant.
 		if ( ! empty( $this->themes_dir ) ) {
 			register_theme_directory( $this->themes_dir );
+
+			// Force new directory scan to ensure the new theme directory is available.
+			search_theme_directories( true );
 		}
 
 		// Return the cleanup function.
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -95,11 +95,11 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 				if ( false !== $index ) {
 					array_splice( $wp_theme_directories, $index, 1 );
 					$wp_theme_directories = array_values( $wp_theme_directories );
+
+					// Force new directory scan to remove the test theme directory.
+					search_theme_directories( true );
 				}
 			}
-
-			// Force new directory scan to remove the test theme directory.
-			search_theme_directories( true );
 		};
 	}
 
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -53,6 +53,8 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @global array $wp_theme_directories
+	 *
 	 * @return callable Cleanup function to revert any changes made here.
 	 *
 	 * @throws Exception Thrown when preparation fails.
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -73,7 +73,7 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 		if ( ! empty( $this->themes_dir ) ) {
 			register_theme_directory( $this->themes_dir );
 
-			// Force new directory scan to ensure the new theme directory is available.
+			// Force new directory scan to ensure the test theme directory is available.
 			search_theme_directories( true );
 		}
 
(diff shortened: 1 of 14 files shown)


includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
```diff
@@ -0,0 +1,142 @@
...
+			remove_filter( 'pre_option_current_theme', array( $this, 'get_theme_name' ) );
+
+			remove_filter( 'pre_option_template_root', array( $this, 'get_theme_root' ) );
+			remove_filter( 'pre_option_stylesheet_root', array( $this, 'get_theme_root' ) );
+
+			if ( ! empty( $this->themes_dir ) ) {
+				$index = array_search( untrailingslashit( $this->themes_dir ), $wp_theme_directories, true );
+				if ( false !== $index ) {
+					array_splice( $wp_theme_directories, $index, 1 );
+					$wp_theme_directories = array_values( $wp_theme_directories );
```
felixarntz, line 99: Since you fixed the transient problem above via `search_theme_directories( true )` call (good catch!), you probably need to do the same here, otherwise the list of themes will continue to include themes from our custom directory even later.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
```diff
@@ -0,0 +1,142 @@
...
+	 * Used as a filter callback.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The theme root.
+	 */
+	public function get_theme_root() {
+		return get_raw_theme_root( $this->theme_slug, true );
+	}
+}
```
felixarntz, line 147: Can you add unit tests for this class?

felixarntz (commented):
@jjgrainger Mostly looks good. Only one thing to fix in the code I think, but also we should add unit tests.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
jjgrainger: @felixarntz unit tests are setup here https://github.com/10up/plugin-check/pull/56/files#diff-f97377ea014b79e48c71fe98cb7b27e474c74960dd44538516850361badd7ac9

Might have been difficult to see with all the theme files included in the changes.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,45 @@
...
+	public function tear_down() {
+		parent::tear_down();
+
+		// Run the cleanup function.
+		( $this->cleanup )();
+	}
+
+	public function test_implements_preparation_interface() {
+		$this->assertInstanceOf( Preparation::class, $this->preparation );
+	}
```
felixarntz: See my previous feedback - let's not test that the class to test has that class, that really is unnecessary :)

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
felixarntz: @jjgrainger Apologies, missed that indeed!

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,45 @@
...
+class Use_Minimal_Theme_Preparation_Tests extends WP_UnitTestCase {
+	public function set_up() {
+		parent::set_up();
+
+		$this->theme_slug = 'wp-empty-theme';
+		$this->theme_dir  = WP_PLUGIN_DIR . '/' . basename( TESTS_PLUGIN_DIR ) . '/test-content/themes';
+
+		$this->preparation = new Use_Minimal_Theme_Preparation( $this->theme_slug, $this->theme_dir );
+
+		$this->cleanup = $this->preparation->prepare();
```
felixarntz: We should have a test for what this method does, rather than calling it in `set_up`.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,45 @@
...
+
+use WordPress\Plugin_Check\Checker\Preparation;
+use WordPress\Plugin_Check\Checker\Preparations\Use_Minimal_Theme_Preparation;
+
+class Use_Minimal_Theme_Preparation_Tests extends WP_UnitTestCase {
+	public function set_up() {
+		parent::set_up();
+
+		$this->theme_slug = 'wp-empty-theme';
+		$this->theme_dir  = WP_PLUGIN_DIR . '/' . basename( TESTS_PLUGIN_DIR ) . '/test-content/themes';
```
felixarntz: Similar to a previous comment, please keep in mind to introduce _actual_ class properties when you reference them somewhere.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,45 @@
...
+
+		$this->cleanup = $this->preparation->prepare();
+	}
+
+	public function tear_down() {
+		parent::tear_down();
+
+		// Run the cleanup function.
+		( $this->cleanup )();
+	}
```
felixarntz: See my other comment above, I think for this the `set_up` and `tear_down` methods are more hindering than they help. We only need to call the `prepare` method in the specific tests where we test it, and then we can call the cleanup callback in just those same tests.

I think you can instantiate the class in the individual test methods, that's perfectly fine.

IMO we should reduce those two methods to just the two `$theme_slug` and `$theme_dir` class properties with their values, so that we don't have to rewrite those values everywhere.

felixarntz (commented):
@jjgrainger Left some feedback on the test class.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
```diff
@@ -0,0 +1,145 @@
...
+			if ( ! empty( $this->themes_dir ) ) {
+				$index = array_search( untrailingslashit( $this->themes_dir ), $wp_theme_directories, true );
+				if ( false !== $index ) {
+					array_splice( $wp_theme_directories, $index, 1 );
+					$wp_theme_directories = array_values( $wp_theme_directories );
+				}
+			}
+
+			// Force new directory scan to remove the test theme directory.
+			search_theme_directories( true );
```
felixarntz: Nit-pick, we only need to run this in the above condition where `$wp_theme_directories` is adjusted. If we make no change to it, we also don't need to run this.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,61 @@
...
+		$cleanup     = $preparation->prepare();
+
+		$this->assertIsCallable( $cleanup );
+		$this->assertSame( $this->theme_slug, get_option( 'template' ) );
+		$this->assertSame( $this->theme_slug, get_option( 'stylesheet' ) );
+		$this->assertSame( $this->theme_name, get_option( 'current_theme' ) );
+		$this->assertSame( str_replace( WP_CONTENT_DIR, '', $this->theme_dir ), get_option( 'template_root' ) );
+		$this->assertSame( str_replace( WP_CONTENT_DIR, '', $this->theme_dir ), get_option( 'stylesheet_root' ) );
+
+		$cleanup();
```
felixarntz: The cleanup will not execute if any of the assertions fail. Can you assign all the call results to variables? This way you can then do the cleanup, and the assertions at the very end.

You already do something like that in the methods below, we need to do it here too.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,61 @@
...
+	protected $theme_name;
+	protected $theme_slug;
+	protected $theme_dir;
+
+	public function set_up() {
+		parent::set_up();
+
+		$this->theme_name = 'WP Empty Theme';
+		$this->theme_slug = 'wp-empty-theme';
+		$this->theme_dir  = WP_PLUGIN_DIR . '/' . basename( TESTS_PLUGIN_DIR ) . '/test-content/themes';
```
felixarntz: Nit-pick, wouldn't the following result in the same but simpler?
```suggestion
		$this->theme_dir  = TESTS_PLUGIN_DIR . '/test-content/themes';
```

felixarntz (commented):
@jjgrainger Almost good to go now, left 3 small final comments.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Tests for the Check_Context class.
```
mukeshpanchal27:
```suggestion
 * Tests for the Use_Minimal_Theme_Preparation class.
```

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
```diff
@@ -0,0 +1,145 @@
...
+	public function __construct( $theme_slug, $themes_dir = '' ) {
+		$this->theme_slug = $theme_slug;
+		$this->themes_dir = $themes_dir;
+	}
+
+	/**
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
```
mukeshpanchal27, line 55:
```suggestion
	 *
	 * @global array $wp_theme_directories
	 *
```

Add documentation for global variable.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,66 @@
...
+		$this->assertSame( str_replace( WP_CONTENT_DIR, '', $this->theme_dir ), $stylesheet_root );
+
+		$cleanup();
+	}
+
+	public function test_get_theme_name() {
+		$preparation = new Use_Minimal_Theme_Preparation( $this->theme_slug, $this->theme_dir );
+		$cleanup     = $preparation->prepare();
+		$theme_name  = $preparation->get_theme_name();
+		$cleanup();
```
mukeshpanchal27, line 51:
```suggestion
		$theme_name  = $preparation->get_theme_name();
```

`test_get_theme_name` does not need `cleanup()` to run?

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Tests for the Check_Context class.
+ *
+ * @package plugin-check
+ */
+
+use WordPress\Plugin_Check\Checker\Preparation;
```
mukeshpanchal27: @jjgrainger Is it necessary to add this class because it already exists in the actual class file? Correct me if I miss anything.

mukeshpanchal27 (commented):
Thanks @jjgrainger, Left few final feedback and questions.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
jjgrainger: @mukeshpanchal27 this is needed in order to activate the test theme and return the correct theme name we're expecting. The cleanup is required and should be run before running assertions.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
jjgrainger: @mukeshpanchal27 good spot and I've removed.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -0,0 +1,65 @@
...
+		$stylesheet_root = get_option( 'stylesheet_root' );
+
+		$this->assertIsCallable( $cleanup );
+		$this->assertSame( $this->theme_slug, $template );
+		$this->assertSame( $this->theme_slug, $stylesheet );
+		$this->assertSame( $this->theme_name, $current_theme );
+		$this->assertSame( str_replace( WP_CONTENT_DIR, '', $this->theme_dir ), $template_root );
+		$this->assertSame( str_replace( WP_CONTENT_DIR, '', $this->theme_dir ), $stylesheet_root );
+
+		$cleanup();
```
felixarntz: This is still problematic following my last comment. Please move it before the assertions so that it'll always execute.

You already assigned variables, but then still kept this at the bottom, which was the main point of doing so :)

test-content/themes/wp-empty-theme/style.css
```diff
@@ -0,0 +1,15 @@
+/*
+Theme Name: WP Empty Theme
+Theme URI: https://team.git.corp.google.com/dev-web-ecosystem/wp-empty-theme
+Author: Felix Arntz, Google
+Author URI: https://opensource.google.com
+Description: "Empty" WordPress theme for testing a barebones WordPress site, as a reference point to compare actual themes and plugins to.
+Requires at least: 5.3
+Tested up to: 5.8
```
mukeshpanchal27, line 8:
```suggestion
Tested up to: 6.1.1
```

Not blocker, I'd like to hear @felixarntz's thoughts on this before the final review.

mukeshpanchal27 (approved):
Thanks @jjgrainger, Only one out standing review https://github.com/10up/plugin-check/pull/56#discussion_r1082163292

test-content/themes/wp-empty-theme/style.css
felixarntz: @mukeshpanchal27 I think we can potentially update that separately. Not a big deal really, given that this theme would not really be used and the end user wouldn't be able to see it.

I'd like to move ahead and get this PR merged, so we can consider doing that later. Nit-pick: I think we should stick to major WordPress releases in these annotations (i.e. make this `6.1` instead of `6.1.1`).

felixarntz (approved):
Thanks @jjgrainger!

PR #57, opened 2023-01-12
Create Checks class

### LINKED ISSUE #12: Create Checks class
## Description

The Checks class will be the main class for running checks against a plugin. This class contains all the logic to collect the Plugin and Check contexts, run Preparations ahead of checks, loop over and run all Checks and collect and return the results.

Containing this logic within a single class decoupled from I/O allows this class to be used in the different contexts in which checks can run, such as WP-CLI or in the WordPress admin, without duplicating efforts and logic which is described more in detail in the [Check Orchestration and Execution Approach](https://docs.google.com/document/d/1Mge9i2_EMH7IEJD37vWXX9Uzyaz-p9awD1pI4AQ76Rg/edit?resourcekey=0-y7fxvr6cmxv_QW5xPylGcg#heading=h.d8g2nxbaaltl).

The class will contain a method to return all available checks that can be run against the plugin, which is filterable so developers can extend/modify which checks should run.

Additionally, a method to run a set of checks (by default all checks) should be provided, so the entire suite of checks can be used or a plugin can be tested against one check alone as checks are provided as a list. The method can iterate through the list of checks and execute them. For example, it allows running a single check when a CLI command is executed with one check to test for a plugin and it also allows running a single check as part of running all checks for a plugin when a CLI command or even AJAX requests is initiated to test a plugin for all checks.

The Checks class is also responsible for preparation work ahead of running tests and cleanup after tests have run.

## Acceptance Criteria

- `Checks` class is created and exists at `includes/Checker`
- Protected property `$check_context` should be there which is an instance of `Check_Context`
- Constructor receives the `$plugin_main_file` as an argument
- Method `get_checks` should be created as public and returns the list of checks
- Method `run_check_with_result` should be created as protected and receives `Check $check` and `Check_Result $check_result` as an arguments.
- Method `run_checks` should be created as public and accepts a single parameter `array $checks` which is an array of the check class names to run. The method will loop over the `$checks` array and run each check by calling `run_check_with_result`.

## Tests Coverage

- `get_checks` method returns the list of expected checks
- `run_check_with_result` runs expected check and result is stored as expected
- `run_checks` runs the expected checks based on the parameter passed

jjgrainger:
Adds the `Checks` class

- Addresses the acceptance criteria in #12 
- TODO: Check's will need to be added to the `get_checks()` method once implemented. Check AC's will be updated to include this step.

Closes #12

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "WordPress\\Plugin_Check\\Tests\\": "tests/testdata"
+      "WordPress\\Plugin_Check\\Test_Data\\": "tests/testdata"
     }
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -30,5 +30,10 @@
     "psr-4": {
       "WordPress\\Plugin_Check\\": "includes/"
     }
+  },
+  "autoload-dev": {
+    "psr-4": {
+      "WordPress\\Plugin_Check\\Tests\\": "tests/utils"
+    }
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "WordPress\\Plugin_Check\\Tests\\": "tests/utils"
+      "WordPress\\Plugin_Check\\Tests\\": "tests/testdata"
     }
   }
 }
--- a/includes/Checker/Checks.php
+++ b/includes/Checker/Checks.php
@@ -8,6 +8,7 @@
 namespace WordPress\Plugin_Check\Checker;
 
 use WordPress\Plugin_Check\Checker\Check_Context;
+use ReflectionClass;
 use Exception;
 
 /**
@@ -37,55 +38,35 @@ class Checks {
 	}
 
 	/**
-	 * Runs all checks against the plugin.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return Check_Result Object containing all check results.
-	 *
-	 * @throws Exception Thrown when checks fail with critical error.
-	 */
-	public function run_all_checks() {
-		$result = new Check_Result( $this->check_context );
-		$checks = $this->get_checks();
-
-		array_walk(
-			$checks,
-			function( Check $check ) use ( $result ) {
-				$this->run_check_with_result( $check, $result );
-			}
-		);
-
-		return $result;
-	}
-
-	/**
 	 * Runs a single check against the plugin.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param string $check Check class name.
+	 * @param array $checks Ab array of Check class names to run.
 	 * @return Check_Result Object containing all check results.
 	 *
 	 * @throws Exception Thrown when check fails with critical error.
 	 */
-	public function run_single_check( $check ) {
-		$result = new Check_Result( $this->check_context );
-		$checks = $this->get_checks();
+	public function run_checks( array $checks ) {
+		$result     = new Check_Result( $this->check_context );
+		$all_checks = $this->get_checks();
 
-		// Look up the check based on the $check variable.
-		$check_index = array_search( $check, $checks, true );
-		if ( false === $check_index ) {
-			throw new Exception(
-				sprintf(
-					/* translators: %s: class name */
-					__( 'Invalid check class name %s.', 'plugin-check' ),
-					$check
-				)
-			);
-		}
+		// Create an array of Check objects to run based on the check names passed.
+		$checks_to_run = array_filter(
+			$all_checks,
+			function( $check ) use ( $checks ) {
+				$check_name = ( new ReflectionClass( $check ) )->getShortName();
+				return in_array( $check_name, $checks, true );
+			}
+		);
 
-		$this->run_check_with_result( $checks[ $check_index ], $result );
+		// Run the checks.
+		array_walk(
+			$checks_to_run,
+			function( Check $check ) use ( $result ) {
+				$this->run_check_with_result( $check, $result );
+			}
+		);
 
 		return $result;
 	}
--- a/includes/Checker/Checks.php
+++ b/includes/Checker/Checks.php
@@ -8,7 +8,6 @@
 namespace WordPress\Plugin_Check\Checker;
 
 use WordPress\Plugin_Check\Checker\Check_Context;
-use ReflectionClass;
 use Exception;
 
 /**
@@ -38,11 +37,11 @@ class Checks {
 	}
 
 	/**
-	 * Runs a single check against the plugin.
+	 * Runs checks against the plugin.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $checks Ab array of Check class names to run.
+	 * @param array $checks An array of Check objects to run.
 	 * @return Check_Result Object containing all check results.
 	 *
 	 * @throws Exception Thrown when check fails with critical error.
@@ -53,10 +52,9 @@ class Checks {
 
 		// Create an array of Check objects to run based on the check names passed.
 		$checks_to_run = array_filter(
-			$all_checks,
-			function( $check ) use ( $checks ) {
-				$check_name = ( new ReflectionClass( $check ) )->getShortName();
-				return in_array( $check_name, $checks, true );
+			$checks,
+			function( $check ) use ( $all_checks ) {
+				return in_array( $check, $all_checks, true );
 			}
 		);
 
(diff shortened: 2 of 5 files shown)


includes/Checker/Checks.php
```diff
@@ -0,0 +1,182 @@
...
+	 * @throws Exception Thrown when checks fail with critical error.
+	 */
+	public function run_all_checks() {
+		if ( ! $this->prepared ) {
+			throw new Exception(
+				__( 'Environment not prepared to run checks. The Checks::prepare() method must be called first.', 'plugin-check' )
+			);
+		}
+
+		$result = new Check_Result( $this->main_context, $this->check_context );
```
felixarntz: This is copied over from the PoC, but doesn't work now. Please double-check, we don't have `$main_context` here. :)

includes/Checker/Checks.php
```diff
@@ -0,0 +1,182 @@
...
+	 * @throws Exception Thrown when check fails with critical error.
+	 */
+	public function run_single_check( $check ) {
+		if ( ! $this->prepared ) {
+			throw new Exception(
+				__( 'Environment not prepared to run checks. The Checks::prepare() method must be called first.', 'plugin-check' )
+			);
+		}
+
+		$result = new Check_Result( $this->main_context, $this->check_context );
```
felixarntz: See above.

includes/Checker/Checks.php
```diff
@@ -0,0 +1,182 @@
...
+	 * Prepares the environment for running checks and returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		// TODO: Add preparations once implemented.
```
felixarntz: What's the intended sequence between this PR and #57? Again, this is the type of thing we should probably have clarified in ACs: Either, if #57 comes first, #56 needs to include adding the preparation in here, or otherwise, #57 needs to include adding the preparation from #56 in here.

I see there is an open comment on the issue related to this, but if we keep this in here, it needs to be addressed.

tests/Checker/Checks_Tests.php
```diff
@@ -0,0 +1,56 @@
...
+
+	public function test_prepare_returns_callable() {
+		$cleanup = $this->checks->prepare();
+
+		$this->assertIsCallable( $cleanup );
+	}
+
+	public function test_get_checks_returns_array_of_expected_checks() {
+		$expected = array(
+			'example_check_class',
```
felixarntz: This test assumes a class name (nit-pick: why lowercase? that's not a good convention for class names), but the code documentation is explicitly referring to class _instances_. Let's align there. I generally think instances provide more flexibility, for example since it allows to use the same class but with different configurations for a check. But there may also be some arguments for using classes (for example it would allow to only load a class PHP file if that class is actually used). Or maybe the most flexible is to allow for a mixed approach:
* Typically require a class name. In that case, the plugin checker will instantiate the class without any constructor param.
* For more flexible class usage, a closure should be provided instead which returns the necessary class instance.

We can think about which way to go with, but right now the test here is inconsistent with what is required per the documentation in the `get_checks` method. WDYT on the different approaches, which one should we go with? We should update the documentation and this test accordingly based on what we decide.

tests/Checker/Checks_Tests.php
```diff
@@ -0,0 +1,56 @@
...
+		$this->expectException( Exception::class );
+
+		$this->checks->run_all_checks();
+	}
+
+	public function test_run_single_check_throws_exception_if_not_prepared() {
+		$this->expectException( Exception::class );
+
+		$this->checks->run_single_check( 'check' );
+	}
```
felixarntz: Can you also add tests for these two methods when the environment _is_ prepared? Right now we only test the error case, but it's arguably even more important to test the success case.

If you had those tests, it would also have shown the PHP error in the `Checks` class above.

felixarntz (commented):
@jjgrainger There are still a few things to address here, left some feedback.

composer.json
```diff
@@ -30,5 +30,10 @@
     "psr-4": {
       "WordPress\\Plugin_Check\\": "includes/"
     }
+  },
+  "autoload-dev": {
+    "psr-4": {
+      "WordPress\\Plugin_Check\\Tests\\": "tests/testdata"
```
felixarntz: Maybe better to use a `WordPress\Plugin_Check\Test_Data` namespace for this? It's more specific to what the folder indicates should go there. For example, if we introduced test utility classes, they would also semantically fall under the namespace used here, but they would not be the same as test data. So maybe for now a test data namespace works better for this purpose.

includes/Checker/Checks.php
```diff
@@ -0,0 +1,125 @@
...
+		$all_checks = $this->get_checks();
+
+		// Create an array of Check objects to run based on the check names passed.
+		$checks_to_run = array_filter(
+			$all_checks,
+			function( $check ) use ( $checks ) {
+				$check_name = ( new ReflectionClass( $check ) )->getShortName();
+				return in_array( $check_name, $checks, true );
+			}
+		);
```
felixarntz: I realize we had not defined that before, but this is a bit overcomplicated. Not a fan of using `ReflectionClass` either.

We will need check orchestration to happen outside this class, in the runner classes probably, so that means those outside classes need access to the check instances anyway. I think we should simply expect an array of `Check` instances for this method. Then the only validation we would need to do is whether the given checks are part of the `$all_checks` array (by instance comparison).

includes/Checker/Checks.php
```diff
@@ -0,0 +1,125 @@
...
+	public function __construct( $plugin_main_file ) {
+		$this->check_context = new Check_Context( $plugin_main_file );
+	}
+
+	/**
+	 * Runs a single check against the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks Ab array of Check class names to run.
```
felixarntz: See other comment, should probably be `An array of Check instances to run.` (also see typo in the first word).

tests/Checker/Checks_Tests.php
```diff
@@ -0,0 +1,77 @@
...
+		);
+
+		add_filter(
+			'wp_plugin_check_checks',
+			function( $checks ) use ( $expected ) {
+				return $expected;
+			}
+		);
+
+		$results = $this->checks->run_checks( array( 'Example_Check' ) );
```
felixarntz: See above, we can pass the same instance here.

felixarntz (commented):
@jjgrainger A few small things, almost good to go.

includes/Checker/Checks.php
```diff
@@ -0,0 +1,123 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of Check objects to run.
+	 * @return Check_Result Object containing all check results.
+	 *
+	 * @throws Exception Thrown when check fails with critical error.
+	 */
+	public function run_checks( array $checks ) {
+		$result     = new Check_Result( $this->check_context );
```
mukeshpanchal27, line 50: @jjgrainger Do we needs to add `use WordPress\Plugin_Check\Checker\Check_Result;`?

jjgrainger: @mukeshpanchal27 it's not necessarily a requirement as it shares the same namespace, but I see no harm in explicitly defining this.

felixarntz: Not really needed, I think we can avoid that in the future. But okay to keep for now.

felixarntz (approved):
Great work, thanks @jjgrainger!

tests/Checker/Checks_Tests.php
```diff
@@ -0,0 +1,87 @@
...
+
+	protected $checks;
+
+	public function set_up() {
+		parent::set_up();
+
+		$this->checks = new Checks( 'test-plugin/test-plugin.php' );
+	}
+
+	public function test_get_checks_returns_array_of_expected_checks() {
```
vishalkakadiya, line 22: @jjgrainger Can we add some simple function doc here, so it will be easy to know which test function covers which actual function? something like below:

```
/**
 * @covers Checks::get_checks()
 */
 ```

vishalkakadiya:
@jjgrainger Added comment to add function doc block, other than that LGTM!, Thanks!

vishalkakadiya (approved):

tests/Checker/Checks_Tests.php
jjgrainger: Thanks @vishalkakadiya the method being tested is actually covered by the test name `test_get_checks...` so there is no need for a docblock for this.

PR #62, opened 2023-01-19
Add i18n usage check class and test plugin files for unit tests

### LINKED ISSUE #18: Create I18n_Usage_Check
## Description

Checks that WordPress internationalization functions are used properly, including using a valid textdomain that matches the plugin’s textdomain in the main file comment header. It will also include checks for each of the localization functions for typos and ensure the correct parameters are passed. This will use the existing `WordPress.WP.I18n` sniff in the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards/).

## Acceptance Criteria

- Class `I18n_Usage_Check` should be created and exists at `includes/Checker/Checks`
- Class `I18n_Usage_Check` should extend the `Abstract_PHP_CodeSniffer_Check ` abstract class
    - `'sniffs' => 'WordPress.WP.I18n'`
    - `'standard' => 'WordPress'`

## Tests Coverage
- `run()` method adds a correct message to the `$check_result` when checking an invalid file
- `run()` method does not add any message to the `$check_result` when checking a valid file

vishalkakadiya:
Adds the I18n_Usage_Check class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/18

Closes https://github.com/10up/plugin-check/issues/18


includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,40 @@
+<?php
+/**
+ * Translation check class.
```
mukeshpanchal27:
```suggestion
 * Class WordPress\Plugin_Check\Checker\Checks\I18n_Usage_Check
```

For consistency, use a similar description.

includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+/**
+ * Interface for a single check.
```
mukeshpanchal27: Should be change. @jjgrainger Any thoughts?

i.e.
```
Check for running WordPress internationalization sniffs. 
```

phpcs.xml.dist
```diff
@@ -47,6 +47,19 @@
 	<rule ref="Squiz.Commenting.FunctionCommentThrowTag.Missing">
 		<exclude-pattern>tests/*</exclude-pattern>
 	</rule>
+	<rule ref="Squiz.Commenting.FileComment.MissingPackageTag">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
```
mukeshpanchal27: It's the "Inline Documentation and Comments" rule. Is an i18n check required? Also, WPCS did not support it. https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress-Docs/ruleset.xml

phpcs.xml.dist
```diff
@@ -47,6 +47,19 @@
...
+	<rule ref="WordPress.WP.I18n.MissingTranslatorsComment">
```
mukeshpanchal27:
```suggestion

	<rule ref="WordPress.WP.I18n.MissingTranslatorsComment">
```

Add empty line before I18n rules.

phpcs.xml.dist
```diff
@@ -47,6 +47,19 @@
...
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+	<rule ref="WordPress.WP.I18n.TextDomainMismatch">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+	<rule ref="WordPress.WP.I18n.NonSingularStringLiteralDomain">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+
```
mukeshpanchal27:
```suggestion
```

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * @covers I18n_Usage_Check::get_args()
+	 */
+	public function test_get_args() {
+
+		$sniffs = $this->i18n_usage_check->get_args();
+
+		$this->assertArrayHasKey( 'extensions', $sniffs );
+		$this->assertArrayHasKey( 'standard', $sniffs );
+		$this->assertArrayHasKey( 'sniffs', $sniffs );
```
mukeshpanchal27: Because `assertEquals` checks the entire array, it is worthwhile to check individual properties.

tests/test-plugin-with-errors/test-plugin-with-errors.php
```diff
@@ -0,0 +1,18 @@
...
+ * Author: WordPress Performance Team
+ * Author URI: https://make.wordpress.org/performance/
+ * License: GPLv2 or later
+ * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ * Text Domain: test-plugin-check-errors
+ *
+ * @package test-plugin-check-errors
+ */
+
+include_once 'i18-usage-error-file.php';
```
mukeshpanchal27: Can we use a single plugin file instead of multiple files? Also, can we test valid and invalid text domains in a single plugin instead of two plugins?

**Valid text domain:**
```
esc_html_e( 'Hello World!', 'test-plugin-check' );

$text_domain = 'test-plugin-check';
esc_html_e( 'Hello World!', $text_domain );
```

**Invalid text domain:**
```
esc_html_e( 'Hello World!', 'test-plugin-check-error' );

$text_domain = 'test-plugin-check-errors';
esc_html_e( 'Hello World!', $text_domain );
```

mukeshpanchal27 (changes requested):
Thanks @vishalkakadiya, Good start! Left some feedback and questions.

phpcs.xml.dist
vishalkakadiya: @mukeshpanchal27 That issue was showing as the phpcs issue in the PR checks. So we might need to update the version here or need to discuss it.

tests/test-plugin-with-errors/test-plugin-with-errors.php
vishalkakadiya: @mukeshpanchal27 Actually, I have added two individual plugins as part of the infrastructure so that all other future checks' template test files can be added to their respective directories, like the "enqueued scripts sizes check" files, and so on.

Let me know if you need more information here. Thank you! 🙂

includes/Checker/Checks/I18n_Usage_Check.php
vishalkakadiya: @mukeshpanchal27 Updated now, thank you!

includes/Checker/Checks/I18n_Usage_Check.php
vishalkakadiya: @mukeshpanchal27 Updated now, thank you!

phpcs.xml.dist
vishalkakadiya: @mukeshpanchal27 Updated now, thank you!

phpcs.xml.dist
vishalkakadiya: @mukeshpanchal27 Updated now, thank you!

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: @mukeshpanchal27 Updated now, thank you!

vishalkakadiya:
@mukeshpanchal27 I have addressed the feedback and replied to your comments. Thank you! 🙂

includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Returns an associative array of arguments to pass to PHPCS.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An associative array of PHPCS CLI arguments.
+	 */
+	public function get_args() {
+
+		return $this->allowed_args;
```
jjgrainger: @vishalkakadiya can we return the array here rather than assigning the array to a property.

includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,40 @@
...
+
+	/**
+	 * List of I18n check arguments.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	public $allowed_args = array(
+		'extensions' => 'php',
+		'standard'   => 'WordPress,WordPress-Core,WordPress-Docs,WordPress-Extra',
```
jjgrainger: @vishalkakadiya do we need to use all these standards? Will either `WordPress` or `WordPress-Core` work?

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,93 @@
...
+			$sniffs
+		);
+	}
+
+	/**
+	 * @covers I18n_Usage_Check::run()
+	 */
+	public function test_run_with_errors() {
+
+		$check_context = new Check_Context( WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'tests/test-plugin-with-errors/test-plugin-with-errors.php' );
```
jjgrainger: @vishalkakadiya can we not use the `TESTS_PLUGIN_DIR` constant set in the tests bootstrap file? This will remove the need to include the plugin file in the test bootstrap file too.

```suggestion
		$check_context = new Check_Context( TESTS_PLUGIN_DIR . 'tests/test-plugin-with-errors/test-plugin-with-errors.php' );
```

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,93 @@
...
+		$this->assertArrayHasKey( 'code', $errors['i18-usage-error-file.php'][15][29][0] );
+		$this->assertEquals( 'WordPress.WP.I18n.NonSingularStringLiteralDomain', $errors['i18-usage-error-file.php'][15][29][0]['code'] );
+	}
+
+	/**
+	 * @covers I18n_Usage_Check::run()
+	 */
+	public function test_run_without_errors() {
+
+		$check_context = new Check_Context( WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'tests/test-plugin-without-errors/test-plugin-without-errors.php' );
```
jjgrainger: @vishalkakadiya same again here regarding the constant. Can we use the `TESTS_PLUGIN_DIR` constant instead?

```suggestion
		$check_context = new Check_Context( TESTS_PLUGIN_DIR . 'tests/test-plugin-without-errors/test-plugin-without-errors.php' );
```

tests/bootstrap.php
```diff
@@ -26,3 +26,6 @@
 
 // Start up the WP testing environment.
 require $_test_root . '/includes/bootstrap.php';
+
+// Load the plugin checker plugin, so the important constants and functions can be used in unit test.
+require_once TESTS_PLUGIN_DIR . '/plugin-check.php';
```
jjgrainger: @vishalkakadiya Can we remove this, I don't think we need be doing this here as we have other constants defined in this file we can use.

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,93 @@
...
+		$check_result = new Check_Result( $check_context );
+
+		$this->i18n_usage_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		$this->assertEmpty( $errors );
+		$this->assertEquals( 0, $check_result->get_error_count() );
+
+		$warnings = $check_result->get_warnings();
```
jjgrainger, line 58: @vishalkakadiya tiny nit-pick here, but let's keep assertions grouped together. Can we assign the `$warnings` variable directly under `$errors`.

tests/test-plugin-with-errors/test-plugin-with-errors.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
jjgrainger: @vishalkakadiya can we move both plugins to the following directory `tests/testdata/plugins/...`. Theres other testdata being added and this will help to organise and separate those from the actual tests.

jjgrainger (changes requested):
Thanks @vishalkakadiya 

Overall looking good and have left feedback on a number of things.

tests/bootstrap.php
vishalkakadiya: @jjgrainger Actually, without loading this plugin file, the `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH` constant is not found while running the `Abstract_PHP_CodeSniffer_Check->run()` function available at [here](https://github.com/10up/plugin-check/blob/trunk/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php#L62) from the test cases. So either we need to change the constant in that `run` method, or I need to load that file in the individual test case. Let me know what's your thoughts here.

Thank you!

includes/Checker/Checks/I18n_Usage_Check.php
vishalkakadiya: Done

includes/Checker/Checks/I18n_Usage_Check.php
vishalkakadiya: @jjgrainger Used `WordPress` only, thank you!

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: Updated this now, thank you!

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: Updated this now, thank you!

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: Updated this now, thank you!

tests/test-plugin-with-errors/test-plugin-with-errors.php
vishalkakadiya: Updated this now. Thank you!

vishalkakadiya:
@jjgrainger I have addressed your feedback now and left one question for you. Thank you!

jjgrainger (approved):
Thanks @vishalkakadiya approved.

vishalkakadiya:
@jjgrainger Thank you! 🙌 

@felixarntz It is ready for your review now. Thank you! 🙂

phpcs.xml.dist
```diff
@@ -47,6 +47,20 @@
...
+	<!-- Do not apply I18n rules for unit tests -->
+	<rule ref="WordPress.WP.I18n.MissingTranslatorsComment">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+	<rule ref="WordPress.WP.I18n.TextDomainMismatch">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+	<rule ref="WordPress.WP.I18n.NonSingularStringLiteralDomain">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
```
felixarntz: There are two problems with adding this here:
* These rules are ignored for all our tests, but they should only be ignored for the test plugins.
* We will need to add more and more exceptions in the future if we do this for every single PHPCodeSniffer check in the future, with more intentionally "wrong" test plugin code.

So I would suggest that we remove this and instead ignore the `tests/testdata/plugins` for _all_ sniffs (i.e. global exclude, somewhere above in this file, e.g. below the `<file>` rules.

includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,32 @@
...
+
+	/**
+	 * Returns an associative array of arguments to pass to PHPCS.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An associative array of PHPCS CLI arguments.
+	 */
+	public function get_args() {
+
```
felixarntz:
```suggestion
```

includes/Checker/Checks/I18n_Usage_Check.php
```diff
@@ -0,0 +1,32 @@
...
+class I18n_Usage_Check extends Abstract_PHP_CodeSniffer_Check {
+
+	/**
+	 * Returns an associative array of arguments to pass to PHPCS.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An associative array of PHPCS CLI arguments.
+	 */
+	public function get_args() {
```
felixarntz: This is not a problem from this PR, but just spotting that this method is unnecessarily public. Can we change it to `protected`, both here and in `Abstract_PHP_CodeSniffer_Check`? I think that would be more appropriate, since this shouldn't be needed from outside the class. cc @jjgrainger

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,92 @@
...
+
+class I18n_Usage_Check_Tests extends WP_UnitTestCase {
+
+	protected $i18n_usage_check;
+
+	public function set_up() {
+		parent::set_up();
+
+		$this->i18n_usage_check = new I18n_Usage_Check();
+	}
```
felixarntz: Maybe a bit of a personal preference, but I don't think we need a `set_up` method for something as simple as this. Makes the test code a bit harder to follow, and there isn't really a benefit here avoiding this "duplicate code".

tests/Checker/Checks/I18n_Usage_Check_Tests.php
```diff
@@ -0,0 +1,92 @@
...
+
+		$this->assertEquals(
+			array(
+				'extensions' => 'php',
+				'standard'   => 'WordPress',
+				'sniffs'     => 'WordPress.WP.I18n',
+			),
+			$sniffs
+		);
+	}
```
felixarntz: See my above comment; if we make this method `protected`, we can't test it here anymore. That's fine though, since it's indirectly covered by the below tests, where it is called. If this method is broken, the tests below would fail.

tests/bootstrap.php
felixarntz: I agree with @jjgrainger that this shouldn't be necessary, let's find a cleaner solution.

My first question is: Why are these constants not loaded? Shouldn't our plugin be loaded when running unit tests? This suggests there's something wrong in our unit test configuration. I think we are missing something like this in our `bootstrap.php` file: https://github.com/WordPress/performance/blob/1.9.0/tests/bootstrap.php#L35-L38

This is a more appropriate solution as it fixes the root problem.

felixarntz (changes requested):
@vishalkakadiya Mostly LGTM, but a few small things that need to be fixed.

tests/bootstrap.php
vishalkakadiya: @felixarntz This solution looks good to me. I'm applying this now. Thank you!
CC @jjgrainger

phpcs.xml.dist
vishalkakadiya: @felixarntz Make sense, I added the exclusion for `tests/testdata/plugins` now. Thank you!

includes/Checker/Checks/I18n_Usage_Check.php
vishalkakadiya: Done.

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: Removed it.

tests/Checker/Checks/I18n_Usage_Check_Tests.php
vishalkakadiya: Correct, we can test this method with the Reflection class, but that is not needed, as you said, so I just removed that for now. Thank you!

vishalkakadiya:
@felixarntz I have addressed the feedback on it. Please review it now. Thank you! 🙂

felixarntz (approved):
@vishalkakadiya Looks great, thanks!

mukeshpanchal27 (approved):
Thanks @vishalkakadiya, The changes look good to me. Great work!

PR #65, opened 2023-01-26
Add With_Shared_Preparations class

### LINKED ISSUE #59: Create With_Shared_Preperations interface
## Description
Create a `With_Shared_Preparations` interface. This interface is used on Checks that require reusable preperations.

## Acceptance Criteria
- A new `With_Shared_Preparations` interface exists in `includes/Checker/With_Shared_Preparations.php`
- The `With_Shared_Preparations` interface contains a single public `get_shared_preparations()` method
- The `get_shared_preparations()` method accepts no parameters
- The `get_shared_preparations()` method returns an array

vishalkakadiya:
Adds the With_Shared_Preparations class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/59

Closes https://github.com/10up/plugin-check/issues/59

### DIFF
--- a/includes/Checker/With_Shared_Preparations.php
+++ b/includes/Checker/With_Shared_Preparations.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Interface WordPress\Plugin_Check\Checker\With_Shared_Preparations
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+/**
+ * Interface for the shared preparations.
+ *
+ * @since n.e.x.t
+ */
+interface With_Shared_Preparations {
+
+	/**
+	 * Get list of preparation classes.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Returns an array of preparation class names.
+	 */
+	public function get_shared_preparations();
+}
--- a/includes/Checker/With_Shared_Preparations.php
+++ b/includes/Checker/With_Shared_Preparations.php
@@ -8,18 +8,21 @@
 namespace WordPress\Plugin_Check\Checker;
 
 /**
- * Interface for the shared preparations.
+ * Interface for a check that relies on shared preparations.
  *
  * @since n.e.x.t
  */
 interface With_Shared_Preparations {
 
 	/**
-	 * Get list of preparation classes.
+	 * Gets the list of shared preparations.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return array Returns an array of preparation class names.
+	 * @return array Returns a map of $class_name => $constructor_args pairs.
+	 *
+	 * NOTE: The expected return value here a map of [class name] => [constructor arguments] pairs,
+	 *       If the class does not need any constructor arguments, it would just be an empty array.
 	 */
 	public function get_shared_preparations();
 }
--- a/includes/Checker/With_Shared_Preparations.php
+++ b/includes/Checker/With_Shared_Preparations.php
@@ -19,10 +19,8 @@ interface With_Shared_Preparations {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return array Returns a map of $class_name => $constructor_args pairs.
-	 *
-	 * NOTE: The expected return value here a map of [class name] => [constructor arguments] pairs,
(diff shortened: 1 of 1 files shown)


includes/Checker/With_Shared_Preparations.php
```diff
@@ -0,0 +1,25 @@
...
+ * @since n.e.x.t
+ */
+interface With_Shared_Preparations {
+
+	/**
+	 * Get list of preparation classes.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Returns an array of preparation class names.
```
felixarntz: I think we need the return description here to expand a bit more: This is not captured in the ACs, so apologies for missing that. @jjgrainger and I had talked about that we also need to support a way to provide constructor arguments here. For example, if one of the class names requires constructor arguments, the consuming code has to know which ones should be used.

So I think we should maybe make the expected return value here a map of class name => constructor arguments pairs. If the class does not need any constructor arguments, it would just be an empty array.
```suggestion
	 * @return array Returns a map of $class_name => $constructor_args pairs.
```

includes/Checker/With_Shared_Preparations.php
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Interface for the shared preparations.
+ *
+ * @since n.e.x.t
+ */
+interface With_Shared_Preparations {
+
+	/**
+	 * Get list of preparation classes.
```
felixarntz: Per WP coding standards, functions/methods should start with a third-person verb. Also I suggest rephrasing a bit more generic since this should return more than just classes (see other comment).
```suggestion
	 * Gets the list of shared preparations.
```

includes/Checker/With_Shared_Preparations.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Interface WordPress\Plugin_Check\Checker\With_Shared_Preparations
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+/**
+ * Interface for the shared preparations.
```
felixarntz: This needs to be clarified:
```suggestion
 * Interface for a check that relies on shared preparations.
```

felixarntz (commented):
@vishalkakadiya A few small things need to be adjusted before this is good to go.

includes/Checker/With_Shared_Preparations.php
vishalkakadiya: @felixarntz I updated this now and also added these details in the function doc comment. Thank you!

includes/Checker/With_Shared_Preparations.php
vishalkakadiya: Updated this, thank you!

includes/Checker/With_Shared_Preparations.php
vishalkakadiya: Updated this, thank you!

vishalkakadiya:
@felixarntz I have updated this now, thank you! 🙌

jjgrainger (approved):
Thanks @vishalkakadiya looks good 👍

includes/Checker/With_Shared_Preparations.php
```diff
@@ -0,0 +1,28 @@
...
+
+	/**
+	 * Gets the list of shared preparations.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Returns a map of $class_name => $constructor_args pairs.
+	 *
+	 * NOTE: The expected return value here a map of [class name] => [constructor arguments] pairs,
+	 *       If the class does not need any constructor arguments, it would just be an empty array.
```
felixarntz: There should not be regular description text following the `@param` or `@return` docs. Why not just make it part of it?
```suggestion
	 * @return array Returns a map of $class_name => $constructor_args pairs. If the class does not
	 *               need any constructor arguments, it would just be an empty array.
```

felixarntz (commented):
@vishalkakadiya One last thing, but I can quickly resolve this. All good otherwise, thanks!

felixarntz (approved):
@vishalkakadiya LGTM!

PR #68, opened 2023-01-31
Add Force_Single_Plugin_Preparation class

### LINKED ISSUE #58: Create Force_Single_Plugin_Preparation
## Description
Create a new `Force_Single_Plugin_Preparation` that will force the plugin to be checks as the only active plugin.

## Acceptance Criteria
- Class `Force_Single_Plugin_Preparation` should be created and exists at `includes/Checker/Preparations`
- Class `Force_Single_Plugin_Preparation` should extend `Preparation` interface
- Class should have a protected property `$plugin_basename`
- Class constructor accepts a single `string $plugin_basename` parameter
- Class contains a `prepare()` method that does the following
    - Check the plugin folder exists in the plugins directory using the `$plugin_basename` property
    - If the plugin does not exist, and exception is thrown
    - Ensures the plugin and the plugin check plugin are the only in the active list of plugins using the following filters:
        - `option_active_plugins`
        - `default_option_active_plugins`
    - Returns a cleanup function that removes the filters added
 - Class contains a public `filter_active_plugins()` method that accepts an `$active_plugins` parameter
    - The method checks if the `$this->plugin_basename` exists in the array, and if it does, ensures the `$plugin_basename` and plugin checker plugin name are only in the active plugins array.

## Test Coverage
- Test the `prepare` method throws and exception if the plugin does not exist in the filesystem.
- Test the `prepare` method prepares the environment by ensuring that the `get_option( 'active_plugins' )` array includes the `$plugin_basename`.
- Test the `filter_active_plugins` returns the `$plugin_basename` property and plugin checker basename in the array passed when it exist.
- Test the `prepare` method returns a callable cleanup function.

vishalkakadiya:
Adds the Force_Single_Plugin_Preparation class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/58

Closes https://github.com/10up/plugin-check/issues/58

### DIFF
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -59,6 +59,7 @@ class Force_Single_Plugin_Preparation implements Preparation {
 
 			throw new Exception(
 				sprintf(
+					// translators: plugin basename.
 					__( 'The plugin %s does not exists', 'plugin-check' ),
 					$this->plugin_basename
 				)
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -69,7 +69,6 @@ class Force_Single_Plugin_Preparation implements Preparation {
 
 		// Return the cleanup function.
 		return function() {
-
 			remove_filter( 'option_active_plugins', array( $this, 'filter_active_plugins' ) );
 			remove_filter( 'default_option_active_plugins', array( $this, 'filter_active_plugins' ) );
 		};
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -44,8 +44,6 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @global array $wp_theme_directories
-	 *
 	 * @return callable Cleanup function to revert any changes made here.
 	 *
 	 * @throws Exception Thrown when preparation fails.
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -56,9 +56,10 @@ class Force_Single_Plugin_Preparation implements Preparation {
 
 			throw new Exception(
 				sprintf(
-					// translators: plugin basename.
-					__( 'The plugin %s does not exists', 'plugin-check' ),
-					$this->plugin_basename
+					/* translators: 1: plugin basename, 2: error message */
+					__( 'Invalid plugin %1$s: %2$s', 'plugin-check' ),
+					$this->plugin_basename,
+					$valid_plugin->get_error_message()
 				)
 			);
 		}
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -48,7 +48,6 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 * @throws Exception Thrown when preparation fails.
 	 */
 	public function prepare() {
-
 		$valid_plugin = validate_plugin( $this->plugin_basename );
 
 		// Check if the plugin exists.
@@ -78,16 +77,14 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 * Filter active plugins.
 	 *
 	 * @param array $active_plugins List of active plugins.
-	 *
 	 * @return array List of active plugins.
 	 */
-	public function filter_active_plugins( $active_plugins = array() ) {
-
-		if ( in_array( $this->plugin_basename, $active_plugins, true ) ) {
+	public function filter_active_plugins( $active_plugins ) {
+		if ( is_array( $active_plugins ) && in_array( $this->plugin_basename, $active_plugins, true ) ) {
 
 			return array(
 				$this->plugin_basename,
-				'plugin-check/plugin-check.php', // At the moment it is added static, we can update this with constant.
+				plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ),
 			);
 		}
 
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -82,19 +82,13 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	public function filter_active_plugins( $active_plugins ) {
 		if ( is_array( $active_plugins ) && in_array( $this->plugin_basename, $active_plugins, true ) ) {
 
-			$plugin_check_base_file = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
-
-			if ( $this->plugin_basename === $plugin_check_base_file ) {
-
-				return array(
-					$this->plugin_basename,
-				);
-			}
-
-			return array(
+			$plugins = array(
 				$this->plugin_basename,
 				plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ),
 			);
+
+			// Make sure no duplicate plugin files send in response.
+			return array_unique( $plugins );
 		}
 
 		return $active_plugins;
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -32,7 +32,7 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param string $plugin_basename Slug of the plugin.
+	 * @param string $plugin_basename Slug of the plugin, E.g. "akismet\akismet.php".
 	 */
 	public function __construct( $plugin_basename ) {
 
@@ -51,23 +51,16 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 * @throws Exception Thrown when preparation fails.
 	 */
 	public function prepare() {
-		// Override the theme slug and name.
-		add_filter( 'template', array( $this, 'get_theme_slug' ) );
-		add_filter( 'stylesheet', array( $this, 'get_theme_slug' ) );
-		add_filter( 'pre_option_template', array( $this, 'get_theme_slug' ) );
-		add_filter( 'pre_option_stylesheet', array( $this, 'get_theme_slug' ) );
-		add_filter( 'pre_option_current_theme', array( $this, 'get_theme_name' ) );
 
-		// Override the theme directory.
-		add_filter( 'pre_option_template_root', array( $this, 'get_theme_root' ) );
-		add_filter( 'pre_option_stylesheet_root', array( $this, 'get_theme_root' ) );
+		// Check if the plugin exists.
+		if ( validate_plugin( $this->plugin_basename ) ) {
 
-		// Register the custom themes directory if relevant.
-		if ( ! empty( $this->themes_dir ) ) {
-			register_theme_directory( $this->themes_dir );
-
-			// Force new directory scan to ensure the test theme directory is available.
-			search_theme_directories( true );
+			throw new Exception(
+				sprintf(
+					'Plugin is not exists at %1$s',
+					$this->plugin_basename
+				)
+			);
 		}
 
 		// Return the cleanup function.
@@ -94,5 +87,20 @@ class Force_Single_Plugin_Preparation implements Preparation {
 				}
 			}
 		};
+	}
+
+	/**
+	 * Filter active plugins.
+	 *
+	 * @param array $active_plugins List of active plugins.
+	 *
+	 * @return void
+	 */
+	public function filter_active_plugins( $active_plugins ) {
+
+		if ( in_array( $this->plugin_basename, $active_plugins, true ) ) {
+
+			// $active_plugins
+		}
 	}
 }
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -66,29 +66,14 @@ class Force_Single_Plugin_Preparation implements Preparation {
 			);
 		}
 
+		add_filter( 'option_active_plugins', array( $this, 'filter_active_plugins' ) );
+		add_filter( 'default_option_active_plugins', array( $this, 'filter_active_plugins' ) );
+
 		// Return the cleanup function.
 		return function() {
-			global $wp_theme_directories;
 
-			remove_filter( 'template', array( $this, 'get_theme_slug' ) );
-			remove_filter( 'stylesheet', array( $this, 'get_theme_slug' ) );
-			remove_filter( 'pre_option_template', array( $this, 'get_theme_slug' ) );
-			remove_filter( 'pre_option_stylesheet', array( $this, 'get_theme_slug' ) );
-			remove_filter( 'pre_option_current_theme', array( $this, 'get_theme_name' ) );
-
-			remove_filter( 'pre_option_template_root', array( $this, 'get_theme_root' ) );
-			remove_filter( 'pre_option_stylesheet_root', array( $this, 'get_theme_root' ) );
-
-			if ( ! empty( $this->themes_dir ) ) {
-				$index = array_search( untrailingslashit( $this->themes_dir ), $wp_theme_directories, true );
-				if ( false !== $index ) {
-					array_splice( $wp_theme_directories, $index, 1 );
-					$wp_theme_directories = array_values( $wp_theme_directories );
-
-					// Force new directory scan to remove the test theme directory.
-					search_theme_directories( true );
-				}
-			}
+			remove_filter( 'option_active_plugins', array( $this, 'filter_active_plugins' ) );
+			remove_filter( 'default_option_active_plugins', array( $this, 'filter_active_plugins' ) );
 		};
 	}
 
@@ -97,13 +82,18 @@ class Force_Single_Plugin_Preparation implements Preparation {
 	 *
 	 * @param array $active_plugins List of active plugins.
 	 *
-	 * @return void
+	 * @return array List of active plugins.
 	 */
-	public function filter_active_plugins( $active_plugins ) {
+	public function filter_active_plugins( $active_plugins = array() ) {
 
 		if ( in_array( $this->plugin_basename, $active_plugins, true ) ) {
 
-			// $active_plugins
+			return array(
+				$this->plugin_basename,
+				'plugin-check/plugin-check.php', // At the moment it is added static, we can update this with constant.
+			);
 		}
+
+		return $active_plugins;
 	}
 }
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -11,7 +11,7 @@ use WordPress\Plugin_Check\Checker\Preparation;
 use Exception;
 
 /**
- * Class for the preparation step to force the plugin to be checks as the only active plugin.
+ * Class for the preparation to force the plugin to be checked as the only active plugin.
  *
  * This ensures the plugin is checked as much in isolation as possible.
  *
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -1,0 +1,98 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Preparations\Force_Single_Plugin_Preparation
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Preparations;
+
+use WordPress\Plugin_Check\Checker\Preparation;
+use Exception;
+
+/**
+ * Class for the preparation step to force the plugin to be checks as the only active plugin.
+ *
+ * This ensures the plugin is checked as much in isolation as possible.
+ *
+ * @since n.e.x.t
+ */
+class Force_Single_Plugin_Preparation implements Preparation {
+
+	/**
+	 * Plugin slug.
+	 *
+	 * @since n.e.x.t
+	 * @var string
+	 */
+	protected $plugin_basename;
+
+	/**
+	 * Sets the plugin slug.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_basename Slug of the plugin.
+	 */
+	public function __construct( $plugin_basename ) {
+
+		$this->plugin_basename = $plugin_basename;
+	}
+
+	/**
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @global array $wp_theme_directories
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		// Override the theme slug and name.
+		add_filter( 'template', array( $this, 'get_theme_slug' ) );
+		add_filter( 'stylesheet', array( $this, 'get_theme_slug' ) );
+		add_filter( 'pre_option_template', array( $this, 'get_theme_slug' ) );
+		add_filter( 'pre_option_stylesheet', array( $this, 'get_theme_slug' ) );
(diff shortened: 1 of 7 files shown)


includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,99 @@
...
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Preparations;
+
+use WordPress\Plugin_Check\Checker\Preparation;
+use Exception;
+
+/**
+ * Class for the preparation step to force the plugin to be checks as the only active plugin.
```
jjgrainger: Minor fix.

```suggestion
 * Class for the preparation to force the plugin to be checked as the only active plugin.
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,99 @@
...
+
+		// Check if the plugin exists.
+		if ( is_wp_error( $valid_plugin ) ) {
+
+			throw new Exception(
+				sprintf(
+					'%1$s %2$s',
+					__( 'Plugin is not exists at', 'plugin-check' ),
+					$this->plugin_basename
+				)
```
jjgrainger: @vishalkakadiya this isn't quite right and I've updated the text.

```suggestion
				sprintf(
					__( 'The plugin %s does not exists', 'plugin-check' ),
					$this->plugin_basename
				)
```

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,101 @@
...
+use WordPress\Plugin_Check\Checker\Preparations\Force_Single_Plugin_Preparation;
+use WP_UnitTestCase;
+use Exception;
+
+class Force_Single_Plugin_Preparation_Tests extends WP_UnitTestCase {
+
+	public function set_up() {
+
+		parent::set_up();
+	}
```
jjgrainger: @vishalkakadiya lets remove as this is unnecessary.

```suggestion
```

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,101 @@
...
+		try {
+
+			$preparation->prepare();
+
+		} catch ( Exception $e ) {
+
+			$message = $e->getMessage();
+		}
+
+		$this->assertEquals( 'Plugin is not exists at akismet/akismet.php', $message );
```
jjgrainger: @vishalkakadiya minor nit pick but there is a lot of spacing in the code that could be neatened up. We should do the same for other tests where theres a lot of spacing.

Also, the error message should be updated based on a previous comment.

```suggestion

		$preparation = new Force_Single_Plugin_Preparation( 'akismet/akismet.php' );
		$message     = '';

		try {
			$preparation->prepare();
		} catch ( Exception $e ) {
			$message = $e->getMessage();
		}

		$this->assertEquals( 'The plugin akismet/akismet.php does not exists', $message );
```

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,101 @@
...
+class Force_Single_Plugin_Preparation_Tests extends WP_UnitTestCase {
+
+	public function set_up() {
+
+		parent::set_up();
+	}
+
+	/**
+	 * @covers Force_Single_Plugin_Preparation::prepare().
+	 */
```
jjgrainger: @vishalkakadiya I don't think we need these docblock comments unless they're required. Let's remove them.

jjgrainger (changes requested):
Thanks @vishalkakadiya overall looking good, but left some minor feedback.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: @jjgrainger I have removed this now. Thank you!

vishalkakadiya:
Thank you! @jjgrainger

@felixarntz It is ready for your review now. Thank you!

vishalkakadiya:
@jjgrainger I have addressed your feedback, can you re-review it again. Thank you! 🙂

jjgrainger (approved):
Thanks @vishalkakadiya approved!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Sets the plugin slug.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_basename Slug of the plugin, E.g. "akismet\akismet.php".
+	 */
+	public function __construct( $plugin_basename ) {
+
```
felixarntz: Let's avoid these unnecessary blank lines on top of the method bodies (same applies below).
```suggestion
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+		// Check if the plugin exists.
+		if ( is_wp_error( $valid_plugin ) ) {
+
+			throw new Exception(
+				sprintf(
+					// translators: plugin basename.
+					__( 'The plugin %s does not exists', 'plugin-check' ),
+					$this->plugin_basename
+				)
```
felixarntz: Better to also provide the more detailed context from the `WP_Error`:
```suggestion
				sprintf(
					/* translators: 1: plugin basename, 2: error message */
					__( 'Invalid plugin %1$s: %2$s', 'plugin-check' ),
					$this->plugin_basename,
					$valid_plugin->get_error_message()
				)
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+			);
+		}
+
+		add_filter( 'option_active_plugins', array( $this, 'filter_active_plugins' ) );
+		add_filter( 'default_option_active_plugins', array( $this, 'filter_active_plugins' ) );
+
+		// Return the cleanup function.
+		return function() {
+
```
felixarntz: See above.
```suggestion
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+	 *
+	 * @return array List of active plugins.
+	 */
+	public function filter_active_plugins( $active_plugins = array() ) {
+
+		if ( in_array( $this->plugin_basename, $active_plugins, true ) ) {
+
+			return array(
+				$this->plugin_basename,
+				'plugin-check/plugin-check.php', // At the moment it is added static, we can update this with constant.
```
felixarntz: We should use the actual plugin basename here right away, otherwise, this is error-prone, since it is not guaranteed that the plugin is in the `plugin-check` folder.

How about the following:
* Add a constant `WP_PLUGIN_CHECK_MAIN_FILE` with `__FILE__` to the `plugin-check.php` main file.
* Change the code here to use `plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE )`.

This way it is ensured we use the correct plugin basename.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+
```
felixarntz: See above.
```suggestion
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * Filter active plugins.
+	 *
+	 * @param array $active_plugins List of active plugins.
+	 *
+	 * @return array List of active plugins.
+	 */
+	public function filter_active_plugins( $active_plugins = array() ) {
+
```
felixarntz, line 89: See above.
```suggestion
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+			remove_filter( 'default_option_active_plugins', array( $this, 'filter_active_plugins' ) );
+		};
+	}
+
+	/**
+	 * Filter active plugins.
+	 *
+	 * @param array $active_plugins List of active plugins.
+	 *
+	 * @return array List of active plugins.
```
felixarntz: `@param` and `@return` docs should always be together, without a blank line in between.
```suggestion
	 * @param array $active_plugins List of active plugins.
	 * @return array List of active plugins.
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public function filter_active_plugins( $active_plugins = array() ) {
```
felixarntz: Why is this parameter optional? I think we should make it required, it should always be passed.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+
+		if ( in_array( $this->plugin_basename, $active_plugins, true ) ) {
```
felixarntz: WordPress filters are not type-safe, so we should ensure `$active_plugins` is an array before we call an array function on it.
```suggestion
		if ( is_array( $active_plugins ) && in_array( $this->plugin_basename, $active_plugins, true ) ) {
```

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+		$this->assertEquals( 'The plugin akismet/akismet.php does not exists', $message );
+	}
+
+	public function test_prepare() {
+
+		$preparation = new Force_Single_Plugin_Preparation( 'plugin-check/plugin-check.php' );
+
+		$plugins = array(
+			'akismet/akismet.php',
+			'plugin-check/plugin-check.php',
```
felixarntz: Please use the actual plugin basename in all the tests here as well instead of hard-coding `plugin-check/plugin-check` (see my comment above).

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+
+		$cleanup = $preparation->prepare();
+
+		$active_plugins = get_option( 'active_plugins' );
+
+		$cleanup();
+
+		$this->assertIsCallable( $cleanup );
+
+		$this->assertContains( 'plugin-check/plugin-check.php', $active_plugins );
```
felixarntz: This assertion is redundant (already covered by the one below) and therefore should be removed.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+
+		update_option( 'active_plugins', $plugins );
+
+		$cleanup = $preparation->prepare();
+
+		$active_plugins = get_option( 'active_plugins' );
+
+		$cleanup();
+
+		$this->assertIsCallable( $cleanup );
```
felixarntz: This assertion is unnecessary - if `$cleanup` is not a callable, this code would have already failed in the call to it above.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+
+		$this->assertContains( 'plugin-check/plugin-check.php', $active_plugins );
+
+		$this->assertEquals(
+			array(
+				'plugin-check/plugin-check.php',
+				'plugin-check/plugin-check.php',
+			),
+			$active_plugins
+		);
```
felixarntz: Can you add another assertion here to ensure the filters have been cleaned up again? Basically call `get_option( 'active_plugins' )` again and assert that it has the original 3 plugin basenames in it.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,97 @@
...
+
+			return array(
+				$this->plugin_basename,
+				'plugin-check/plugin-check.php', // At the moment it is added static, we can update this with constant.
+			);
```
felixarntz: If `$this->plugin_basename` is the same as the `plugin-check` basename (effectively someone checking the code of the plugin check plugin itself), we need to only return that one item, otherwise this may result in a duplicate attempt to load the same plugin in core which would trigger a fatal error.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+
+		$active_plugins = get_option( 'active_plugins' );
+
+		$cleanup();
+
+		$this->assertIsCallable( $cleanup );
+
+		$this->assertContains( 'plugin-check/plugin-check.php', $active_plugins );
+
+		$this->assertEquals(
```
felixarntz: Nit-pick: Let's use `assertSame` to be as strict as possible.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+			array(
+				'plugin-check/plugin-check.php',
+				'plugin-check/plugin-check.php',
+			),
```
felixarntz: See my comment above, this should only be in here once, otherwise it would trigger a fatal error in core later.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -0,0 +1,81 @@
...
+		$preparation = new Force_Single_Plugin_Preparation( 'akismet/akismet.php' );
+		$message     = '';
+
+		try {
+			$preparation->prepare();
+		} catch ( Exception $e ) {
+			$message = $e->getMessage();
+		}
+
+		$this->assertEquals( 'The plugin akismet/akismet.php does not exists', $message );
```
felixarntz: Instead of this, we should use `expectException`, something like this:
```php
$this->expectException( 'Exception' );
$this->expectExceptionMessage( ... );
$preparation->prepare();
```

felixarntz (changes requested):
@vishalkakadiya This is going in the right direction. There are a few quirks with the production logic that need to be ironed out, but otherwise just suggestions to enhance the tests and a few nit-picks.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
vishalkakadiya: Fixed it, thank you!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
vishalkakadiya: Fixed it, thank you!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
vishalkakadiya: @felixarntz Yes, that works for us. Thank you! 🙌

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Updated this now, thank you!

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Updated this now, thank you!

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Updated this now.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
vishalkakadiya: Updated this now.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: @felixarntz I have added that test, but we don't need to set the `active_plugins` option for this. We can directly pass the plugin files that are not added while `Force_Single_Plugin_Preparation` instantiating so it will return the actual files array we send - You can check code for that assertion here - https://github.com/10up/plugin-check/pull/68/files#diff-4a0736f42c4d3624b9741527c78372d58c249c9fa517ad3df1c15cb78307627cR87

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Removed it, thank you!

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Done

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: Done

vishalkakadiya:
@felixarntz I have addressed your feedback and left one comment, you can do the second round of review now. Thank you!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -0,0 +1,96 @@
...
+	public function filter_active_plugins( $active_plugins ) {
+		if ( is_array( $active_plugins ) && in_array( $this->plugin_basename, $active_plugins, true ) ) {
+
+			$plugins = array(
+				$this->plugin_basename,
+				plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ),
+			);
+
+			// Make sure no duplicate plugin files send in response.
+			return array_unique( $plugins );
```
felixarntz: This works, but is a little less performant than simply checking for whether `$this->plugin_basename === plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE )` and returning a different array in that case. Not a big deal though.

plugin-check.php
```diff
@@ -18,6 +18,7 @@
 define( 'WP_PLUGIN_CHECK_VERSION', 'n.e.x.t' );
 define( 'WP_PLUGIN_CHECK_MINIMUM_PHP', '5.6' );
 define( 'WP_PLUGIN_CHECK_PLUGIN_DIR_PATH', plugin_dir_path( __FILE__ ) );
+define( 'WP_PLUGIN_CHECK_MAIN_FILE', __FILE__ );
```
felixarntz, line 20: Would be great to use this new constant anywhere else where we refer to this file, e.g. line 20 and 46.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
felixarntz: That assertion you're referring to is not the same though. That assertion tests whether the `filter_active_plugins()` method works correctly.

I think we need another assertion in this method here, to assert that _after_ calling the `$cleanup()` method, the return value from `get_option( 'active_plugins' )` is the same as what we originally specified - not because of how `filter_active_plugins()` works, but because the filter callback should have been removed entirely.

felixarntz (commented):
@vishalkakadiya Almost good to go. One follow up point from my previous review that wasn't fully addressed, and two very small nit-picks - they wouldn't be blocking, but since the one test feedback needs to be addressed anyway, it would be great if you could address those two small things as well.

tests/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
vishalkakadiya: @felixarntz We have one issue/blocker with this situation: the `active_plugins` setting, defined (see below) in the `bootstrap.php` that always overrides the `active_plugins` option which we set from the individual tests functions. And it is always returning the `plugin-check/plugin-check.php` file as the single active plugin. I'm finding out the solutions to it, maybe via something like `tests_add_filter`, but at the moment, I'm not sure how this will be fixed.

```
$GLOBALS['wp_tests_options'] = array(
  'active_plugins' => array( basename( TESTS_PLUGIN_DIR ) . '/plugin-check.php' ),
);
```

CC @jjgrainger if you have any suggestions on it, thank you!

felixarntz: @vishalkakadiya Looking at the WP core test suite code should provide some clarity on how we can solve this. I think we need to simply remove that filter on top of this test method?

Maybe try this:
```php
remove_filter( 'pre_option_active_plugins', 'wp_tests_options' );
```

vishalkakadiya: @felixarntz Yeah, it is working fine with this solution. Thank you! 🙌 

@jjgrainger Thank you for committing that code. 🙇

plugin-check.php
vishalkakadiya: Done.

vishalkakadiya:
@felixarntz I have addressed the remaining feedback on it. It is ready for review now. Thank you!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
vishalkakadiya: Done

tests/Checker/Check_Context_Tests.php
```diff
@@ -12,11 +12,11 @@ public function set_up() {
 		parent::set_up();
 
 		$this->plugin_name   = basename( TESTS_PLUGIN_DIR );
-		$this->check_context = new Check_Context( WP_PLUGIN_DIR . '/' . $this->plugin_name . '/plugin-check.php' );
+		$this->check_context = new Check_Context( WP_PLUGIN_DIR . '/' . plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ) );
```
felixarntz: This code is a bit unnecessarily complicated now. It now relies on `plugin_basename()`, when really this is just about the main file.
```suggestion
		$this->check_context = new Check_Context( WP_PLUGIN_CHECK_MAIN_FILE );
```

tests/Plugin_Context_Tests.php
```diff
@@ -12,11 +12,11 @@ public function set_up() {
 		parent::set_up();
 
 		$this->plugin_name    = basename( TESTS_PLUGIN_DIR );
-		$this->plugin_context = new Plugin_Context( WP_PLUGIN_DIR . '/' . $this->plugin_name . '/plugin-check.php' );
+		$this->plugin_context = new Plugin_Context( WP_PLUGIN_DIR . '/' . plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ) );
```
felixarntz: See above.
```suggestion
		$this->plugin_context = new Plugin_Context( WP_PLUGIN_CHECK_MAIN_FILE );
```

tests/Plugin_Main_Tests.php
```diff
@@ -12,7 +12,7 @@ class Plugin_Main_Tests extends WP_UnitTestCase {
 	public function set_up() {
 		parent::set_up();
 
-		$this->plugin_main = new Plugin_Main( basename( dirname( __DIR__ ) ) . '/plugin-check.php' );
+		$this->plugin_main = new Plugin_Main( plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE ) );
```
felixarntz: While related to my suggestions above, this code here is actually incorrect (also was incorrect before): It needs to be the full path to the plugin main file, not the plugin base name (see parameter documentation).
```suggestion
		$this->plugin_main = new Plugin_Main( WP_PLUGIN_CHECK_MAIN_FILE );
```

felixarntz (approved):
@vishalkakadiya Pretty much LGTM, just 3 nit-picks that would be good to address. Feel free to merge afterwards without another review.

vishalkakadiya:
@felixarntz Thank you for the feedback, I have fixed those and merging this now.

PR #69, opened 2023-02-01
Add Universal_Runtime_Preparation class

### LINKED ISSUE #61: Create a Universal_Runtime_Preparation
## Description
Create a `Universal_Runtime_Preparation` to handle all preparations required for when at least one `Runtime_Check` is being run.

This Preparation depends on the following issues and should be started once they are complete.

- #56 
- #58 

## Acceptance Criteria
- Class `Universal_Runtime_Preparation` should be created and exists at `includes/Checker/Preparations/Universal_Runtime_Preparation.php`.
- Class `Universal_Runtime_Preparation` should extend `Preparation` interface.
- Class has a protected `$check_context` property.
- Class constructor accepts an instance of the `Check_Context` and assigns it to the `$check_context` property.
- Class contains a `prepare()` method that accepts no parameters and will do the following:
    - Instantiate the `Use_Minimal_Theme_Preparation` to use the `wp-empty-theme` in the Plugin Checker.
    - Instantiate the `Force_Single_Plugin_Preparation` to activate the plugin being checked.
    - Runs the `prepare()` method on the instances of both `Use_Minimal_Theme_Preparation` and `Force_Single_Plugin_Preparation`
    - Returns a cleanup function that loops over the cleanup functions returned by both the `Use_Minimal_Theme_Preparation` and `Force_Single_Plugin_Preparation` prepare methods.

## Test Coverage
- Test the `prepare` method returns a callable cleanup function.
- Test that `Use_Minimal_Theme_Preparation` was successfully prepared by testing `get_option( 'template' )` returns `wp-empty-theme` slug.
- Test that `Force_Single_Plugin_Preparation` was successfully prepared by testing `get_option( 'active_plugins' )` returns the plugins basename.

vishalkakadiya:
Adds the Universal_Runtime_Preparation class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/61

Closes https://github.com/10up/plugin-check/issues/61

### DIFF
--- a/includes/Checker/Preparations/Universal_Runtime_Preparation.php
+++ b/includes/Checker/Preparations/Universal_Runtime_Preparation.php
@@ -10,7 +10,6 @@ namespace WordPress\Plugin_Check\Checker\Preparations;
 use WordPress\Plugin_Check\Checker\Check_Context;
 use WordPress\Plugin_Check\Checker\Preparation;
 use Exception;
-use function PHPUnit\Framework\callback;
 
 /**
  * Class handle all preparations required for when at least one `Runtime_Check` is being run.
@@ -39,7 +38,11 @@ class Universal_Runtime_Preparation implements Preparation {
 	}
 
 	/**
-	 * Runs the preparations for the theme and plugin. And returns a cleanup function.
+	 * Runs preparation step for the environment by modifying the plugins and theme to use,
+	 * and returns a closure as a cleanup function.
+	 *
+	 * This preparation needs to be called very early in the WordPress lifecycle, before
+	 * plugins are loaded, e.g. from a drop-in like `object-cache.php`.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -61,8 +64,7 @@ class Universal_Runtime_Preparation implements Preparation {
 		return function () use ( $cleanup_functions ) {
 
 			foreach ( $cleanup_functions as $cleanup_function ) {
-
-				callback( $cleanup_function );
+				$cleanup_function();
 			}
 		};
 	}
--- a/includes/Checker/Preparations/Universal_Runtime_Preparation.php
+++ b/includes/Checker/Preparations/Universal_Runtime_Preparation.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Preparations;
+
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Preparation;
+use Exception;
+use function PHPUnit\Framework\callback;
+
+/**
+ * Class handle all preparations required for when at least one `Runtime_Check` is being run.
+ *
+ * @since n.e.x.t
+ */
+class Universal_Runtime_Preparation implements Preparation {
+
+	/**
+	 * Context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 * @var Check_Context
+	 */
+	protected $check_context;
+
+	/**
+	 * Sets the context for the plugin to check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Context $check_context Check context instance for the plugin.
+	 */
+	public function __construct( Check_Context $check_context ) {
+		$this->check_context = $check_context;
+	}
+
+	/**
+	 * Runs the preparations for the theme and plugin. And returns a cleanup function.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Cleanup function to revert changes made by theme and plugin preparation classes.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+
+		$cleanup_functions = array();
+
+		$use_minimal_theme_preparation = new Use_Minimal_Theme_Preparation( 'wp-empty-theme', WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/test-content/themes' );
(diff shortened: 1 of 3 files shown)


includes/Checker/Preparations/Universal_Runtime_Preparation.php
```diff
@@ -0,0 +1,69 @@
...
+
+		$force_single_plugin_preparation = new Force_Single_Plugin_Preparation( 'plugin-check/plugin-check.php' );
+		$cleanup_functions[]             = $force_single_plugin_preparation->prepare();
+
+		// Return the cleanup function.
+		return function () use ( $cleanup_functions ) {
+
+			foreach ( $cleanup_functions as $cleanup_function ) {
+
+				callback( $cleanup_function );
```
felixarntz: We can simplify this to just call the function in the variable. Also nit-pick, but no need for all the blank lines here.
```suggestion
			foreach ( $cleanup_functions as $cleanup_function ) {
				$cleanup_function();
```

includes/Checker/Preparations/Universal_Runtime_Preparation.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Context $check_context Check context instance for the plugin.
+	 */
+	public function __construct( Check_Context $check_context ) {
+		$this->check_context = $check_context;
+	}
+
+	/**
+	 * Runs the preparations for the theme and plugin. And returns a cleanup function.
```
felixarntz: Adding a bit of clarification here. Most importantly, we should explain that this needs to be run early.
```suggestion
	 * Runs preparation step for the environment by modifying the plugins and theme to use,
	 * and returns a closure as a cleanup function.
	 *
	 * This preparation needs to be called very early in the WordPress lifecycle, before
	 * plugins are loaded, e.g. from a drop-in like `object-cache.php`.
```

includes/Checker/Preparations/Universal_Runtime_Preparation.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+
+		$cleanup_functions = array();
+
+		$use_minimal_theme_preparation = new Use_Minimal_Theme_Preparation( 'wp-empty-theme', WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/test-content/themes' );
+		$cleanup_functions[]           = $use_minimal_theme_preparation->prepare();
+
+		$force_single_plugin_preparation = new Force_Single_Plugin_Preparation( 'plugin-check/plugin-check.php' );
```
felixarntz: This is not correct, it needs to be based on the plugin basename from the `Check_Context` provided in the constructor.

tests/Checker/Preparations/Universal_Runtime_Preparation_Tests.php
```diff
@@ -0,0 +1,46 @@
...
+
+		$cleanup = $universal_runtime_preparation->prepare();
+
+		$cleanup();
+
+		$this->assertEquals( 'wp-empty-theme', get_option( 'template' ) );
+
+		$active_plugins = get_option( 'active_plugins' );
+
+		$this->assertSame( array( $plugin_basename_file ), $active_plugins );
```
felixarntz: The test here isn't really testing what it should, but more importantly, since we already have detailed test coverage for `Use_Minimal_Theme_Preparation` and `Force_Single_Plugin_Preparation`, we can keep the test coverage here at a higher level. We can simply assert that the filters have been added correctly, e.g. something like this:
```php
$this->assertTrue( has_filter( 'option_active_plugins' ) );
$this->assertTrue( has_filter( 'default_option_active_plugins' ) );
$this->assertTrue( has_filter( 'stylesheet' ) );
$this->assertTrue( has_filter( 'template' ) );
```

We should then also assert that those same filters are _no longer_ set after running the cleanup.

includes/Checker/Preparations/Universal_Runtime_Preparation.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Preparations;
+
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Preparation;
+use Exception;
+use function PHPUnit\Framework\callback;
```
felixarntz: We should never import a function from test code in production code; also see my other comment, this is unnecessary.

felixarntz (changes requested):
@vishalkakadiya There are a few small things to fix here before we can merge this. Also the test can be simplified, since the detailed testing already happens for the two classes used within this class.

includes/Checker/Preparations/Universal_Runtime_Preparation.php
vishalkakadiya: Sorry, that was accidentally added. I have removed it now, thank you!

includes/Checker/Preparations/Universal_Runtime_Preparation.php
vishalkakadiya: Make sense, I have updated it now. Thank you!

tests/Checker/Preparations/Universal_Runtime_Preparation_Tests.php
vishalkakadiya: Make sense, I have updated it now. Thank you!

vishalkakadiya:
@felixarntz I have addressed all the feedback here, you can review it now. Thanks!

jjgrainger (approved):
Thanks @vishalkakadiya approved

felixarntz (approved):
Thank you for the updates @vishalkakadiya, LGTM!

PR #72, opened 2023-02-07
Add Plugin_Check_Command class

### LINKED ISSUE #30: Create the WP-CLI subcommand wp plugin check

vishalkakadiya:
Add Plugin_Check_Command class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/30

Closes https://github.com/10up/plugin-check/issues/30

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -121,14 +121,11 @@ class Plugin_Check_Command {
 
 		$assoc_args = $this->get_options( $assoc_args );
 
-		Plugin_Request_Utility::get_plugin_basename_from_input( $args[0] );
-		Plugin_Request_Utility::initialize_runner();
-
-		$cli_runner = Plugin_Request_Utility::get_runner();
-
 		try {
 
-			$result = $cli_runner->run();
+			Plugin_Request_Utility::initialize_runner();
+			$cli_runner = Plugin_Request_Utility::get_runner();
+			$result     = $cli_runner->run();
 
 		} catch ( Exception $error ) {
 
@@ -229,7 +226,10 @@ class Plugin_Check_Command {
 			'code',
 			'message',
 		);
-		$default_fields = wp_parse_args( $assoc_args['fields'], $default_fields );
+
+		if ( isset( $assoc_args['fields'] ) ) {
+			$default_fields = wp_parse_args( $assoc_args['fields'], $default_fields );
+		}
 
 		// If both errors and warnings are included, display the type of each result too.
 		if ( empty( $assoc_args['ignore_errors'] ) && empty( $assoc_args['ignore_warnings'] ) ) {
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -106,14 +106,22 @@ class Plugin_Check_Command {
 		// Get the CLI Runner.
 		$runner = Plugin_Request_Utility::get_runner();
 
+		// Create the runner if not already initialized.
 		if ( is_null( $runner ) ) {
 			$runner = new CLI_Runner();
-			$runner->set_plugin( $plugin );
-			$runner->set_check_slugs( $checks );
+		}
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof CLI_Runner ) ) {
+			WP_CLI::error(
+				__( 'CLI Runner was not initialized correctly.', 'plugin-check' )
+			);
 		}
 
 		// Run checks against the plugin.
 		try {
+			$runner->set_plugin( $plugin );
+			$runner->set_check_slugs( $checks );
 			$result = $runner->run();
 		} catch ( Exception $error ) {
 			WP_CLI::error( $error->getMessage() );
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -110,7 +110,7 @@ class Plugin_Check_Command {
 	 */
 	public function check( $args, $assoc_args ) {
 
-		$options          = $this->get_options( $assoc_args );
+		$assoc_args       = $this->get_options( $assoc_args );
 		$plugin_base_file = $this->get_plugin_from_args( $args );
 
 		$cli_runner = new CLI_Runner();
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -7,6 +7,7 @@
 
 namespace WordPress\Plugin_Check\CLI;
 
+use WordPress\Plugin_Check\Checker\CLI_Runner;
 use WordPress\Plugin_Check\Plugin_Context;
 use Exception;
 
@@ -112,13 +113,46 @@ class Plugin_Check_Command {
 		$options          = $this->get_options( $assoc_args );
 		$plugin_base_file = $this->get_plugin_from_args( $args );
 
+		$cli_runner = new CLI_Runner();
+
 		try {
 
-			// TODO: Call `run()` method of the `CLI_Runner` class.
+			$result = $cli_runner->run();
 
 		} catch ( Exception $error ) {
 
 			\WP_CLI::error( $error->getMessage() );
+		}
+
+		// Get errors and warnings from the results.
+		$errors = array();
+		if ( empty( $assoc_args['ignore_errors'] ) ) {
+			$errors = $result->get_errors();
+		}
+		$warnings = array();
+		if ( empty( $assoc_args['ignore_warnings'] ) ) {
+			$warnings = $result->get_warnings();
+		}
+
+		// Get formatter.
+		$formatter = $this->get_formatter( $assoc_args );
+
+		// Print the formatted results.
+		// Go over all files with errors first and print them, combined with any warnings in the same file.
+		foreach ( $errors as $file_name => $file_errors ) {
+			$file_warnings = array();
+			if ( isset( $warnings[ $file_name ] ) ) {
+				$file_warnings = $warnings[ $file_name ];
+				unset( $warnings[ $file_name ] );
+			}
+			$file_results = $this->flatten_file_results( $file_errors, $file_warnings );
+			$this->display_results( $formatter, $file_name, $file_results );
+		}
+
+		// If there are any files left with only warnings, print those next.
+		foreach ( $warnings as $file_name => $file_warnings ) {
+			$file_results = $this->flatten_file_results( array(), $file_warnings );
+			$this->display_results( $formatter, $file_name, $file_results );
 		}
 	}
 
@@ -219,20 +253,103 @@ class Plugin_Check_Command {
 	}
 
 	/**
-	 * Get formatter class.
+	 * Gets the formatter instance to format check results.
 	 *
-	 * @param array $assoc_args List of the associative arguments.
+	 * @since 1.0.0
+	 *
+	 * @param array $assoc_args Associative arguments.
+	 * @return \WP_CLI\Formatter The formatter instance.
 	 */
 	protected function get_formatter( $assoc_args ) {
+
+		$default_fields = array(
+			'line',
+			'column',
+			'code',
+			'message',
+		);
+
+		// If both errors and warnings are included, display the type of each result too.
+		if ( empty( $assoc_args['ignore_errors'] ) && empty( $assoc_args['ignore_warnings'] ) ) {
+			$default_fields = array(
+				'line',
+				'column',
+				'type',
+				'code',
+				'message',
+			);
+		}
+
+		return new \WP_CLI\Formatter(
+			$assoc_args,
+			$default_fields
+		);
 	}
 
 	/**
-	 * Combining a files errors and warning into a single array and order them by file number.
+	 * Flattens and combines the given associative array of file errors and file warnings into a two-dimensional array.
 	 *
-	 * @param array $file_errors   List of errors.
-	 * @param array $file_warnings List of warnings.
+	 * @since 1.0.0
+	 *
+	 * @param array $file_errors   Errors from a Check_Result, for a specific file.
+	 * @param array $file_warnings Warnings from a Check_Result, for a specific file.
+	 * @return array Combined file results.
 	 */
 	protected function flatten_file_results( $file_errors, $file_warnings ) {
+		$file_results = array();
+
+		foreach ( $file_errors as $line => $line_errors ) {
+			foreach ( $line_errors as $column => $column_errors ) {
+				foreach ( $column_errors as $column_error ) {
+
+					$file_results[] = array_merge(
+						$column_error,
+						array(
+							'type'   => 'ERROR',
+							'line'   => $line,
+							'column' => $column,
+						)
+					);
+				}
+			}
+		}
+
+		foreach ( $file_warnings as $line => $line_warnings ) {
+			foreach ( $line_warnings as $column => $column_warnings ) {
+				foreach ( $column_warnings as $column_warning ) {
+
+					$file_results[] = array_merge(
+						$column_warning,
+						array(
+							'type'   => 'WARNING',
+							'line'   => $line,
+							'column' => $column,
+						)
+					);
+				}
+			}
+		}
+
+		usort(
+			$file_results,
+			function( $a, $b ) {
+				if ( $a['line'] < $b['line'] ) {
+					return -1;
+				}
+				if ( $a['line'] > $b['line'] ) {
+					return 1;
+				}
+				if ( $a['column'] < $b['column'] ) {
+					return -1;
+				}
+				if ( $a['column'] > $b['column'] ) {
+					return 1;
+				}
+				return 0;
+			}
+		);
+
+		return $file_results;
 	}
 
 	/**
@@ -243,6 +360,17 @@ class Plugin_Check_Command {
 	 * @param array             $file_results Results.
 	 */
 	protected function display_results( $formatter, $file_name, $file_results ) {
+		\WP_CLI::line(
+			sprintf(
+				'FILE: %s',
+				$file_name
+			)
+		);
+
+		$formatter->display_items( $file_results );
+
+		\WP_CLI::line();
+		\WP_CLI::line();
 	}
 
 	/**
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -50,17 +50,17 @@ class Plugin_Check_Command {
 	}
 
 	/**
-	 * Run plugin check.
+	 * Runs plugin check.
 	 *
 	 * ## OPTIONS
 	 *
-	 * [<plugin>]
+	 * <plugin>
 	 * : The plugin to check. Plugin name.
 	 *
-	 * [--checks]
+	 * [--checks=<checks>]
 	 * : Only runs checks provided as an argument in comma-separated values, e.g. enqueued-scripts, escaping. Otherwise runs all checks.
 	 *
-	 * [--format]
+	 * [--format=<format>]
 	 * : Format to display the results. Options are table, csv, and json. The default will be a table.
 	 * ---
 	 * default: table
@@ -70,7 +70,7 @@ class Plugin_Check_Command {
 	 *   - json
 	 * ---
 	 *
-	 * [--fields]
+	 * [--fields=<fields>]
 	 * : Limit displayed results to a subset of fields provided.
 	 *
 	 * [--ignore-warnings]
@@ -90,7 +90,7 @@ class Plugin_Check_Command {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $args List of the positional arguments.
+	 * @param array $args       List of the positional arguments.
 	 * @param array $assoc_args List of the associative arguments.
 	 *
 	 * @throws Exception Throws exception.
@@ -152,7 +152,7 @@ class Plugin_Check_Command {
 	}
 
 	/**
-	 * Validate associative arguments.
+	 * Validates the associative arguments.
 	 *
 	 * @since n.e.x.t
 	 *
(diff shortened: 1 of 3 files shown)


vishalkakadiya:
@felixarntz @jjgrainger The unit tests are not yet completed, but the command's class is a bit large, so that you can review the CLI command code. I will complete the unit test for this soon. Thank you! 🙂

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,423 @@
...
+		if ( empty( $plugin_base_file ) ) {
+
+			\WP_CLI::error(
+				sprintf(
+				/* translators: 1: plugin basename */
+					__( '"%1$s" plugin not exists.', 'plugin-check' ),
+					$plugin_slug
+				)
+			);
+		}
```
felixarntz: Instead of this logic, we should use what has already been implemented in `Plugin_Request_Utility::get_plugin_basename_from_input()` (see https://github.com/10up/plugin-check/blob/trunk/includes/Utilities/Plugin_Request_Utility.php#L45).

Related: Note that `$args[0]` can be a plugin slug or plugin basename.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,423 @@
...
+
+		$cli_runner = new CLI_Runner();
+
+		try {
+
+			$result = $cli_runner->run();
+
+		} catch ( Exception $error ) {
+
+			\WP_CLI::error( $error->getMessage() );
```
felixarntz: Instead of always using the `\` prefix, why not import this class via `use WP_CLI`?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,423 @@
...
+	 *
+	 * Uses the same filter core uses in plugins.php to determine which plugins
+	 * should be available to manage through the WP_Plugins_List_Table class.
+	 *
+	 * @return array
+	 */
+	private function get_all_plugins() {
+
+		return apply_filters( 'all_plugins', get_plugins() );
+	}
```
felixarntz: See my above comment, those two methods won't be needed. Please keep in mind to not introduce public methods unnecessarily - especially here since it would mean that they then would become CLI commands, which we don't want.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,423 @@
...
+	 * @param array $args List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws Exception Throws exception.
+	 */
+	public function check( $args, $assoc_args ) {
+
+		$assoc_args = $this->get_options( $assoc_args );
+
+		$this->get_plugin_from_args( $args );
```
felixarntz: This doesn't do anything right now. We should probably put it into a variable, and then ensure that that plugin is actually being tested by comparing it to the plugin that the `CLI_Runner` contains.

But at this point we can't do that since that class doesn't expose the plugin tested anywhere. Let's just skip handling the plugin for now.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,423 @@
...
+
+		$cli_runner = new CLI_Runner();
```
felixarntz: This should not instantiate a new `CLI_Runner` but instead get it from `Plugin_Request_Utility`.

felixarntz (commented):
@vishalkakadiya Left some feedback. Overall this looks solid.

There are a few things that we need to reconsider and discuss outside of this PR, e.g. how to better handle setup of the runner classes. But we can address that in a separate follow up PR later once we have figured out the approach.

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Fixed this now, thanks!

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Fixed this now, thanks!

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Removed it now, thanks!

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Done

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Removed it now, thanks!

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,339 @@
...
+	 * @param array $args List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws Exception Throws exception.
+	 */
+	public function check( $args, $assoc_args ) {
+
+		$assoc_args = $this->get_options( $assoc_args );
+
+		Plugin_Request_Utility::get_plugin_basename_from_input( $args[0] );
```
jjgrainger: @vishalkakadiya this method returns the plugin file and we're currently not storing this as a variable or using it anywhere? Can we remove?

The `CLI_Runner` itself uses this to get the plugin file early in the process. Unless @felixarntz has reasons to be using this here, I don't believe it's needed as the plugin will be captured and validated by the runner class.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,339 @@
...
+		Plugin_Request_Utility::initialize_runner();
+
+		$cli_runner = Plugin_Request_Utility::get_runner();
```
jjgrainger: @vishalkakadiya related to the above, the runner class will throw an exception if the plugin to check is invalid. Maybe we should include both the `Plugin_Request_Utility::initialize_runner();` and `Plugin_Request_Utility::get_runner();` calls inside the following try/catch block so those exceptions can be handled properly.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,339 @@
...
+		if ( ! in_array( $options['flag'], $this->check_flags, true ) ) {
+
+			WP_CLI::error(
+				sprintf(
+					// translators: 1. Check flags.
+					__( 'Invalid flag argument, valid value will be one of [%1$s]', 'plugin-check' ),
+					implode( ', ', $this->check_flags )
+				)
+			);
+		}
```
jjgrainger: @vishalkakadiya I see we have this option to choose what level of check stability we can run (stable, experimental etc) though we do not use this is the `check` method here.

I will discuss with @felixarntz on what needs to be done here to implement this or whether we even want to implement this as part of milestone 1.

jjgrainger (changes requested):
Thanks @vishalkakadiya a couple of small changes here but there is 1 part of this work I want to discuss with @felixarntz on next steps too.

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Done, thanks!

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: Make sense, I have updated it now. Thanks!

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: @jjgrainger Sure, we can discuss this with @felixarntz. Meanwhile, I have fixed all your other feedback. Thanks!

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,339 @@
...
+				foreach ( $column_errors as $column_error ) {
+
+					$file_results[] = array_merge(
+						$column_error,
+						array(
+							'type'   => 'ERROR',
+							'line'   => $line,
+							'column' => $column,
+						)
+					);
```
joemcgill, line 255: Avoid using `array_merge` in loops because PHP will store a copy of the array in memory for each loop.
See: https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md

In this case, I think you should be able to use the spread operator to unpack the `$column_error` value directly. Same applies to warnings below.


```suggestion
					$file_results[] = array(
						...$column_error,
						'type'   => 'ERROR',
						'line'   => $line,
						'column' => $column,
					);
```

joemcgill (commented):
One suggestion, otherwise this is looking pretty good.

includes/CLI/Plugin_Check_Command.php
vishalkakadiya: @joemcgill Yes, and this is a good point. This way works for us, but the spread operator with an associative array will not work for older than PHP 8.1 versions. Currently, the `Requires PHP is 5.6` in the plugin's doc comment. We might need to discuss this before implementing this change, but thank you for pointing this out. 🙌 

@joemcgill @jjgrainger @felixarntz Let me know what's your thoughts here. Thanks! 🙂

vishalkakadiya:
@joemcgill I have addressed your feedback on this PR and left one comment. You can review it now, thanks!

includes/CLI/Plugin_Check_Command.php
joemcgill: You're right @vishalkakadiya, I was conflating argument unpacking with array unpacking in my example. On a second look, the way you're using `array_merge` here isn't too bad, considering that you're not merging with an array that grows overtime. There are probably optimizations to be made here, but this is likely fine for now.

includes/CLI/Plugin_Check_Command.php
jjgrainger: Thanks @vishalkakadiya

I've spoken @felixarntz and we plan to remove this functionality and add it in later as it requires additional changes to checks so we can tell what stability level they are. No need to update anything here, I'll tackle this later in the sprint once we finalise other changes.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,339 @@
...
+	);
+
+	/**
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Plugin_Context $plugin_context Plugin context.
+	 */
+	public function __construct( $plugin_context ) {
```
spacedmonkey: Worth adding a type hint here?

```
__construct( Plugin_Context  $plugin_context ) {
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,302 @@
...
+	 *
+	 * @throws Exception Throws exception.
+	 */
+	public function check( $args, $assoc_args ) {
+
+		$assoc_args = $this->get_options( $assoc_args );
+
+		try {
+			Plugin_Request_Utility::initialize_runner();
+			$cli_runner = Plugin_Request_Utility::get_runner();
```
felixarntz: The runner would typically already be initialized at this point, since that would happen in our `object-cache.php` drop-in.

We need to only call `get_runner()`, and then only if none is set yet, instantiate the `CLI_Runner` class manually (in that case there's no need to use `Plugin_Request_Utility`).

All that logic can happen outside the `try` clause I think, since it's not expected to throw exceptions.

Last but not least, we then also need to set the plugin parameter and the checks to run on the `CLI_Runner` instance.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,302 @@
...
+	 *
+	 * @param array $args List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws Exception Throws exception.
+	 */
+	public function check( $args, $assoc_args ) {
+
+		$assoc_args = $this->get_options( $assoc_args );
+
```
felixarntz, line 105: Additionally to parsing the `$assoc_args`, we also need to get the plugin tested via `$args[0]`. That will then need to be passed to the `CLI_Runner` instance.

felixarntz (changes requested):
@jjgrainger This is almost good to go, there are just a few critical things missing around the runner setup, which we recently prepared for via #95.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+				}
+				return 0;
+			}
+		);
+
+		return $file_results;
+	}
+
+	/**
+	 * Display results.
```
```suggestion
	 * Displays the results.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+				}
+				if ( $a['line'] > $b['line'] ) {
+					return 1;
+				}
+				if ( $a['column'] < $b['column'] ) {
+					return -1;
+				}
+				if ( $a['column'] > $b['column'] ) {
+					return 1;
+				}
```
mukeshpanchal27, line 290: Can we combine the conditions? WDYT?

```suggestion
				if ( ( $a['line'] < $b['line'] ) || ( $a['column'] < $b['column'] ) ) {
					return -1;
				}
				if ( ( $a['line'] > $b['line'] ) || ( $a['column'] > $b['column'] ) ) {
					return 1;
				}
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param Plugin_Context $plugin_context Plugin context.
+	 */
+	public function __construct( Plugin_Context $plugin_context ) {
+		$this->plugin_context = $plugin_context;
+	}
+
+	/**
+	 * Run plugin check.
```
```suggestion
	 * Runs plugin check.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+	 *   wp plugin check akismet
+	 *   wp plugin check akismet --checks=escaping
+	 *   wp plugin check akismet --format=json
+	 *
+	 * @subcommand check
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $args List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
```
```suggestion
	 * @param array $args       List of the positional arguments.
	 * @param array $assoc_args List of the associative arguments.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @param Plugin_Context $plugin_context Plugin context.
+	 */
+	public function __construct( Plugin_Context $plugin_context ) {
+		$this->plugin_context = $plugin_context;
+	}
+
+	/**
+	 * Run plugin check.
+	 *
+	 * ## OPTIONS
```
mukeshpanchal27, line 55: @jjgrainger Is the options argument working for you? or will it be in an upcoming PR?

`npm run wp-env run cli wp plugin check akismet --format=json` it give me an error `Error: Too many positional arguments: json`

mukeshpanchal27 (commented):
Thanks @jjgrainger, I left some minor update feedback, one of which related to the option in the command not working.

includes/CLI/Plugin_Check_Command.php
jjgrainger: Thanks @mukeshpanchal27 I don't think we can as we want to sort by line first and then column. So if we combine them I believe this could result in ordering items incorrectly.

includes/CLI/Plugin_Check_Command.php
jjgrainger: Thanks @mukeshpanchal27 I experienced the same issue here but am uncertain what the cause is. I noticed that the docblock comment for the command wasn't correct for these options, however updating these has not resolved the issue. I'll keep looking into this.

felixarntz: Since you're invoking WP-CLI in the container, I think you're missing an extra `--`. You need to use `npm run wp-env run cli -- wp plugin check akismet --format=json`

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+		$plugin = isset( $args[0] ) ? $args[0] : '';
+		$checks = wp_parse_list( $options['checks'] );
+
+		// Get the CLI Runner.
+		$runner = Plugin_Request_Utility::get_runner();
+
+		if ( is_null( $runner ) ) {
+			$runner = new CLI_Runner();
+			$runner->set_plugin( $plugin );
+			$runner->set_check_slugs( $checks );
```
felixarntz: These calls need to happen unconditionally, even if you're using the runner returned by `Plugin_Request_Utility::get_runner`. That is the entire reason why we implemented the validation in `Abstract_Check_Runner`.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -0,0 +1,312 @@
...
+	public function check( $args, $assoc_args ) {
+		// Get options based on the CLI arguments.
+		$options = $this->get_options( $assoc_args );
+
+		// Create the plugin and checks array from CLI arguments.
+		$plugin = isset( $args[0] ) ? $args[0] : '';
+		$checks = wp_parse_list( $options['checks'] );
+
+		// Get the CLI Runner.
+		$runner = Plugin_Request_Utility::get_runner();
```
felixarntz, line 107: Nit-pick: Not sure if this can realistically ever happen, but we should probably cater for it just in case: If this returns any object that is _not_ a `CLI_Runner`, we should error out. For example, imagine something goes wrong in the `object-cache.php` invoked logic and then somehow this returns an `AJAX_Runner` here. Extremely unlikely, but possible based on the method implementation, so we may want to cater for that.

includes/CLI/Plugin_Check_Command.php
felixarntz: +1, also combining clauses is not always good for readability. Seeing the comparisons one by one after each other is probably easier to parse.

felixarntz (approved):
@jjgrainger One more thing here, and one nit-pick, but should be quick to address. I'll give this my approval already since the tweaks should be trivial. Excited to see this shape!

joemcgill (approved):
I'm happy with the current iteration once @felixarntz's comments have been addressed.

jjgrainger:
Thanks all, I've updated the PR based on the latest feedback. Will merge this now

jjgrainger (approved):

PR #73, opened 2023-02-08
Create Abstract Check Runner & CLI_Runner classes

### LINKED ISSUE #14: Create CLI_Runner class
## Description 

The CLI runner class will implement the `Abstract_Check_Runner` and its methods to use Checks class to run checks for a plugin provided. CLI command class will invoke or use the instance of the CLI Runner class to use the methods of the Checks class to run the checks.

This issue is dependant on:
- #63 

## Acceptance Criteria 
- Class `CLI_Runner` should be created and exists at `includes/Checker`
- The `CLI_Runner` class extends the `Abstract_Check_Runner`
- The class contains a `is_plugin_check()` method.
    - The `is_plugin_check()` method accepts no parameters.
    - The `is_plugin_check()` method will use the global `$_REQUEST['argv']` array, and any other relevant data, to determine if the current request is for the Plugin Checker via the CLI by:
        - Checking the `$_REQUEST['argv']` contains the `wp plugin check` command (see #30) 
    - If the above conditions are correct, the method returns `true`
    - If the above conditions do not match, the method returns `false`.
- The class contains a protected `setup_checks()` method. 
    - The `setup_checks()` method accepts no parameters.
    - The `setup_checks()` method will pass the plugin name from the `$_REQUEST['argv']` to the `Checks` constructor and assign the object to the `$checks` property.
    - The `setup_checks()` method will check the `$_REQUEST['argv']` array of slugs for the checks to run. If this does not exist all checks will be run from the `Checks` `get_checks()` method.
     - The Check instances retrieved from the `Checks` class instance `get_checks()` and assigned to the `$checks_to_run` property.
- The class contains a `run()` method.
    - The `run()` method accepts no parameters.
    - The `run()` method will call the `Checks` class `run_checks()` method passing the `$checks_to_run` property. (`$this->checks->run_checks( $this->checks_to_run )`)
    - The `run()` method will return the instance of `Check_Results` returned by the `run_checks()` method.

## Test Coverage

- Test `is_plugin_check()` returns `true` when the `$_REQUEST['argv']` array contains the expected values for a Plugin Checker CLI request.
- Test `is_plugin_check()` returns `false` when the `$_REQUEST['argv']` array does not contains the expected values for a Plugin Checker CLI request.
- The `requires_universal_preperations` returns false when no Checks passed implements the `Runtime_Check` interface
- The `requires_universal_preperations` returns true when one or more Checks passed implements the `Runtime_Check` interface
- Test `prepare()` method returns `null` when the checks to run does not contain a `Check` that implements the `Runtime_Check` interface.
- Test `prepare()` method returns a `callable` cleanup function when the checks to run contains a `Check` that implements the `Runtime_Check` interface.
- Test `run()` method returns an instance of `Check_Results`.
- Test `Check_Results` returned by `run()` contains no errors/warnings when running the test `Empty_Check`
- Test `Check_Results` returned by `run()` contains errors when running the test `Error_Check`

### LINKED ISSUE #63: Create Abstract_Check_Runner class
## Description

The Abstract_Check_Runner class will contain common logic used to run Checks across both CLI and AJAX. The abstract class will be used to build out the following runner classes:

- #14 
- #15 

## Acceptance Criteria

- A new `Abstract_Check_Runner` class will exist in `includes/Checker`.
- The class will implement the `Check_Runner` interface.
- The class has a protected `$checks` property.
- The class has a protected `$checks_to_run` property.
- The class has an abstract `is_plugin_check()` method.
- The class has an abstract protected `setup_checks()` method.
- The class contains a `prepare()` method. 
    - The `prepare()` method accepts no parameters.
    - The `prepare()` method is used to run any universal preparations early in the WordPress load.
        - The array of `$checks_to_run` array is passed to the `requires_universal_preparations()` method (#63). 
        - If the method returns true the `Universal_Runtime_Preparation` will be used.
        - The prepare method will instantiate the `Universal_Runtime_Preparation` (#61) and call the `prepare()` method.
        - The cleanup function returned by the `prepare()` method is then returned.
- The class has a protected `requires_universal_preparations( array $checks )` method.
    - The method accepts an array of Check instances that will be run.
    - The method loops over the Check instances and tests to see if any Checks implement the `Runtime_Check` interface.
    - If any Check implements the `Runtime_Check` interface, the method returns `true`, if not, the method returns `false`.
- The class implements a private `get_shared_preparations( array $checks )` method.
    - The method accepts an array of Check instances that will be run.
    - The method assigns an empty array to a `$preparations` variable.
    - The method loops of the Check instants and tests for Checks that implement the `With_Shared_Preparations` interface.
    - If the Check implements the `With_Shared_Preparations` interface, the Checks `get_shared_preparations ` method is called to return an array of preparations.
    - The preparations returned are checked for any duplicates by evaluating the preparation name and the array of constructor arguments to see if these match any existing shared preparations in the `$preparations` array.
    - These preparations are merged into the `$preparations` variable and returned by the method.
- The class implements the `run()` method.
    - The `run` method will gather the Check instances from `$this->checks->get_checks()` using the requested checks in the `$checks_to_run` property.
     - The `get_shared_preparations` is called passing the `$checks_to_run` array. The returned shared preparations are then looped over calling each preparations `prepare()` method. The preparations cleanup functions are then stored to a `$cleanups` variable.
     - The Checks are then run by passing the `$checks_to_run` array is then passed to `$this->checks->run_checks()` and the `Check_Results` instanced returned stored as a variable.
     - If the `get_shared_preparations` preparations where prepared the `$cleanups` functions returned are called.
     - The `run` method returns the `$check_results` instance.

jjgrainger:
Add Abstract_Check_Runner class

- Addresses the acceptance criteria in #63 
- Addresses the acceptance criteria in #14 

Closes #63, closes #14

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -144,4 +144,22 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 
 		return array_values( $shared_preparations );
 	}
+
+	/**
+	 * Returns the Check instances to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances.
+	 */
+	protected function get_checks_to_run() {
+		$check_slugs = $this->get_check_slugs_to_run();
+		$all_checks  = $this->get_checks_instance()->get_checks();
+
+		if ( empty( $checks ) ) {
+			return $all_checks;
+		}
+
+		return array_intersect_key( $all_checks, array_flip( $check_slugs ) );
+	}
 }
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -19,44 +19,40 @@ use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
 abstract class Abstract_Check_Runner implements Check_Runner {
 
 	/**
-	 * Instance of the Checks class.
-	 *
-	 * @since n.e.x.t
-	 * @var Checks
-	 */
-	protected $checks;
-
-	/**
-	 * Instance of the Check_Context class.
-	 *
-	 * @since n.e.x.t
-	 * @var Check_Context
-	 */
-	protected $context;
-
-	/**
-	 * Array of Check instances to run.
-	 *
-	 * @since n.e.x.t
-	 * @var array
-	 */
-	protected $checks_to_run;
-
-	/**
 	 * Determines if the current request is intended for the plugin checker.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return boolean Returns true if the check is for plugin else false.
+	 * @return bool Returns true if the check is for plugin else false.
 	 */
 	abstract public function is_plugin_check();
 
 	/**
-	 * Sets up the check context, checks and preparations based on the request.
+	 * Creates and returns an instance of the Checks class based on the request.
 	 *
 	 * @since n.e.x.t
+	 *
+	 * @return Checks An instances of the Checks class.
 	 */
-	abstract protected function setup_checks();
+	abstract protected function get_checks();
+
+	/**
+	 * Creates and returns the Check_Context for the plugin to check based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context The Check_Context for the plugin to check.
+	 */
+	abstract protected function get_context();
+
+	/**
+	 * Creates and returns an array of Check instances to run based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances.
+	 */
+	abstract protected function get_checks_to_run();
 
 	/**
 	 * Prepares the environment for running the requested checks.
@@ -68,12 +64,12 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @throws Exception Thrown exception when preparation fails.
 	 */
 	public function prepare() {
-		if ( ! $this->requires_universal_preparations( $this->checks_to_run ) ) {
-			$preparation = new Universal_Runtime_Preparation( $this->context );
+		if ( ! $this->requires_universal_preparations( $this->get_checks_to_run() ) ) {
+			$preparation = new Universal_Runtime_Preparation( $this->get_context() );
 			return $preparation->prepare();
 		}
 
-		return null;
+		return function() {};
 	}
 
 	/**
@@ -84,8 +80,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @return Check_Result
 	 */
 	public function run() {
-		$checks = $this->checks->get_checks( $this->checks_to_run );
-
+		$checks       = $this->get_checks_to_run();
 		$preparations = $this->get_shared_preparations( $checks );
 		$cleanups     = array();
 
@@ -94,7 +89,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 			$cleanups[] = $instance->prepare();
 		}
 
-		$results = $this->checks->run_checks( $checks );
+		$results = $this->get_checks()->run_checks( $checks );
 
 		if ( ! empty( $cleanups ) ) {
 			foreach ( $cleanups as $cleanup ) {
@@ -111,8 +106,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @since n.e.x.t
 	 *
 	 * @param array $checks An array of check instances to run.
-	 *
-	 * @return boolean Returns true if one or more checks requires the universal runtime preparation.
+	 * @return bool Returns true if one or more checks requires the universal runtime preparation.
 	 */
 	protected function requires_universal_preparations( array $checks ) {
 		foreach ( $checks as $check ) {
@@ -130,7 +124,6 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @since n.e.x.t
 	 *
 	 * @param array $checks An array of Check instances to run.
-	 *
 	 * @return array An array of reparations to run.
 	 */
 	private function get_shared_preparations( array $checks ) {
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -1,0 +1,145 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Abstract_Check_runner
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+use WordPress\Plugin_Check\Checker\Check_Runner;
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+/**
+ * Abstract Check Runner class.
+ *
+ * @since n.e.x.t
+ */
+abstract class Abstract_Check_Runner implements Check_Runner {
+
+	/**
+	 * Instance of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 * @var Checks
+	 */
+	protected $checks;
+
+	/**
+	 * Array of Check instances to run.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	protected $checks_to_run;
+
+	/**
+	 * Determines if the current request is intended for the plugin checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return boolean Returns true if the check is for plugin else false.
+	 */
+	abstract public function is_plugin_check();
+
+	/**
+	 * Setup checks and preparations based on the request.
+	 *
+	 * @since n.e.x.t
+	 */
+	abstract protected function setup_checks();
+
+	/**
+	 * Prepare the environment for running the requested checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown exception when preparation fails.
+	 */
+	public function prepare() {
+		if ( ! $this->requires_universal_preparations( $this->checks_to_run ) ) {
+			$preparation = new Universal_Runtime_Preparation();
+			return $preparation->prepare();
+		}
+
+		return null;
+	}
+
+	/**
+	 * Determine if any of the checks requires the universal runtime preparation.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of check instances to run.
+	 *
+	 * @return boolean Returns true if one or more checks requires the universal runtime preparation.
+	 */
+	protected function requires_universal_preparations( array $checks ) {
+		foreach ( $checks as $check ) {
+			if ( $check instanceof Runtime_Check ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * Return all shared preparations used by the checks to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of check instances to run.
+	 *
+	 * @return array An array of preparations to run.
+	 */
+	private function get_shared_preparations( array $checks ) {
+		$shared_preparations = array();
+
+		foreach ( $checks as $check ) {
+			if ( ! $check instanceof With_Shared_Preparations ) {
+				continue;
+			}
+
+			$preparations = $check->get_shared_preparations();
+
+			foreach ( $preparations as $class => $parameters ) {
+				if ( ! isset( $shared_preparations[ $class ] ) ) {
+					$shared_preparations[ $class ] = $parameters;
+				}
+			}
+		}
+
+		return $shared_preparations;
+	}
+
+	/**
+	 * Run the checks against the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Result
+	 */
+	public function run() {
+		$checks = $this->checks->get_checks( $this->checks_to_run );
+
+		$preparations = $this->get_shared_preparations( $checks );
+		$cleanups     = array();
+
+		foreach ( $preparations as $preparation ) {
+			$cleanups[] = $preparation->prepare();
+		}
+
+		$results = $this->checks->run_checks( $checks );
+
+		if ( ! empty( $cleanups ) ) {
+			foreach ( $cleanups as $cleanup ) {
+				$cleanup();
+			}
+		}
+
+		return $results;
+	}
+}
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -35,15 +35,6 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	abstract protected function get_checks();
 
 	/**
-	 * Creates and returns the Check_Context for the plugin to check based on the request.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return Check_Context The Check_Context for the plugin to check.
-	 */
-	abstract protected function get_context();
-
-	/**
 	 * Creates and returns an array of Check instances to run based on the request.
 	 *
 	 * @since n.e.x.t
@@ -63,7 +54,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	public function prepare() {
 		if ( $this->requires_universal_preparations( $this->get_checks_to_run() ) ) {
-			$preparation = new Universal_Runtime_Preparation( $this->get_context() );
+			$preparation = new Universal_Runtime_Preparation( $this->get_checks()->context() );
 			return $preparation->prepare();
 		}
 
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -156,7 +156,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 		$check_slugs = $this->get_check_slugs_to_run();
 		$all_checks  = $this->get_checks_instance()->get_checks();
 
-		if ( empty( $checks ) ) {
+		if ( empty( $check_slugs ) ) {
 			return $all_checks;
 		}
 
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -7,7 +7,7 @@
 
 namespace WordPress\Plugin_Check\Checker;
 
-use Preparations\Universal_Runtime_Preparation;
+use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
 
 /**
  * Abstract Check Runner class.
@@ -62,7 +62,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @throws Exception Thrown exception when preparation fails.
 	 */
 	public function prepare() {
-		if ( ! $this->requires_universal_preparations( $this->get_checks_to_run() ) ) {
+		if ( $this->requires_universal_preparations( $this->get_checks_to_run() ) ) {
 			$preparation = new Universal_Runtime_Preparation( $this->get_context() );
 			return $preparation->prepare();
 		}
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -7,9 +7,7 @@
 
 namespace WordPress\Plugin_Check\Checker;
 
-use WordPress\Plugin_Check\Checker\Check_Runner;
-use WordPress\Plugin_Check\Checker\Check_Result;
-use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
+use Preparations\Universal_Runtime_Preparation;
 
 /**
  * Abstract Check Runner class.
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -9,6 +9,7 @@ namespace WordPress\Plugin_Check\Checker;
 
 use WordPress\Plugin_Check\Checker\Check_Runner;
 use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
 
 /**
  * Abstract Check Runner class.
@@ -24,6 +25,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @var Checks
 	 */
 	protected $checks;
+
+	/**
+	 * Instance of the Check_Context class.
+	 *
+	 * @since n.e.x.t
+	 * @var Check_Context
+	 */
+	protected $context;
 
 	/**
 	 * Array of Check instances to run.
@@ -43,7 +52,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	abstract public function is_plugin_check();
 
 	/**
-	 * Setup checks and preparations based on the request.
+	 * Setup the check context, checks and preparations based on the request.
 	 *
 	 * @since n.e.x.t
 	 */
@@ -60,7 +69,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	public function prepare() {
 		if ( ! $this->requires_universal_preparations( $this->checks_to_run ) ) {
-			$preparation = new Universal_Runtime_Preparation();
+			$preparation = new Universal_Runtime_Preparation( $this->context );
 			return $preparation->prepare();
 		}
 
@@ -91,9 +100,9 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $checks An array of check instances to run.
+	 * @param array $checks An array of Check instances to run.
 	 *
-	 * @return array An array of preparations to run.
+	 * @return array An array of reparations to run.
 	 */
 	private function get_shared_preparations( array $checks ) {
 		$shared_preparations = array();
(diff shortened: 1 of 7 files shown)


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 * @return boolean Returns true if the check is for plugin else false.
+	 */
+	abstract public function is_plugin_check();
+
+	/**
+	 * Setup the check context, checks and preparations based on the request.
+	 *
+	 * @since n.e.x.t
+	 */
+	abstract protected function setup_checks();
```
felixarntz: Shouldn't this method be called somewhere in here (e.g. at the beginning of `run()`)? Otherwise, we shouldn't require an abstract method if calling that method as well would be responsibility of the child class.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+
+	/**
+	 * Run the checks against the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Result
+	 */
+	public function run() {
+		$checks = $this->checks->get_checks( $this->checks_to_run );
```
felixarntz: Related to the above, `$this->checks` will be undefined if `setup_checks()` is not called, or implemented incorrectly.

Seeing this, we should probably avoid the `$this->checks` and `$this->checks_to_run` class properties entirely. It would be a much clearer API to have e.g. `setup_checks()` _return_ the `Checks` instance rather than implicitly setting it in a class property, as that would then be clarified already by the method documentation.

Not sure how to best handle `$this->checks_to_run` - maybe introduce another abstract protected method like `get_checks_to_run()` that returns that list of checks to run?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 *
+	 * @return boolean Returns true if the check is for plugin else false.
+	 */
+	abstract public function is_plugin_check();
+
+	/**
+	 * Setup the check context, checks and preparations based on the request.
+	 *
+	 * @since n.e.x.t
+	 */
```
felixarntz, line 43: See my other comment, right now the documentation here doesn't make it clear what exactly this method needs to do. By avoiding the class properties here and instead requiring a specific return value, we can naturally document that via `@return Checks ...` which will make this a lot more intuitive.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	protected $context;
+
+	/**
+	 * Array of Check instances to run.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	protected $checks_to_run;
```
felixarntz: See my other comments, I think it's cleaner to avoid all of those properties. A child class can use these properties if needed, but it shouldn't have to. A more declarative API would be to introduce methods with specific return values, each method effectively replacing one of these class properties (`setup_checks()` would be the one just for the `Checks` instance).

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Class WordPress\Plugin_Check\Checker\Abstract_Check_runner
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+use WordPress\Plugin_Check\Checker\Check_Runner;
+use WordPress\Plugin_Check\Checker\Check_Result;
```
felixarntz: No need to import these as they are in the same namespace. It's also okay to have them here, but either way we should be consistent about it. For example, you're not including the `Checks` class here, which is actually just as relevant since it is referred to in a doc-block.

So either we should remove those `use` statements and always only add those that are from another namespace, or we should include _all_ of them.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
```
felixarntz, line 10: Nit-pick, you can simplify this as a relative import I think.
```suggestion
use Preparations\Universal_Runtime_Preparation;
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 *
+	 * @throws Exception Thrown exception when preparation fails.
+	 */
+	public function prepare() {
+		if ( ! $this->requires_universal_preparations( $this->checks_to_run ) ) {
+			$preparation = new Universal_Runtime_Preparation( $this->context );
+			return $preparation->prepare();
+		}
+
+		return null;
```
felixarntz: This is an invalid return value. Per the doc block, a function has to be returned, so please use an empty closure.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+
+				// Add to the shared preparations if the shared preparation is unique.
+				if ( $unique ) {
+					$shared_preparations[] = array(
+						'class' => $class,
+						'args'  => $args,
+					);
+				}
+			}
+		}
```
felixarntz: I think we can simplify this quite a bit by using array indexes alone to ensure uniqueness. For example, you could use something like `$class . '::' . md5( $args )` as a key for each entry. Then all that complex logic here won't be needed, since by nature duplicates would be avoided as they'd have the same key.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+
+		// Map over the shared preparations and instantiate the classes.
+		$shared_preparations = array_map(
+			function( $preparation ) {
+				return new $preparation['class']( ...$preparation['args'] );
+			},
+			$shared_preparations
+		);
```
felixarntz: Let's avoid this here. We should have this method still return just the array with `class` and `args` that you have above, since it's best to only initialize classes "lazily", once they're actually used, to avoid unnecessary instantiation. So essentially you can instantiate each shared preparation in the `foreach` loop in the `run()` method right before calling its respective `prepare()` method.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 * @var array
+	 */
+	protected $checks_to_run;
+
+	/**
+	 * Determines if the current request is intended for the plugin checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return boolean Returns true if the check is for plugin else false.
```
felixarntz: Per WP docs standards:
```suggestion
	 * @return bool Returns true if the check is for plugin else false.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	/**
+	 * Determine if any of the checks requires the universal runtime preparation.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of check instances to run.
+	 *
+	 * @return boolean Returns true if one or more checks requires the universal runtime preparation.
```
felixarntz: Per WP docs standards:
```suggestion
	 * @param array $checks An array of check instances to run.
	 * @return bool Returns true if one or more checks requires the universal runtime preparation.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	/**
+	 * Return all shared preparations used by the checks to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of Check instances to run.
+	 *
+	 * @return array An array of reparations to run.
```
felixarntz: See above:
```suggestion
	 * @param array $checks An array of Check instances to run.
	 * @return array An array of reparations to run.
```

Though we probably need to change the return description if we instead return the array with `class` and `args` entries here.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+
+		return null;
+	}
+
+	/**
+	 * Run the checks against the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Result
```
felixarntz: This still needs a return description.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 * Determines if the current request is intended for the plugin checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return boolean Returns true if the check is for plugin else false.
+	 */
+	abstract public function is_plugin_check();
+
+	/**
+	 * Setup the check context, checks and preparations based on the request.
```
felixarntz: This will need to be updated related to my other feedback. Please keep in mind to start function/method descriptions with a third-person verb.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+	 *
+	 * @since n.e.x.t
+	 */
+	abstract protected function setup_checks();
+
+	/**
+	 * Prepare the environment for running the requested checks.
```
felixarntz: Per WP docs standards:
```suggestion
	 * Prepares the environment for running the requested checks.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+		if ( ! $this->requires_universal_preparations( $this->checks_to_run ) ) {
+			$preparation = new Universal_Runtime_Preparation( $this->context );
+			return $preparation->prepare();
+		}
+
+		return null;
+	}
+
+	/**
+	 * Run the checks against the plugin.
```
felixarntz:
```suggestion
	 * Runs the checks against the plugin.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+			foreach ( $cleanups as $cleanup ) {
+				$cleanup();
+			}
+		}
+
+		return $results;
+	}
+
+	/**
+	 * Determine if any of the checks requires the universal runtime preparation.
```
felixarntz:
```suggestion
	 * Determines if any of the checks requires the universal runtime preparation.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,179 @@
...
+			if ( $check instanceof Runtime_Check ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * Return all shared preparations used by the checks to run.
```
felixarntz:
```suggestion
	 * Returns all shared preparations used by the checks to run.
```

felixarntz (changes requested):
@jjgrainger I left a number of comments here for a few things that I think are worth iterating on. Most importantly, using declarative methods to return specific parts make for a more intuitive API than relying on one method that has to set multiple specific class properties. It is like that in the ACs, but looking at the code it makes for a rather clunky experience.

One higher level thing that we have not considered so far, but I think it may be relevant here is that runtime checks as a whole should only ever be _eligible_ to run if the plugin to check is in fact active. We may want to add a clause somewhere here that early on throws an exception if the plugin is not active and any of the provided checks to run is a runtime check.

Most comments not related to the above are nit-picks, e.g. about more closely following WP docs standards.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: @felixarntz this was missed and meant to be called in the `prepare` method. I will update accordingly.

jjgrainger: @felixarntz after reviewing your other comments I've gone with additional methods to return the specific class instances we need for preparing and running checks. I've updated the comments for each to explain what they should create and return.

jjgrainger:
@felixarntz I've updated the PR based on your feedback, this is ready for you to review. Thanks

jjgrainger:
@felixarntz I've updated this PR to include the work for #14 in order to test the approach we discussed. This is ready for review.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: @felixarntz after updating this I had errors when running tests saying the class could not be found? Not entirely sure why?

includes/Checker/Checks.php
```diff
@@ -54,7 +54,7 @@ public function run_checks( array $checks ) {
 		$checks_to_run = array_filter(
 			$checks,
 			function( $check ) use ( $all_checks ) {
-				return in_array( $check, $all_checks, true );
+				return in_array( $check, $all_checks ); // phpcs:ignore WordPress.PHP.StrictInArray.MissingTrueStrict
```
jjgrainger: @felixarntz This change was required in order for the `Check` instances passed to match with those returned in `get_checks()`. I know we have discussed setting up the checks array with slugs for keys to better match with what would be passed via CLI/AJAX.

I can update this if you agree with that approach?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,151 @@
...
+	abstract protected function get_checks();
+
+	/**
+	 * Creates and returns the Check_Context for the plugin to check based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context The Check_Context for the plugin to check.
+	 */
+	abstract protected function get_context();
```
felixarntz: I don't think we need this. Here's why:
* In order to create the `Checks` instance (see `get_checks()`), you need to create the `Check_Context` anyway.
* We should rather add a public `context()` method to the `Checks` class in order to then access that `Check_Context` from the same instance instead of requiring an additional function for something that's already happening anyway in `get_checks()`.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,151 @@
...
+
+	/**
+	 * Creates and returns an array of Check instances to run based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances.
+	 */
+	abstract protected function get_checks_to_run();
```
felixarntz: In order to return a list of `Check` instances, we will need to use the `Checks::get_checks()` method, and the `Checks` instance is being separately created in the `get_checks()` method.

So I think this method would make more sense and be much simpler if its responsibility was to only return the check slugs of the checks to run, based on the request, not the `Check` instances, as for that it would need to implement more logic that could instead be commonly handled in this base class.

My suggestion would be:
* Change this method to be something like `get_check_slugs_to_run()` (also update the doc block accordingly).
* Introduce another protected, non-abstract `get_checks_to_run()` method in here which calls `get_check_slugs_to_run()` and `get_checks()` and through that returns the list of `Check` instances.

On that note: We still need to change the `Checks::get_checks()` documentation (both method and filter) to clarify that this should be a map of check slug and check instance, not a simple list of check instances.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,151 @@
...
+	abstract public function is_plugin_check();
+
+	/**
+	 * Creates and returns an instance of the Checks class based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Checks An instances of the Checks class.
+	 */
+	abstract protected function get_checks();
```
felixarntz: I'm wondering whether we should call this `get_checks_instance()` as that would make it a lot more clear? `get_checks()` makes me think this should return an array of checks, but that's not the purpose of this function.

Alternatively, maybe we should just rename the `Checks` class to something more meaningful, e.g. `Checks_Controller` or `Checks_Manager` or something like that. Then we could also rename this method in the same way.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,151 @@
...
+				if ( ! isset( $shared_preparations[ $key ] ) ) {
+					$shared_preparations[ $key ] = array(
+						'class' => $class,
+						'args'  => $args,
+					);
+				}
+			}
+		}
+
+		return $shared_preparations;
```
felixarntz: Nit-pick: Probably best to drop the (now useless) keys at the end
```suggestion
		return array_values( $shared_preparations );
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,151 @@
...
+		return false;
+	}
+
+	/**
+	 * Returns all shared preparations used by the checks to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of Check instances to run.
+	 * @return array An array of reparations to run.
```
felixarntz: This needs to be clarified in regards to what format this is exactly being returned in.

includes/Checker/CLI_Runner.php
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	public function is_plugin_check() {
+		if ( empty( $_SERVER['argv'] ) ) {
+			return false;
+		}
+
+		if (
+			'wp' === $_SERVER['argv'][0] &&
+			'plugin' === $_SERVER['argv'][1] &&
+			'check' === $_SERVER['argv'][2]
```
felixarntz, line 42: These array accesses could cause PHP warnings. We either need to have an `isset` for each of them, or (probably simpler) we need to bail before if `count( $_SERVER['argv'] ) < 3 )`.

includes/Checker/CLI_Runner.php
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The absolute path to the plugin main file.
+	 */
+	private function get_plugin_main_file() {
+		// Get the plugin name from the command line arguments.
+		$plugin_name = $_SERVER['argv'][3];
+
+		return WP_PLUGIN_DIR . '/' . $plugin_name . '/' . $plugin_name . '.php';
```
felixarntz: This is unreliable, there is no guarantee that the plugin's main file has the same name as the plugin directory.

For reference, see https://github.com/felixarntz/plugin-check/blob/main/includes/CLI/Plugin_Check_Command.php#L189: This method supports both a plugin slug or a full plugin basename being passed, and it only works if that plugin is actually installed.

Maybe introduce a class like `Plugin_Request_Utility` with a static method `get_plugin_basename_from_input( $plugin_slug_or_basename )`? That way we could use that logic here, but also later in the actual CLI command, the `AJAX_Runner`, etc.

All of that also means that this _may_ fail, if some invalid slug or basename is passed, in which case we should probably throw an exception. So we should probably clarify in this method, and also more importantly in the `get_checks()` method where this private method is called.

includes/Checker/CLI_Runner.php
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * Returns an instance of the Check_Context class.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context
+	 */
+	protected function get_context() {
+		return new Check_Context( $this->get_plugin_main_file() );
+	}
```
felixarntz, line 69: See my feedback further above, we can probably drop this in favor of a `Checks::context()` method.

includes/Checker/CLI_Runner.php
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+
+		$all_checks = $this->get_checks()->get_checks();
+
+		if ( empty( $checks ) ) {
+			return $all_checks;
+		}
+
+		return array_intersect_key( $all_checks, array_flip( $checks ) );
```
felixarntz: See my comment above, better to have this logic in the abstract class, as it'll be just the same for the AJAX runner. The only thing that's different between CLI and AJAX is how the slugs need to be detected from the request.

includes/Checker/Checks.php
felixarntz: We should still change that, but for now, I'd suggest a simpler fix: The only reason the `true` here didn't work is because every `get_checks()` call creates new instances right now. That's somewhat problematic in itself, so what we should probably do is modify `Checks::get_checks()` to create that list (and run the filter) only once. We can just "memoize" the result after the first run in a static or class variable.

felixarntz (commented):
@jjgrainger Thanks for the changes, this looks much better. I left some follow up feedback.

includes/Checker/Checks.php
jjgrainger: Thanks @felixarntz I've updated this so we have a single instance of the `Checks` class.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I saw this and imagined it would need to be refactored. I've taken your suggestion of adding the `context()` method to the `Checks` class.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I've taken your suggestion and updated this to include the `get_check_slugs_to_run()` method.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz for now I've gone with the `get_check_instance()` method as it was the quickest change here.

includes/Checker/CLI_Runner.php
jjgrainger: Thanks @felixarntz I've added the `Plugin_Request_Utility` class and tests

includes/Checker/Checks.php
```diff
@@ -36,6 +44,17 @@ public function __construct( $plugin_main_file ) {
 		$this->check_context = new Check_Context( $plugin_main_file );
 	}
 
+	/**
+	 * Returns the Check Context.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context
```
felixarntz: Please add a description.

includes/Checker/Checks.php
```diff
@@ -105,19 +124,25 @@ protected function run_check_with_result( Check $check, Check_Result $result ) {
...
-		 *
-		 * @since n.e.x.t
-		 *
-		 * @param array $checks List of plugin check class instances implementing the Check interface.
-		 */
-		return apply_filters( 'wp_plugin_check_checks', $checks );
+		if ( ! isset( $this->checks ) ) {
+			// TODO: Add checks once implemented.
+			$checks = array(
+				'i18n-usage' => new Checks\I18n_Usage_Check(),
```
felixarntz: Using underscores is a bit of a better convention than using hyphens. For example, underscores are allowed in variable names and in JS object properties, while hyphens are not.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,165 @@
...
+
+		return array_values( $shared_preparations );
+	}
+
+	/**
+	 * Returns the Check instances to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances.
```
felixarntz: This should probably be changed to describe the same map that is now returned by the `Checks::get_checks()` method.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -0,0 +1,165 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks An array of Check instances to run.
+	 * @return array {
+	 *      An array of Preparations to run where each item is an array with the class name and args.
+	 *
+	 *      @type string $class The full class name of the Preparation.
+	 *      @type array  $args  An array of parameters to pass to the class constructor.
+	 * }
```
felixarntz: While this is correct as a human-readable description, technically speaking in terms of PHPDoc this is not correct, as the current shape would suggest the return value is an associative array that includes `class` and `args` keys. However, the return value is an indexed array where each entry is that above array.

I think for simplicity sake we should therefore stick with just a precise description, otherwise the correct return type annotation would become somewhat complex.
```suggestion
	 * @return array An array of Preparations to run where each item is an array with keys `class` and `args`.
```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -0,0 +1,63 @@
...
+	 * @param string $plugin_slug The plugin slug or basename.
+	 * @return string The plugin basename.
+	 *
+	 * @throws Exception Thrown if an invalid basename or plugin slug is provided.
+	 */
+	public static function get_plugin_basename_from_input( $plugin_slug ) {
+		require_once ABSPATH . 'wp-admin/includes/plugin.php';
+
+		if ( empty( $plugin_slug ) ) {
+			throw new Exception( 'Missing positional argument. Please provide the plugin slug as first positional argument.' );
```
felixarntz: Please update the message here to be context-independent. This message worked well when it was directly scoped to the CLI command, but now referring to a "positional argument" no longer makes sense.

Same applies to the other messages below.
```suggestion
			throw new Exception( 'Invalid plugin slug: Plugin slug must not be empty.' );
```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -0,0 +1,63 @@
...
+
+		// Is the provided value is a full plugin basename?
+		if ( isset( $plugins[ $plugin_slug ] ) ) {
+			return $plugin_slug;
+		}
+
+		if ( strpos( $plugin_slug, '/' ) ) {
+			throw new Exception(
+				sprintf(
+					'Invalid positional argument. Plugin with basename %s is not installed.',
```
felixarntz:
```suggestion
					'Invalid plugin basename: Plugin with basename %s is not installed.',
```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -0,0 +1,63 @@
...
+
+		foreach ( $plugins as $plugin_basename => $plugin_data ) {
+			if ( strpos( $plugin_basename, $plugin_slug . '/' ) === 0 ) {
+				return $plugin_basename;
+			}
+		}
+
+		throw new Exception(
+			sprintf(
+				'Invalid positional argument. Plugin with slug %s is not installed.',
```
felixarntz:
```suggestion
				'Invalid plugin slug: Plugin with slug %s is not installed.',
```

felixarntz (approved):
@jjgrainger This looks great. Only a few nit-picks, but nothing blocking.

mukeshpanchal27 (approved):
Thanks @jjgrainger, LGTM!

PR #76, opened 2023-02-10
Add Enqueued_Scripts_Size_Check class

### LINKED ISSUE #17: Create Enqueued_Scripts_Size_Check
## Description

Checks the number of scripts and the total size of all scripts registered in the plugin. This check will run the wp_enqueue_scripts action to register all plugin scripts. Using wp_scripts(), all scripts registered by the plugin will be looped over to get the size of the script along with any additional inline scripts. The number and total file size of all scripts will be calculated and then tested against the threshold, warning if either exceeds a threshold. Check should run relevant actions to make the data as accurate as possible as scripts can be registered and enqueued with any actions provided by WP like in the header or in the document body. Check should run through multiple contexts to make sure scripts enqueued by the plugin are covered in the analysis. Contexts could be any URLs from the site like the homepage, post types, templates, etc.

This issue is dependant on:
- #40 

#### Enqueued_Scripts_Size_Check class
- Class `Enqueued_Scripts_Size_Check` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Scripts_Size_Check` should extend the `Abstract_Runtime_Check` class.
- Class `Enqueued_Scripts_Size_Check` will use the `URL_Aware` trait.
- Class `Enqueued_Scripts_Size_Check` will implement the`With_Shared_Preparations` interface.
- Class has a protected `$threshold_size` property with a value of `300000`.
- A `prepare()` method will be implemented in the class.
    - This method will save the `$GLOBALS['wp_scripts'];` to a temporary variable.
    - Then set the `$GLOBALS['wp_scripts'];` to null if set.
    - A clean up function is returned which uses the temporary variable to reinstate the original `$GLOBALS['wp_scripts'];` value.
- A `get_shared_preparations()` method will be implemented.
    - The `get_shared_preparations()` method will be used to return a preparation that creates a demo post for all publicly viewable post types using #40 
    - The method will return an array where the `Demo_Post_Creation_Preparation` class name is an array key and it's value is an array of posts data to create, compatible with [`wp_insert_post`](https://developer.wordpress.org/reference/functions/wp_insert_post/).
- A protected `get_urls()` method will be created.
    - The method will return an array of URLs containing the home URL and URLs for all publicly viewable post types.  
- Method `run( Check_Result $results )` should be implemented from the `Runtime_Check` abstract class.
    - The `run()` method will call the `run_for_urls()` method passing an array of URLs returned by the `get_urls()` method, a callable which will be the classes `check_url()` method, and the `Check_Results` instance.
    -  The `Check_Result $check_result` will be returned by the `run` method.
- A protected `check_url()` method will be created.
    - The method will accept an instance of the `Check_Result` class and a URL string as parameters.  
    - The method will reset the `$GLOBALS['wp_scripts'];`
    - Then will call the `wp_enqueue_scripts()` to enqueue all of the checked plugins scripts.
    - Using the object returned by `wp_scripts()` all enqueued scripts for the URL will be looped over:
        - Check the script is part of the plugin being checked by matching the plugin URL with the script URL
        - If the script is part of the plugin, calculate the size of the script including any additional inline scripts.
        - Add the script `path` and `size` to a `$plugin_scripts` array to be used as part of the results.
        - Record the total plugin scripts size in a `$plugin_scripts_size` variable to be used as part of the results.
    - After looping over all plugin scripts, the total `$plugin_scripts_size` is checked agains the `$threshold_size`
    - If the total script size is larger than the threshold, warnings are added to the `$check_results`
         - Each script in the `$plugin_scripts` is looped over an a warning message added for each.
         - Each warning includes the current URL, the threshold, the script path and size.

#### URL Aware Trait
- A new `URL_Aware` trait exists in `includes\Traits\URL_Aware`
- Trait contains a public `go_to( $url )` method that will simulate in WordPress the passed URL.
    - The `go_to()` method will simulate the URL, similar to the WordPress Core test suite, by modifying the global `WP_Query` to match the URL passed.
- Trait contains a public `run_for_urls( array $urls, callable $callback )`
    - The method will loop over each of the URLs passed and simulate the URL using the `go_to()` method.
    - Then it will call the `$callback` function passed for each of the URLs passing an instance of the `Check_Result` class and the URL as parameters.

## Tests Coverage
- `Enqueued_Scripts_Size_Check` class is expected to extend `Runtime_Check` class
- `Enqueued_Scripts_Size_Check` class is expected to be instance of `Runtime_Check` class

vishalkakadiya:
Add Enqueued_Scripts_Size_Check class

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/17

Closes https://github.com/10up/plugin-check/issues/17


includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Traits;
+
+trait URL_Aware {
+
+	/**
+	 * Simulate the URL, similar to the WordPress Core test suite.
```
felixarntz:
```suggestion
	 * Sets the global state to as if a given URL has been requested.
	 *
	 * The implementation is similar to the one from the WordPress Core test suite.
```

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,83 @@
...
+	 *
+	 * @param string $url Url.
```
felixarntz:
```suggestion
	 * @param string $url URL to simulate request for.
```

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,83 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Traits\URL_Aware
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Traits;
+
+trait URL_Aware {
```
felixarntz, line 10: One thing I notice is missing here is a way to back up all the globals that are modified by the `go_to()` method, and have a way to reset them. Without that, we'll mess up global state without cleaning up which will pollute other tests.

Maybe we should introduce protected methods like `backup_globals()` and `restore_globals()` here? These methods could then be called in any `prepare()` method and its cleanup callback of a `Check` that uses this trait.

Also see https://github.com/felixarntz/plugin-check/blob/main/includes/Checker/Checks/Abstract_Request_Check.php#L53 from the POC, but note that this is using an abstract class so that exact code approach is not as suitable for a trait as the two separate methods mentioned above (since the check class itself will already have a `prepare()` method).

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * Simulate the URL, similar to the WordPress Core test suite.
+	 *
+	 * @param string $url Url.
+	 */
+	public function go_to( $url ) {
```
felixarntz: Why is this public? It's not called anywhere outside this class. Let's default to the least visibility on any method and only make it more visible if truly needed.

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * Simulate all the urls like WP and run the cleanup function.
+	 *
+	 * @param array    $urls     An array of URLs to run.
+	 * @param callable $callback Callback function to run for each URL.
+	 *
+	 * @return void
+	 */
+	public function run_for_urls( $urls, $callback ) {
```
felixarntz: Same here, why is this public?

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+					'post_content' => 'Test content',
+					'post_type'    => $post_type,
+					'post_status'  => 'publish',
+				);
+			},
+			$this->get_viewable_post_types()
+		);
+
+		return array(
+			'WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation' => array( $demo_posts ),
```
felixarntz: Instead of hard-coding the full class name, it's better to actually import the class above (via `use`) and then use the `::class` suffix on it here to get the full class name.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+
+	use URL_Aware;
+
+	/**
+	 * Threshold for script size to surface a warning for.
+	 *
+	 * @since 1.0.0
+	 * @var int
+	 */
+	public $threshold_size = 300000;
```
felixarntz: Similar question as elsewhere, why is this public? Especially for class variables, those should almost never be public. I realize it is like that in the proof of concept, but that is just a proof of concept, so doesn't need clean APIs as much :)

I think a good idea would be to instead implement a constructor to this class where this variable can optionally be set, where `300000` would just be the default value.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		$orig_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
+
+		return function() use ( $orig_scripts ) {
+			$GLOBALS['wp_scripts'] = $orig_scripts;
+		};
+	}
```
felixarntz, line 83: See my other comment on the `URL_Aware` trait: Here we also need to back up the globals that `go_to()` modifies, and in the cleanup they need to be reset. Probably the cleanest way is to implement methods in the `URL_Aware` trait that this code here then calls.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+ * @since n.e.x.t
+ */
+class Enqueued_Scripts_Size_Check extends Abstract_Runtime_Check implements With_Shared_Preparations {
+
+	use URL_Aware;
+
+	/**
+	 * Threshold for script size to surface a warning for.
+	 *
+	 * @since 1.0.0
```
felixarntz: Please use `@since n.e.x.t` for every since annotation.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+
+		return array(
+			'WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation' => array( $demo_posts ),
+		);
+	}
+
+	/**
+	 * Runs the check on the plugin and amends results.
+	 *
+	 * @param Check_Result $result The check results to amend and the plugin context.
```
felixarntz, line 116: Missing a since annotation.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+			}
+		}
+
+		return $urls;
+	}
+
+	/**
+	 * Returns an array of viewable post type names.
+	 *
+	 * @return array An array of post type names.
```
felixarntz: See above. Please also check the rest of the PR for missing or incorrect since annotations.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		$orig_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
+
+		return function() use ( $orig_scripts ) {
+			$GLOBALS['wp_scripts'] = $orig_scripts;
```
felixarntz: Nit-pick: If the `$wp_scripts` global was originally not set, we should rather unset it here than set it to `null`. It's _almost_ the same, but not quite :)

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+	 * Gets the list of URLs to run this check for.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @return array List of URL strings (either full URLs or paths).
+	 */
+	protected function get_urls() {
+		$urls = array( home_url() );
+
+		foreach ( $this->get_viewable_post_types() as $post_type ) {
```
felixarntz: Instead of calling this method in `get_shared_preparations()` and again here, maybe we can instead set up the list of viewable post types within the constructor of this check, in a private variable? This way we can simply access this variable and don't have to run the logic multiple times. By using a variable, we also ensure that the list is definitely the same both times - with calling the method both times, it could be that the results are different which would be a problem.

We could even consider then making the list of viewable post types just the default, for example the constructor could optionally allow providing another list of post types (e.g. if you only want to check this for posts) - although that's probably not really needed.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,196 @@
...
+			// Get size of script src.
+			$script_path = str_replace( $result->plugin()->url(), $result->plugin()->path(), $script->src );
+			$script_size = function_exists( 'wp_filesize' ) ? wp_filesize( $script_path ) : filesize( $script_path );
+
+			// Get size of additional inline scripts.
+			if ( ! empty( $script->extra['after'] ) ) {
+				foreach ( $script->extra['after'] as $extra ) {
+					$script_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
+				}
+			}
```
felixarntz, line 206: I think we also need to have the same for any `before` scripts.

tests/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
```diff
@@ -0,0 +1,99 @@
...
+
+		$errors   = $check_result->get_errors();
+		$warnings = $check_result->get_warnings();
+
+		$this->assertEmpty( $errors );
+		$this->assertEmpty( $warnings );
+
+		$this->assertEquals( 0, $check_result->get_error_count() );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
+	}
```
felixarntz: There's a lot of duplicate code between these two methods, and a lot of that we will probably need for any other runtime check test code as well.

How about we introduce a `Runtime_Check_UnitTestCase` class? It could have some of those helper methods, e.g.:
* A method to return the `Check_Context` for a given plugin.
* A method to prepare the environment (based on the `Check_Context` and the specific check instance to test, covering all preparation, universal, shared, and check-specific) and return the cleanup closure
* A method to run the check (based on the `Check_Context` and the specific check instance to test) and return the `Check_Result` instance

We could use these methods in both of these test methods to make them a lot shorter and less repetitive, and other runtime check tests could later use it as well.

tests/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
```diff
@@ -0,0 +1,99 @@
...
+			$preparation = new $class( ...$args );
+			$cleanups    = $preparation->prepare();
+		}
+
+		return function() use ( $cleanups ) {
+			foreach ( $cleanups as $cleanup ) {
+				$cleanup();
+			}
+		};
+	}
```
felixarntz, line 117: See my comment above, this method could (in this form or slightly modified) be moved to a reusable `Runtime_Check_UnitTestCase`.

tests/bootstrap.php
```diff
@@ -26,7 +26,10 @@
 
 // Force plugin to be active.
 $GLOBALS['wp_tests_options'] = array(
-	'active_plugins' => array( basename( TESTS_PLUGIN_DIR ) . '/plugin-check.php' ),
+	'active_plugins' => array(
+		basename( TESTS_PLUGIN_DIR ) . '/plugin-check.php',
+		basename( TESTS_PLUGIN_DIR ) . '/tests/testdata/plugins/test-plugin-with-errors/test-plugin-with-errors.php',
```
felixarntz: This does not look right. We shouldn't generally force activate any plugin except the one that the unit tests are for (the plugin check plugin).

tests/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
```diff
@@ -0,0 +1,99 @@
...
+		$runtime_cleanup();
+
+		$errors   = $check_result->get_errors();
+		$warnings = $check_result->get_warnings();
+
+		$this->assertEmpty( $errors );
+		$this->assertEmpty( $warnings );
+
+		$this->assertEquals( 0, $check_result->get_error_count() );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
```
felixarntz: This is not an appropriate test; it is supposed to test errors, but the assertions here actually do exactly the same as in the test for checking no errors.

This is likely due to the JS file tested being to small, but that doesn't mean we just shouldn't test it.

I understand it is a bit questionable to create a 300000 byte JS file just for testing, and we shouldn't do it. BUT: There is the `$threshold_size` class property in the `Enqueued_Scripts_Size_Check` class: How about in this test we just set it to a very low value to ensure our test script exceeds that threshold?

tests/testdata/plugins/test-plugin-with-errors/test-plugin-with-errors.php
```diff
@@ -16,3 +16,4 @@
  */
 
 include_once 'i18-usage-error-file.php';
+include_once 'enqueued-scripts-size-check.php';
```
felixarntz: Rather than creating a single test plugin that has a bunch of errors, we should create _individual_ test plugins, one for each kind of test; otherwise things will get messy very quickly with our test code as the single plugin will do a bunch of things over time.

So let's not add to the existing test plugin but rather add a new one like `test-plugin-with-enqueued-scripts-size-errors`. We may also want to rename the existing `test-plugin-with-errors` to `test-plugin-with-i18n-usage-errors`.

tests/testdata/plugins/test-plugin-with-errors/enqueued-scripts-size-check.php
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+function plugin_check_test_enqueue_large_script() {
+	wp_enqueue_script(
+		'plugin_check_large_script',
+		plugin_dir_url( __FILE__ ) . 'assets/large-script.js'
+	);
+}
+
+add_action( 'wp_enqueue_scripts', 'plugin_check_test_enqueue_large_script' );
```
felixarntz: This code allows us to test the main script size. We should also find a way to test the logic in `Enqueued_Scripts_Size_Check` that covers inline scripts.

We could for example add an inline script here, and in the test set the threshold to a value that will only be exceeded by the main script _plus_ the inline script.

felixarntz (changes requested):
@vishalkakadiya There are a couple things that need iteration here; mostly small things, but a few are critical to be addressed prior to merge (e.g. the modification to the test bootstrap script, and a cleanup routine for globals).

Please see my below feedback.

felixarntz:
@jjgrainger @vishalkakadiya Regarding the problem that the test plugin code never runs in the unit tests, the issue is that the test plugin is never actually loaded. And we shouldn't force it to load generally via the test bootstrap file as that would pollute other tests (see my comment in the review above).

Since plugins are only loaded once, very early in the WordPress load cycle, we have to find ways to write test plugins so that they can be loaded on demand in the test suite without polluting global scope.
Here's one idea that could work:
* We could write every test plugin as a plugin that adds nothing to the global scope that can't be unset (e.g. it's okay to add hooks, or modify a global variable, but it's not okay to introduce functions or classes).
* Basically this means that every test plugin we write should only ever use closures, never named functions. And all it should do to the global scope is add actions or filters, using closures for the callbacks. This way, everything that the plugin does can be reset without disrupting other subsequent tests.
* This way, in any test that needs to simulate this plugin being active, we can `require` the plugin main file (not `require_once`). As long as it doesn't define any functions or classes, we can load it multiple times, in multiple tests. And any hooks that loading the plugin adds will be reset afterwards (as all WordPress hooks are reset automatically after each test).
* Of course this wouldn't work as a general requirement for plugins, but since we're only speaking about the test plugins for this repository to use in unit tests, I think that's a reasonable requirement.

Also see my related review feedback above, we should use separate test single-file plugins per individual check that we test, otherwise, if we have a single plugin with all errors, things will get messy and they may conflict with each other. We should be able to change the test plugin with errors that we have so far to make it two separate plugins and have both of those plugins satisfy the above requirements in order not to irrevocably pollute global scope.

includes/Traits/URL_Aware.php
jjgrainger: Thanks @felixarntz I've added this logic to the trait and called the new methods within the Check.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz updated!

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz I've updated this and other properties/methods to private/protected where reasonable.

I have used your idea with a class constructor for the threshold as this would help with testing.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz I have gone over this class and others in the PR and added the `@since n.e.x.t` annotations where missing

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz updated

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz updated and made this part of the constructor

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
jjgrainger: Thanks @felixarntz I've included the logic to capture before scripts

tests/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
jjgrainger: Thanks @felixarntz I was thinking of creating something similar while writing the tests but was trying to get tests working before refactoring. I have updated the PR with a `Runtime_Check_UnitTestCase` like you described

tests/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
jjgrainger: Thanks @felixarntz I was uncomfortable creating and committing a large file just for testing. I liked the idea of updating the threshold size for testing and have implemented this with the class constructor.

tests/bootstrap.php
jjgrainger: Thanks @felixarntz I agree, I wasn't comfortable with this and only tried this method to get the plugin tests working. I've removed this based on the new approach discussed.

tests/testdata/plugins/test-plugin-with-errors/test-plugin-with-errors.php
jjgrainger: Thanks @felixarntz I've updated all the test plugins in the PR to follow this new structure.

tests/testdata/plugins/test-plugin-with-errors/enqueued-scripts-size-check.php
jjgrainger: Thanks @felixarntz I've added an inline script and used the `$threshold_size` we pass in the constructor to test when the inline script goes over the threshold.

jjgrainger:
Thanks @felixarntz the feedback here and the discussions with @joemcgill on Slack have been helpful here!

I've gone over the changes requested and made the updates to the test plugins based on your feedback. This has been helpful in cleaning things up and avoiding issues around loading the plugins for testing.

The changes I've made has got test working but there are a couple of things I want to highlight.
- When trying to 'activate' the test plugin for this check the `Force_Single_Plugin_Preparation` would throw an exception because the plugin isn't recognised. This error comes from the `validate_plugin()` call because the test plugin does not exist in the array returned by `get_plugins()`. The way I've worked around this is by using the Plugin Checker plugin as the check context as the test plugin file exists within this plugin.
- After making all the changes I still faced the same issue where the script being enqueued by the test plugin would never end up in the `wp_scripts()->done` array. To resolve this I have also included the `wp_head()` and `wp_footer()` calls along with `wp_enqueue_scripts()` in the `Enqueued_Scripts_Size_Check` class. I haven't looked deeper into this to determine if this is the correct approach or there is a different way to resolve this.

Let me know if you have any questions or feedback on this.

Thanks

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,229 @@
...
+	protected function check_url( Check_Result $result, $url ) {
+		// Reset the WP_Scripts instance.
+		unset( $GLOBALS['wp_scripts'] );
+
+		// Run the 'wp_enqueue_script' action, wrapped in an output buffer in case of any callbacks printing scripts
+		// directly. This is discouraged, but some plugins or themes are still doing it.
+		ob_start();
+		wp_enqueue_scripts();
+		wp_head();
+		wp_footer();
```
joemcgill: My main question about this approach is that it can introduce other side effects when all of this is called. I'm also unsure how this accounts for situations where someone is conditionally loading scripts based on some state associated with the URL, like the template type, etc.

Instead of calling wp_head() and wp_footer() directly, you could call `WP_Scripts::do_head_items()` and `WP_Scripts::do_footer_items()`, respectively without introducing other side effects if there are other actions hooked to header/footer actions.

jjgrainger: Thanks @joemcgill I agree it would be cleaner using the `WP_Scripts` methods you've mentioned. I've updated the PR to use these.

joemcgill: Thanks @jjgrainger. Can you clarify my other question about how the `$url` is intended to be used in relationship to what scripts are being enqueued?

jjgrainger: @joemcgill we simulate the URL being requested by modifying the main `WP_Query`. We have a `URL_Aware` trait which handles this and provides methods to use within checks. The `run` method in this check uses those to loop over each of the URLs and run the check logic against each publicly viewable post type and home page. Let me know if you have any other questions.

joemcgill (approved):
I'm not seeing anything that stands out as a blocker here. I think it will be good to get this merged and do some field analysis on a few known plugins to make sure this performing the way we expect, particularly since mocking state changes to globals can be tricky to account for.

composer.json
```diff
@@ -33,7 +33,8 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "WordPress\\Plugin_Check\\Test_Data\\": "tests/testdata"
+      "WordPress\\Plugin_Check\\Test_Data\\": "tests/testdata",
+      "WordPress\\Plugin_Check\\Tests\\": "tests/utils"
```
felixarntz: The namespace should better be something like `Test_Utils`, to match the folder name. A `Tests` namespace should be reserved if we would potentially load PHP code that is directly within the `tests` folder.

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,188 @@
...
+		}
+		if ( isset( $GLOBALS['wp_query'] ) ) {
+			$global_vars['wp_query'] = $GLOBALS['wp_query'];
+		}
+		if ( isset( $GLOBALS['wp_the_query'] ) ) {
+			$global_vars['wp_the_query'] = $GLOBALS['wp_the_query'];
+		}
+		if ( isset( $GLOBALS['wp'] ) ) {
+			$global_vars['wp'] = $GLOBALS['wp'];
+		}
```
felixarntz: ~~This logic is exactly the same as in the `foreach` loop above. Maybe we should just add those three keys to `$this->query_globals`?~~

**Update:** I see in `go_to()` they are handled differently. In that case, maybe we introduce another private property for this separate list of three globals? That way we can use the same `foreach` loop to go over both of these lists (by combining them before in a local variable).

includes/Traits/URL_Aware.php
```diff
@@ -0,0 +1,188 @@
...
+		if ( isset( $global_vars['wp_the_query'] ) ) {
+			$GLOBALS['wp_the_query'] = $global_vars['wp_the_query'];
+		} else {
+			unset( $GLOBALS['wp_the_query'] );
+		}
+		if ( isset( $global_vars['wp'] ) ) {
+			$GLOBALS['wp'] = $global_vars['wp'];
+		} else {
+			unset( $GLOBALS['wp'] );
+		}
```
felixarntz: See above.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,229 @@
...
+	 * @var array
+	 */
+	private $viewable_post_types;
+
+	/**
+	 * Set the threshold size for script sizes to surface warnings.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param integer $threshold_size The threshold in bytes for script size to surface warnings.
```
felixarntz: Per WP docs standards, this should say `int` instead of `integer`.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,229 @@
...
+		$check->backup_globals();
+
+		return function() use ( $orig_scripts, $check ) {
+			if ( is_null( $orig_scripts ) ) {
+				unset( $GLOBALS['wp_scripts'] );
+			} else {
+				$GLOBALS['wp_scripts'] = $orig_scripts;
+			}
+
+			$check->restore_globals();
```
felixarntz: The assignment to a `$check` variable is unnecessary. Closures that are defined in a class can access `$this` to reference the class instance (as of PHP 5.4, which is lower than our minimum requirement).
```suggestion
		$this->backup_globals();

		return function() use ( $orig_scripts ) {
			if ( is_null( $orig_scripts ) ) {
				unset( $GLOBALS['wp_scripts'] );
			} else {
				$GLOBALS['wp_scripts'] = $orig_scripts;
			}

			$this->restore_globals();
```

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,229 @@
...
+			$posts = get_posts(
+				array(
+					'posts_per_page' => 1,
+					'post_type'      => $post_type,
+					'post_status'    => 'publish',
+				)
+			);
+			if ( isset( $posts[0] ) ) {
+				$urls[] = get_permalink( $posts[0] );
+			}
```
felixarntz: Rather than silently only adding the URL if the post is present in the result, we should throw an exception if a post is missing. If the post is not returned, something went wrong, and our code should fail, rather than the check not being reliable for an "unknown" reason.

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -0,0 +1,229 @@
...
+				$result->add_message(
+					false,
+					sprintf(
+						'This script has a size of %1$s which in combination with the other scripts enqueued on %2$s exceeds the script size threshold of %3$s.',
+						size_format( $plugin_script['size'] ),
+						$url,
+						size_format( $this->threshold_size )
+					),
+					array(
+						'code' => 'EnqueuedScripts',
```
felixarntz: This should rather use `EnqueuedScriptsSize` to properly refer to this check.

Maybe we should follow the convention for PHPCS? We could use something like `EnqueuedScriptsSize.ScriptSizeGreaterThanThreshold`? This convention will allow us to use specific codes in the future for different kinds of warnings and errors within a single check class.

tests/utils/TestCase/Runtime_Check_UnitTestCase.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * Abstract Runtime_Check_UnitTestCase.
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Tests\TestCase;
```
felixarntz: See above comment regarding namespace. This could be `WordPress\Plugin_Check\TestUtils\TestCase`.

tests/utils/TestCase/Runtime_Check_UnitTestCase.php
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Prepares the test environment and runs the check returning the results.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check         $check   The Check instance to run.
+	 * @param Check_Context $context The check context for the plugin to be checked.
+	 * @return Check_Result An object containing all check results.
+	 */
+	public function run_check( Check $check, Check_Context $context ) {
```
felixarntz: No need for this to be `public` I think?

tests/utils/TestCase/Runtime_Check_UnitTestCase.php
```diff
@@ -0,0 +1,82 @@
...
+
+		// Prepare any shared preparations for the check.
+		if ( $check instanceof With_Shared_Preparations ) {
+			foreach ( $check->get_shared_preparations() as $class => $args ) {
+				$cleanups[] = ( new $class( ...$args ) )->prepare();
+			}
+		}
+
+		// Prepare the check.
+		$cleanups[] = $check->prepare();
```
felixarntz: This only must be called if the `$check` implements the `Preparation` interface. This is the case in our `Abstract_Runtime_Check` class, but it is not enforced to use that class, so we should make sure we don't assume that it is being used.

tests/utils/TestCase/Runtime_Check_UnitTestCase.php
```diff
@@ -0,0 +1,82 @@
...
+	 *
+	 * @param Check         $check   The Check instance to run.
+	 * @param Check_Context $context The check context for the plugin to be checked.
+	 * @return Check_Result An object containing all check results.
+	 */
+	public function run_check( Check $check, Check_Context $context ) {
+		$results = new Check_Result( $context );
+		$cleanup = $this->prepare_environment( $check, $context );
+		$check->run( $results );
+		$cleanup();
```
felixarntz: We need to make sure that the cleanup runs even if the check itself throws an exception.
```suggestion
		try {
			$check->run( $results );
		} catch ( \Exception $e ) {
			$cleanup();
			throw $e;
		}
		$cleanup();
```

felixarntz (commented):
@vishalkakadiya @jjgrainger This looks great now, thank you for the iterations!

I have a few last bits of feedback here, but it's almost good to merge.

jjgrainger:
Thanks @felixarntz I've updated the PR based on the feedback. This is ready for review

felixarntz (approved):
@jjgrainger Thank you for the updates, LGTM!

PR #77, opened 2023-02-10
Add admin tools page

### LINKED ISSUE #32: Create Tools page
## Description

The plugin should introduce a new page under Tools to allow users to run checks of plugins.

## Acceptance Criteria

- Class `Admin_Page` should exist at `includes/Admin`
- `Admin_Page` class should have the `__construct()` method.
- `add_hooks()` method should exists and have the `add_action( 'admin_menu', array( $this, 'add_page' ) );` hook registering. (**NOTE:** `add_hooks()` method will be called from wherever the class is instantiated).
- The `add_page()` will have `add_management_page()` with below params.
   - `$page_title` should be "Plugin Check"
   - `$menu_title` should be "Plugin Check"
   - `$capability` should be "activate_plugins"
   - `$menu_slug` should be "plugin-check"
   - Define callback method in class `Admin_Page` for `render_page `
   - `$position` should be anything or default
 - The plugin check admin page should be available as below:
 Screenshot 2022-11-21 at 7 56 24 PM
   
- `Admin_Page` class will have the `render_page()`, and it should print the dropdown with the list of plugins and the `check it!` button.
- Use the `get_plugins()` method to get a list of all the activated plugins, exclude `plugin check` from it and show it in the dropdown, like below:

Screenshot 2022-11-21 at 7 58 21 PM

## Tests Coverage

**Automated Tests:**
- Create an `Admin_Page_Test` class that checks whether the `admin_menu` filter is registering properly. 
- Checks that the `Plugin Check` page is registered or not in the `global $submenu`
- Checks the output of the `render_page()` by comparing the output of the plugin's dropdown list.

**Manual Tests:**
- Login to the admin area, and find the `Plugin Check` page available under the `Tools` menu.
- Check whether the dropdown with `the list of plugins` and `check it!` button is available.
- Check on the check it button and check if that is clickable or doing any process.

vishalkakadiya:
Add admin tools page

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/32

Closes https://github.com/10up/plugin-check/issues/32

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -69,6 +69,6 @@ class Admin_Page {
 	public function render_page() {
 		$available_plugins = $this->get_available_plugins();
 
-		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
+		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
 	}
 }
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -16,6 +16,8 @@ class Admin_Page {
 
 	/**
 	 * Initializes hooks.
+	 *
+	 * @since n.e.x.t
 	 */
 	public function add_hooks() {
 		add_action( 'admin_menu', array( $this, 'add_page' ) );
@@ -23,6 +25,8 @@ class Admin_Page {
 
 	/**
 	 * Registers the admin page under the tools menu.
+	 *
+	 * @since n.e.x.t
 	 */
 	public function add_page() {
 		add_management_page(
@@ -37,10 +41,11 @@ class Admin_Page {
 	/**
 	 * Returns the list of plugins.
 	 *
-	 * @return array List of plugins.
+	 * @since n.e.x.t
+	 *
+	 * @return array List of available plugins.
 	 */
 	private function get_available_plugins() {
-
 		$available_plugins = get_plugins();
 
 		if ( empty( $available_plugins ) ) {
@@ -58,9 +63,10 @@ class Admin_Page {
 
 	/**
 	 * Render the "Plugin Check" page.
+	 *
+	 * @since n.e.x.t
 	 */
 	public function render_page() {
-
 		$available_plugins = $this->get_available_plugins();
 
 		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Admin\Admin_Page
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Admin;
+
+/**
+ * Class is handling admin tools page functionality.
+ *
+ * @since n.e.x.t
+ */
+class Admin_Page {
+
+	/**
+	 * Initialize hooks.
+	 */
+	public function add_hooks() {
+		add_action( 'admin_menu', array( $this, 'add_page' ) );
+	}
+
+	/**
+	 * Register the admin page under the tools menu.
+	 */
+	public function add_page() {
+		add_management_page( __( 'Plugin Check', 'plugin-check' ), __( 'Plugin Check', 'plugin-check' ), 'activate_plugins', 'plugin-check', array( $this, 'render_page' ) );
+	}
+
+	/**
+	 * Returns the list of plugins.
+	 *
+	 * @return array List of plugins.
+	 */
+	public function get_available_plugins() {
+
+		$available_plugins = get_plugins();
+
+		if ( empty( $available_plugins ) ) {
+			return array();
+		}
+
+		$plugin_check_base_name = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
+
+		if ( isset( $available_plugins[ $plugin_check_base_name ] ) ) {
+			unset( $available_plugins[ $plugin_check_base_name ] );
+		}
+
+		return $available_plugins;
+	}
+
+	/**
+	 * Render the "Plugin Check" page.
+	 */
+	public function render_page() {
+
+		$available_plugins = $this->get_available_plugins();
+
+		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
+	}
+}
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -15,17 +15,23 @@ namespace WordPress\Plugin_Check\Admin;
 class Admin_Page {
 
 	/**
-	 * Initialize hooks.
+	 * Initializes hooks.
 	 */
 	public function add_hooks() {
 		add_action( 'admin_menu', array( $this, 'add_page' ) );
 	}
 
 	/**
-	 * Register the admin page under the tools menu.
+	 * Registers the admin page under the tools menu.
 	 */
 	public function add_page() {
-		add_management_page( __( 'Plugin Check', 'plugin-check' ), __( 'Plugin Check', 'plugin-check' ), 'activate_plugins', 'plugin-check', array( $this, 'render_page' ) );
+		add_management_page(
+			__( 'Plugin Check', 'plugin-check' ),
+			__( 'Plugin Check', 'plugin-check' ),
+			'activate_plugins',
+			'plugin-check',
+			array( $this, 'render_page' )
+		);
 	}
 
 	/**
@@ -33,7 +39,7 @@ class Admin_Page {
 	 *
 	 * @return array List of plugins.
 	 */
-	public function get_available_plugins() {
+	private function get_available_plugins() {
 
 		$available_plugins = get_plugins();
 
--- a/templates/admin-page.php
+++ b/templates/admin-page.php
@@ -5,36 +5,40 @@
  * @package plugin-check
  */
 
-if ( empty( $available_plugins ) ) {
-	return;
-}
 ?>
 
 <div class="wrap">
 
 	<h1><?php esc_html_e( 'Plugin Check', 'plugin-check' ); ?></h1>
 
-	<div class="card">
+	<div class="plugin-check-content">
 
-		<form>
-			<h2>
-				<label class="title" for="plugin-check__plugins">
-					<?php esc_html_e( 'Check the Plugin', 'plugin-check' ); ?>
-				</label>
-			</h2>
+		<?php if ( ! empty( $available_plugins ) ) { ?>
 
-			<select id="plugin-check__plugins">
-				<option><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
-				<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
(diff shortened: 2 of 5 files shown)


includes/Admin/Admin_Page.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Class is handling admin tools page functionality.
+ *
+ * @since n.e.x.t
+ */
+class Admin_Page {
+
+	/**
+	 * Initialize hooks.
```
felixarntz: Please remember that all methods and function descriptions should start with a third-person verb. Also, every method and function should have a `@since n.e.x.t` annotation.
```suggestion
	 * Initializes hooks.
	 *
	 * @since n.e.x.t
```

includes/Admin/Admin_Page.php
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	public function add_hooks() {
+		add_action( 'admin_menu', array( $this, 'add_page' ) );
+	}
+
+	/**
+	 * Register the admin page under the tools menu.
```
felixarntz: See above.
```suggestion
	 * Registers the admin page under the tools menu.
	 *
	 * @since n.e.x.t
```

Please update this for the other class methods as well.

includes/Admin/Admin_Page.php
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	public function add_page() {
+		add_management_page( __( 'Plugin Check', 'plugin-check' ), __( 'Plugin Check', 'plugin-check' ), 'activate_plugins', 'plugin-check', array( $this, 'render_page' ) );
```
felixarntz: Nit-pick, but maybe break this down into multiple lines? It's a long line and therefore hard to parse right now.

includes/Admin/Admin_Page.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	/**
+	 * Returns the list of plugins.
+	 *
+	 * @return array List of plugins.
+	 */
+	public function get_available_plugins() {
```
felixarntz: This should not be a public method. It's not called anywhere else, and we should only make methods public if they have to be.

templates/admin-page.php
```diff
@@ -0,0 +1,40 @@
+<?php
+/**
+ * Template for the Admin page.
+ *
+ * @package plugin-check
+ */
+
+if ( empty( $available_plugins ) ) {
+	return;
+}
```
felixarntz: Let's remove this. Imagine a site has no plugins installed. It would be a bad experience showing the user a blank admin page.

templates/admin-page.php
```diff
@@ -0,0 +1,40 @@
...
+				<option><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
+				<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
+					<option value="<?php echo esc_attr( $plugin_basename ); ?>">
+						<?php echo esc_html( $available_plugin['Name'] ); ?>
+					</option>
+				<?php } ?>
+			</select>
+
+			<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" />
+		</form>
```
felixarntz: Instead, I think this code should be wrapped in a check `if ( ! empty( $available_plugins ) )`.

We should then also add an `else` clause that outputs e.g. a paragraph like `No plugins available.`.

templates/admin-page.php
```diff
@@ -0,0 +1,40 @@
...
+if ( empty( $available_plugins ) ) {
+	return;
+}
+?>
+
+<div class="wrap">
+
+	<h1><?php esc_html_e( 'Plugin Check', 'plugin-check' ); ?></h1>
+
+	<div class="card">
```
felixarntz: Why display this content in a card?

tests/Admin/Admin_Page_Tests.php
```diff
@@ -0,0 +1,54 @@
...
+	public function test_get_plugins() {
+		$available_plugins      = get_plugins();
+		$plugin_check_base_name = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
+
+		if ( isset( $available_plugins[ $plugin_check_base_name ] ) ) {
+			unset( $available_plugins[ $plugin_check_base_name ] );
+		}
+
+		$this->assertEquals( $available_plugins, $this->admin_page->get_available_plugins() );
+	}
```
felixarntz: Instead of testing this method (which should be private, see above), we should rather test the `render_page()` method for whether it produces the correct output.

felixarntz (commented):
@vishalkakadiya Mostly looks good. A few small things to iterate on.

includes/Admin/Admin_Page.php
vishalkakadiya: I have updated this now, thank you!

includes/Admin/Admin_Page.php
vishalkakadiya: I have updated this now, thank you!

includes/Admin/Admin_Page.php
vishalkakadiya: I have updated this now, thank you!

includes/Admin/Admin_Page.php
vishalkakadiya: Done

templates/admin-page.php
vishalkakadiya: Done

templates/admin-page.php
vishalkakadiya: I thought it will look good in card (like below screenshot), but I have removed it now we can circle back on this when we actually implement output of the checks.

Screenshot 2023-02-20 at 12 39 35 PM

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Done

vishalkakadiya:
@felixarntz I have addressed your feedback. Thank you!

templates/admin-page.php
vishalkakadiya: Done

tests/Admin/Admin_Page_Tests.php
```diff
@@ -0,0 +1,68 @@
...
+		set_current_screen( $current_screen );
+	}
+
+	public function test_render_page() {
+		$available_plugins      = get_plugins();
+		$plugin_check_base_name = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
+
+		if ( isset( $available_plugins[ $plugin_check_base_name ] ) ) {
+			unset( $available_plugins[ $plugin_check_base_name ] );
+		}
```
felixarntz: What does this do? Seems unrelated to the test code below.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -0,0 +1,68 @@
...
+
+		wp_set_current_user( $admin_user );
+		set_current_screen( 'dashboard' );
+
+		$this->admin_page->add_page();
+
+		$this->assertArrayHasKey( 'plugin-check', $_parent_pages );
+		$this->assertEquals( 'tools.php', $_parent_pages['plugin-check'] );
+
+		set_current_screen( $current_screen );
```
felixarntz, line 46: This does not work reliably - if an assertion above fails, this code doesn't run.

We need to assign the `$_parent_pages` global to a local variable _before_ doing the assertions, then reset `set_current_screen`, and only at the very end run the assertions on the local variable.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -0,0 +1,68 @@
...
+		ob_start();
+		$this->admin_page->render_page();
+		$output = ob_get_contents();
+		ob_end_clean();
+
+		$this->assertStringContainsString( 'Check the Plugin', $output );
+		$this->assertStringContainsString( ' id="plugin-check__plugins"', $output );
+		$this->assertStringContainsString( ' name="plugin_check_plugins"', $output );
+		$this->assertStringContainsString( 'Select Plugin', $output );
+		$this->assertStringNotContainsString( ' name="Check it!"', $output );
```
felixarntz: It would be good to assert here that the `$output` _does not_ include an `<option>` for the `$plugin_check_base_name`, to ensure that the exclusion logic in `Admin_Page::get_available_plugins()` works correctly.

felixarntz (commented):
@vishalkakadiya A few last bits of feedback here, but looks almost good to go now.

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Removed this now, thanks!

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Done

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Done

vishalkakadiya:
@felixarntz I have addressed the last feedback. Thanks!

felixarntz (approved):
@vishalkakadiya Thanks, LGTM!

tests/Admin/Admin_Page_Tests.php
```diff
@@ -0,0 +1,66 @@
...
+
+		$parent_pages = $_parent_pages;
+
+		set_current_screen( $current_screen );
+
+		$this->assertArrayHasKey( 'plugin-check', $parent_pages );
+		$this->assertEquals( 'tools.php', $parent_pages['plugin-check'] );
+	}
+
+	public function test_render_page() {
```
joemcgill, line 52: This process of unit testing for certain strings when the template is rendered seems somewhat fragile to me, since changes to the template would also require us to update the strings these assertions are looking for. I'd prefer that we mock rendering the template file with some expected data in `$available_plugins`. We can probably override the global `$wp_object_cache` object so the 'plugins' key in the 'plugins' group contains our expected values (since that is what `get_plugins()` uses under the hood) and then confirm that the output we're getting is what we expect.

We should also include a test to ensure that we're getting the correct output whenever no plugins are availble.

joemcgill (commented):
This looks pretty good to me besides one suggestion about how we're unit testing the render callback for the admin page.

jjgrainger (approved):
Thanks @vishalkakadiya looks good!

includes/Admin/Admin_Page.php
```diff
@@ -0,0 +1,68 @@
...
+ * Class is handling admin tools page functionality.
+ *
+ * @since n.e.x.t
+ */
+class Admin_Page {
+
+	/**
+	 * Initializes hooks.
+	 */
+	public function add_hooks() {
```
felixarntz, line 22: This and the other methods in this class are missing `@since n.e.x.t` tags.

felixarntz (commented):
@vishalkakadiya While addressing @joemcgill's feedback, can you please also add the missing `@since` annotations?

tests/Admin/Admin_Page_Tests.php
jjgrainger: Thanks @joemcgill I've updated the tests to use the object cache to set the expected plugins and tested those. I've also included a test for when there are no plugins available.

includes/Admin/Admin_Page.php
jjgrainger: Thanks @felixarntz I've updated the PR with the missing `@since` tags.

jjgrainger:
Thanks @felixarntz @joemcgill as the PR has 2 approvals and the remaining feedback has been addressed I'll merge this PR.

PR #79, opened 2023-02-14
Create AJAX_Runner

### LINKED ISSUE #15: Create AJAX_Runner class
## Description 

The AJAX runner class will implement the `Abstract_Check_Runner` and its methods to use the Checks class to run checks for a plugin provided. The AJAX function to handle the plugin check requests will use the instance of the AJAX runner class and call the `run()` method to run the checks.

This issue is dependant on:
- #63 

## Acceptance Criteria

- Class `AJAX_Runner` should be created and exists at `includes/Checker`
- The  `AJAX_Runner` class extends the `Abstract_Check_Runner`.
- The class contains a `is_plugin_check()` method.
    - The `is_plugin_check()` method accepts no parameters.
    - The `is_plugin_check()` method will use the global `$_REQUEST` array, and any other relevant data, to determine if the current request is for the Plugin Checker via an AJAX request by:
        - Checking the `$_REQUEST['action']` matches `plugin_check_run_checks` (see #31) 
        - Checking the requested URL matches `admin-ajax.php`.
        - Checking the nonce passed is valid.
    - If the above conditions are correct, the method returns `true`
    - If the above conditions do not match, the method returns `false`.
- The class contains a protected `setup_checks()` method. 
    - The `setup_checks()` method accepts no parameters.
    - The `setup_checks()` method will pass the `$_REQUEST['plugin']` to the `Checks` constructor and assign the object to the `$checks` property.
    - The `setup_checks()` method will check the `$_REQUEST['checks']` array. As it is an AJAX request only 1 check should be passed in this array. This array contains an list of Check slugs to match with those in the `Checks` class.
        - The check slugs array is compared with the Check instances retrieved from the `Checks` class instance `get_checks()`
        - The array of `Check` objects are assigned to the `$checks_to_run` property.
- The class contains a `run()` method.
    - The `run()` method accepts no parameters.
    - The `run()` method will call the `Checks` class `run_checks()` method passing the `$checks` property. (`$this->checks->run_checks( $this->checks_to_run )`)
    - The `run()` method will return the instance of `Check_Results` returned by the `run_checks()` method.

## Test Coverage

- Test `is_plugin_check()` returns `true` when the `$_REQUEST` array contains the expected values for a Plugin Checker AJAX request.
- Test `is_plugin_check()` returns `false` when the `$_REQUEST` array does not contains the expected values for a Plugin Checker AJAX request.
- The `requires_universal_preperations` returns false when no Checks passed implements the `Runtime_Check` interface
- The `requires_universal_preperations` returns true when one or more Checks passed implements the `Runtime_Check` interface
- Test `prepare()` method returns `null` when the `$_REQUEST['checks']` array does not contain a `Check` that implements the `Runtime_Check` interface.
- Test `prepare()` method returns a `callable` cleanup function when the `$_REQUEST['checks']` array contains a `Check` that implements the `Runtime_Check` interface.
- Test `run()` method returns an instance of `Check_Results`.
- Test `Check_Results` returned by `run()` contains no errors/warnings when running the test `Empty_Check`
- Test `Check_Results` returned by `run()` contains errors when running the test `Error_Check`

jjgrainger:
Adds the `AJAX_Runner` class.

- Addresses acceptance criteria in #15 

Closes #15

### DIFF
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -18,7 +18,7 @@ use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 class AJAX_Runner extends Abstract_Check_Runner {
 
 	/**
-	 * An instances of the Checks class.
+	 * An instance of the Checks class.
 	 *
 	 * @since n.e.x.t
 	 * @var Checks
@@ -45,11 +45,11 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Retruns an instance of the Checks class.
+	 * Creates and returns an instance of the Checks class based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return Checks An instances of the Checks class.
+	 * @return Checks An instance of the Checks class.
 	 *
 	 * @throws Exception Thrown if the plugin main file cannot be found based on the AJAX input.
 	 */
@@ -69,7 +69,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Returns an array of Check slugs to run.
+	 * Returns an array of Check slugs to run based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -7,6 +7,7 @@
 
 namespace WordPress\Plugin_Check\Checker;
 
+use Exception;
 use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
 /**
@@ -29,18 +30,14 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return bool
+	 * @return bool Returns true if is an AJAX request for the plugin check else false.
 	 */
 	public function is_plugin_check() {
-		if ( 0 !== strpos( $_SERVER['REQUEST_URI'], '/admin-ajax.php' ) ) {
+		if ( ! wp_doing_ajax() ) {
 			return false;
 		}
 
 		if ( ! isset( $_REQUEST['action'] ) || 'plugin_check_run_checks' !== $_REQUEST['action'] ) {
-			return false;
-		}
-
-		if ( ! check_ajax_referer( 'plugin_check_run_checks' ) ) {
 			return false;
 		}
 
@@ -52,15 +49,18 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return Checks
+	 * @return Checks An instances of the Checks class.
 	 *
 	 * @throws Exception Thrown if the plugin main file cannot be found based on the AJAX input.
 	 */
 	protected function get_checks_instance() {
 		if ( ! isset( $this->checks ) ) {
+			if ( ! isset( $_REQUEST['plugin'] ) ) {
+				throw new Exception( 'Invalid plugin slug: Plugin slug must not be empty.' );
+			}
+
 			// Get the plugin name from the AJAX request.
-			$plugin_slug = isset( $_REQUEST['plugin'] ) ? $_REQUEST['plugin'] : '';
-			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin_slug );
+			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $_REQUEST['plugin'] );
 
 			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
 		}
@@ -69,18 +69,18 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Returns an array of Check instances to run.
+	 * Returns an array of Check slugs to run.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return array An array of Check instances to run.
+	 * @return array An array of Check slugs to run.
 	 */
 	protected function get_check_slugs_to_run() {
 		$checks = array();
 
 		if ( isset( $_REQUEST['checks'] ) ) {
 			// Checks are passed as a comma separated string.
-			$checks = explode( ',', $_REQUEST['checks'] );
+			$checks = wp_parse_list( $_REQUEST['checks'] );
 		}
 
 		return $checks;
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\AJAX_Runner
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
+
+/**
+ * AJAX Runner class.
+ *
+ * @since n.e.x.t
+ */
+class AJAX_Runner extends Abstract_Check_Runner {
+
+	/**
+	 * An instances of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 * @var Checks
+	 */
+	protected $checks;
+
+	/**
+	 * Checks if the current request is an AJAX request for the Plugin Checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool
+	 */
+	public function is_plugin_check() {
+		if ( 0 !== strpos( $_SERVER['REQUEST_URI'], '/admin-ajax.php' ) ) {
+			return false;
+		}
+
+		if ( ! isset( $_REQUEST['action'] ) || 'plugin_check_run_checks' !== $_REQUEST['action'] ) {
+			return false;
+		}
+
+		if ( ! check_ajax_referer( 'plugin_check_run_checks' ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Retruns an instance of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Checks
+	 *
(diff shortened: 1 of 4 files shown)


includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+
+	/**
+	 * Checks if the current request is an AJAX request for the Plugin Checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool
+	 */
+	public function is_plugin_check() {
+		if ( 0 !== strpos( $_SERVER['REQUEST_URI'], '/admin-ajax.php' ) ) {
```
felixarntz: WordPress has a function for that which is more reliable and we should probably use instead: `wp_doing_ajax()`

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+			return false;
+		}
+
+		if ( ! isset( $_REQUEST['action'] ) || 'plugin_check_run_checks' !== $_REQUEST['action'] ) {
+			return false;
+		}
+
+		if ( ! check_ajax_referer( 'plugin_check_run_checks' ) ) {
+			return false;
+		}
```
felixarntz: This doesn't seem relevant for this method. Whether the referer / nonce are valid doesn't define whether it's a plugin check request.

This validation is critical of course, but only at the point where we validate the request, not to determine whether it _is_ a plugin check request or not.

tests/Checker/AJAX_Runner_Tests.php
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+use WordPress\Plugin_Check\Checker\AJAX_Runner;
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+class AJAX_Runner_Tests extends WP_UnitTestCase {
+
+	public function test_is_plugin_check_returns_true() {
+		// Mock the AJAX request.
+		$_SERVER['REQUEST_URI'] = '/admin-ajax.php?action=plugin_check_run_checks';
```
felixarntz: See above: Instead of this (or in addition to this), you'd need to use the `wp_doing_ajax` filter to force it to `true`.

tests/Checker/AJAX_Runner_Tests.php
```diff
@@ -0,0 +1,145 @@
...
+		$runner = new AJAX_Runner();
+
+		$this->assertTrue( $runner->is_plugin_check() );
+	}
+
+	public function test_is_plugin_check_returns_false() {
+		$runner = new AJAX_Runner();
+
+		$this->assertFalse( $runner->is_plugin_check() );
+	}
```
felixarntz: Would be good to also cover specifically the case where it's still an AJAX request, but the `action` is something else.

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return Checks
+	 *
+	 * @throws Exception Thrown if the plugin main file cannot be found based on the AJAX input.
+	 */
+	protected function get_checks_instance() {
+		if ( ! isset( $this->checks ) ) {
+			// Get the plugin name from the AJAX request.
+			$plugin_slug = isset( $_REQUEST['plugin'] ) ? $_REQUEST['plugin'] : '';
```
felixarntz: Let's be strict. If this is empty or not set, this should probably throw an exception.

Letting bad requests silently go through makes debugging problems unnecessarily hard later.

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances to run.
+	 */
+	protected function get_check_slugs_to_run() {
+		$checks = array();
+
+		if ( isset( $_REQUEST['checks'] ) ) {
+			// Checks are passed as a comma separated string.
+			$checks = explode( ',', $_REQUEST['checks'] );
```
felixarntz: Better: use `wp_parse_list()` which is a little more robust.

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @var Checks
+	 */
+	protected $checks;
+
+	/**
+	 * Checks if the current request is an AJAX request for the Plugin Checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool
```
felixarntz: Please add a description (every parameter or return value should have one).

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Retruns an instance of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Checks
```
felixarntz: See above.

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+
+		return $this->checks;
+	}
+
+	/**
+	 * Returns an array of Check instances to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check instances to run.
```
felixarntz: This doc block is not quite correct, this method returns check slugs, not check instances. Just seeing that it is also incorrect in `Abstract_Check_Runner` and `CLI_Runner`, can you please fix it in those classes too?

felixarntz (commented):
@jjgrainger Overall this looks solid, left a few small points of feedback.

includes/Checker/AJAX_Runner.php
jjgrainger: Thanks @felixarntz the `Plugin_Request_Utility::get_plugin_basename_from_input` call does throw an exception when the plugin is empty/invalid. But I've added a new exception in this method to capture this earlier.

jjgrainger:
Thanks @felixarntz I've made the updates based on your feedback.

includes/Checker/AJAX_Runner.php
felixarntz: @jjgrainger Ah good point, I had missed that. I guess it's okay either way then 👍

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Retruns an instance of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Checks An instances of the Checks class.
```
felixarntz:
```suggestion
	 * @return Checks An instance of the Checks class.
```

includes/Checker/CLI_Runner.php
```diff
@@ -52,7 +52,7 @@ public function is_plugin_check() {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return Checks
+	 * @return Checks An instances of the Checks class.
```
felixarntz:
```suggestion
	 * @return Checks An instance of the Checks class.
```

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+
+		return $this->checks;
+	}
+
+	/**
+	 * Returns an array of Check slugs to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of Check slugs to run.
```
felixarntz, line 76: Noting that the same docs still need to be adjusted in the abstract base class.

felixarntz (commented):
@jjgrainger Basically LGTM, just a few nit-picks. I can address them really quickly to unblock.

felixarntz (approved):
Thanks @jjgrainger!

felixarntz:
@joemcgill Can you give this a quick sanity check and 2nd approval (just to unblock because Mukesh is traveling this week)?

includes/Checker/AJAX_Runner.php
```diff
@@ -0,0 +1,88 @@
...
+	protected function get_checks_instance() {
+		if ( ! isset( $this->checks ) ) {
+			if ( ! isset( $_REQUEST['plugin'] ) ) {
+				throw new Exception( 'Invalid plugin slug: Plugin slug must not be empty.' );
+			}
+
+			// Get the plugin name from the AJAX request.
+			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $_REQUEST['plugin'] );
+
+			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
```
joemcgill, line 65: This assumes that the plugin is always being loaded from inside the plugins directory rather than as an mu-plugin or required via code, as is often the case during unit testing setups, or some platform configurations. Is that an intentional limitation at this point?

tests/Checker/AJAX_Runner_Tests.php
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Tests for the Checks class.
+ *
+ * @package plugin-check
+ */
+
+use WordPress\Plugin_Check\Checker\AJAX_Runner;
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+class AJAX_Runner_Tests extends WP_UnitTestCase {
```
joemcgill, line 11: Just an observation, but since all of these are filtering the `wp_doing_ajax` value, is there any value in moving this shared code to a fixture that runs before each test? I also wonder if there's any value in having a test case that confirms the expected behavior if `wp_doing_ajax` is `false`?

joemcgill (approved):
A couple questions, but this all looks good to me.

includes/Checker/AJAX_Runner.php
jjgrainger: Thanks @joemcgill thats a great point and something we should probably look into in the future. For this initial milestone I believe the intention is to only check plugins found within the main plugin directory.

tests/Checker/AJAX_Runner_Tests.php
jjgrainger: Thanks @joemcgill I've added an additional test to check behaviour when `wp_doing_ajax` is `false`. We can potentially refactor the tests here in future to reduce duplicate code here.

PR #80, opened 2023-02-16
Create Runtime_Environment_Setup and object-cache.php

### LINKED ISSUE #42: Create test DB tables using object-cache.php to generate demo data
## Description

A clean environment is required for checks to run smoothly. There are a few possible scenarios that can be followed to run checks cleanly without affecting the site. So that users can run it on any site environment like production or staging. 

The approach should be not to generate test content to the primary database tables, and all modifications to the site DB should be made through test DB tables. The plugin should create test DB tables with a different prefix specially generated through the plugin to populate test data for the site. Using [wpdb::set_prefix](https://developer.wordpress.org/reference/classes/wpdb/set_prefix/), the plugin can change the base prefix of the site test DB tables. This approach can be run on the production site as well because the plugin is generating tables for demo content on need in the test tables. 

Also, this approach will only be used for runtime checks that require DB operations to run. Runtime checks should only be run if the plugin is already active on the site. The plugin can prompt the user to activate the plugin when the check is initiated.

The plugin can use its own **object-cache.php** drop-in file added into the wp-content directory which overrides the DB table prefix. It can be used to run early and pause plugins other than the one being tested. It should not actually implement an object cache. If the site is already using its own object-cache.php file, the file should be renamed and conditionally required by the plugin checker’s object-cache.php file (see below note on an additional consideration there).

This approach will need to add to the database directly and preparations will need to do clean-ups to the database content. For example, some checks would require demo content to be added to the test DB tables via preparations.

**Note**: If a site is using a persistent object cache, it should be paused for the check requests as well so that it doesn’t add any of the demo test data into the real object cache of the site. The plugin can bypass it using the object-cache.php file.

## Acceptance Criteria

- A `Runtime_Environment_Setup` class exists in `includes/Checker`
    - The `Runtime_Environment_Setup` class will be responsible for setting up the WordPress environment to run Runtime checks.
    - A public `setup()` method exists in the class and will handle the following steps:
        - Creates the demo database tables (if they do not already exist)
            - The demo tables will use a new `wppc_` prefix.
            - The demo tables will be populated with the required installation data by running the regular WP installation functions.
        - Add the object-cache.php drop-in file to the wp-content folder. If an object-cache.php file already exists, this step is skipped.
    - A public `cleanup()` method exists in the class to reverse the changes from the `setup()` method
        - Delete the demo database tables.
        - Remove the plugin checker object-cache.php file. 
- A `Plugin_Request_Utility` class exists in `includes\Utilities`
    - A public static `initialize_runner()` method  exists in the class that will handle the following steps.
        - Instantiate both runner classes
            - #14 
            - #15 
        - Call both runner classes `is_plugin_check()` methods.
        - If the `is_plugin_method` call returns true, the runners `prepare()` method is called.
        - If any check requested is a `Runtime_Check`, the `prepare()` method will update the wpdb prefix with [wpdb::set_prefix](https://developer.wordpress.org/reference/classes/wpdb/set_prefix/) to match the demo tables prefix.
    - A public static `get_runner()` method exists.
        - This method returns the runner instance that is being used for the current request, or `null` if neither is being relevant (e.g. when it is not a plugin checker request).
- Include an `object-cache.php` drop-in file.
    - The object-cache.php drop-in file will call the `Plugin_Request_Utility::initialize_runner()` method. 

### References
Below are issues/PR's from the Performance Lab plugin related to the object-cache.php drop in.

- https://github.com/WordPress/performance/issues/612
- https://github.com/WordPress/performance/pull/616
- https://github.com/WordPress/performance/issues/551
- https://github.com/WordPress/performance/pull/553

jjgrainger:
Addresses acceptance criteria in #42 

Closes #42

### DIFF
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -18,8 +18,6 @@ class Runtime_Environment_Setup {
 	 * Sets up the WordPress environment for runtime checks
 	 *
 	 * @since n.e.x.t
-	 *
-	 * @return void
 	 */
 	public function setup() {
 		global $wpdb, $wp_filesystem;
@@ -27,7 +25,7 @@ class Runtime_Environment_Setup {
 		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
 
 		// Set the new prefix.
-		$prefix = $wpdb->set_prefix( 'wppc_' );
+		$old_prefix = $wpdb->set_prefix( 'wppc_' );
 
 		// Create and populate the test database tables if they do not exist.
 		if ( 'wppc_posts' !== $wpdb->get_var( "SHOW TABLES LIKE 'wppc_posts'" ) ) {
@@ -39,7 +37,13 @@ class Runtime_Environment_Setup {
 			);
 		}
 
-		$wpdb->set_prefix( $prefix );
+		// Restore the old prefix.
+		$wpdb->set_prefix( $old_prefix );
+
+		// Return early if the plugin check object cache already exists.
+		if ( defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) && WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION ) {
+			return;
+		}
 
 		// Create the object-cache.php file.
 		if ( $wp_filesystem || WP_Filesystem() ) {
@@ -54,22 +58,26 @@ class Runtime_Environment_Setup {
 	 * Cleans up the runtime environment setup.
 	 *
 	 * @since n.e.x.t
-	 *
-	 * @return void
 	 */
 	public function cleanup() {
 		global $wpdb, $wp_filesystem;
 
 		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
 
-		$prefix = $wpdb->set_prefix( 'wppc_' );
-		$tables = $wpdb->tables();
+		$old_prefix = $wpdb->set_prefix( 'wppc_' );
+		$tables     = $wpdb->tables();
 
 		foreach ( $tables as $table ) {
 			$wpdb->query( "DROP TABLE IF EXISTS `$table`" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 		}
 
-		$wpdb->set_prefix( $prefix );
+		// Restore the old prefix.
+		$wpdb->set_prefix( $old_prefix );
+
+		// Return early if the plugin check object cache does not exists.
+		if ( ! defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) || ! WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION ) {
+			return;
+		}
 
 		// Remove the object-cache.php file.
 		if ( $wp_filesystem || WP_Filesystem() ) {
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Runtime_Environment_Setup
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+/**
+ * Class to setup the Runtime Environment for Runtime checks.
+ *
+ * @since n.e.x.t
+ */
+class Runtime_Environment_Setup {
+
+	/**
+	 * Sets up the WordPress environment for runtime checks
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	public function setup() {
+		global $wpdb, $wp_filesystem;
+
+		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+
+		// Set the new prefix.
+		$wpdb->set_prefix( 'wppc_' );
+
+		// Create and populate the test database tables if they do not exist.
+		if ( 'wppc_posts' !== $wpdb->get_var( "SHOW TABLES LIKE 'wppc_posts'" ) ) {
+			wp_install(
+				'Plugin Check',
+				'plugincheck',
+				'demo@plugincheck.test',
+				false
+			);
+		}
+
+		// Create the object-cache.php file.
+		if ( $wp_filesystem || WP_Filesystem() ) {
+			// Do not replace the object-cache.php file if it already exists.
+			if ( ! $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+				$wp_filesystem->copy( WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'object-cache.copy.php', WP_CONTENT_DIR . '/object-cache.php' );
+			}
+		}
+	}
+
+	/**
+	 * Cleans up the runtime environment setup.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	public function cleanup() {
+		global $wpdb, $wp_filesystem;
+
+		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+
+		$prefix = $wpdb->set_prefix( 'wppc_' );
+		$tables = $wpdb->tables();
+
+		foreach ( $tables as $table ) {
+			$wpdb->query( "DROP TABLE IF EXISTS `$table`" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		}
+
+		$wpdb->set_prefix( $prefix );
+
+		// Remove the object-cache.php file.
+		if ( $wp_filesystem || WP_Filesystem() ) {
+			if ( ! $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+				return;
+			}
+
+			// Check the drop-in file matches the copy.
+			$original_content = $wp_filesystem->get_contents( WP_CONTENT_DIR . '/object-cache.php' );
+			$copy_content     = $wp_filesystem->get_contents( WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'object-cache.copy.php' );
+
+			if ( $original_content && $original_content === $copy_content ) {
+				$wp_filesystem->delete( WP_CONTENT_DIR . '/object-cache.php' );
+			}
+		}
+	}
+}
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -27,7 +27,7 @@ class Runtime_Environment_Setup {
 		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
 
 		// Set the new prefix.
-		$wpdb->set_prefix( 'wppc_' );
+		$prefix = $wpdb->set_prefix( 'wppc_' );
 
 		// Create and populate the test database tables if they do not exist.
 		if ( 'wppc_posts' !== $wpdb->get_var( "SHOW TABLES LIKE 'wppc_posts'" ) ) {
@@ -38,6 +38,8 @@ class Runtime_Environment_Setup {
 				false
 			);
 		}
+
+		$wpdb->set_prefix( $prefix );
 
 		// Create the object-cache.php file.
 		if ( $wp_filesystem || WP_Filesystem() ) {
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -74,7 +74,7 @@ class Runtime_Environment_Setup {
 		// Restore the old prefix.
 		$wpdb->set_prefix( $old_prefix );
 
-		// Return early if the plugin check object cache does not exists.
+		// Return early if the plugin check object cache does not exist.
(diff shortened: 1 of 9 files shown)


object-cache.copy.php
```diff
@@ -0,0 +1,55 @@
...
+		return;
+	}
+
+	require_once $plugin_dir . 'vendor/autoload.php';
+
+	if ( class_exists( 'WordPress\Plugin_Check\Utilities\Plugin_Request_Utility' ) ) {
+		// Initialize the Check Runner class based on the request.
+		WordPress\Plugin_Check\Utilities\Plugin_Request_Utility::initialize_runner();
+
+		var_dump('initialize!');
```
joemcgill: Is this leftover from testing?

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * @return void
+	 */
+	public function setup() {
+		global $wpdb, $wp_filesystem;
+
+		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+
+		// Set the new prefix.
+		$prefix = $wpdb->set_prefix( 'wppc_' );
```
joemcgill: For clarity, I suggest making this something like `$old_prefix`, and then add an inline comment below where you restore the previous prefix—and the same in the `cleanup` method..

tests/Checker/Runtime_Environment_Setup_Tests.php
```diff
@@ -0,0 +1,59 @@
...
+		global $wp_filesystem, $wpdb;
+
+		$this->set_up_mock_filesystem();
+
+		$runtime_setup = new Runtime_Environment_Setup();
+		$runtime_setup->setup();
+		$runtime_setup->cleanup();
+
+		$this->assertTrue( 0 <= strpos( $wpdb->last_query, 'wppc_' ) );
+		$this->assertFalse( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) );
```
joemcgill, line 39: Should we add a test for when an object cache already exists?

tests/testdata/Filesystem/WP_Filesystem_MockFilesystem.php
```diff
@@ -0,0 +1,141 @@
+<?php
+/**
+ * This file needs to be in the global namespace due to how WordPress requires loading it.
```
joemcgill, line 3: 👍🏻

joemcgill (commented):
Looking good. Left a few questions/suggestions.

object-cache.copy.php
```diff
@@ -0,0 +1,55 @@
...
+ * too late to configure the test environment.
+ *
+ * This file respects any real object cache implementation the site may already
+ * be using, and it is implemented in a way that there is no risk for breakage.
+ *
+ * If you do not want the Plugin Check plugin to place this file and thus be
+ * limited to static checks, you can remove this file and set the following
+ * constant (e.g. in wp-config.php):
+ *
+ *     define( 'WP_PLUGIN_CHECK_DISABLE_OBJECT_CACHE_DROPIN', true );
```
felixarntz: I don't think we need that here, as we're not as "aggressively" placing our drop-in as the PL plugin does (which I still need to revise). I'd prefer if we remove this constant therefore.

object-cache.copy.php
```diff
@@ -0,0 +1,55 @@
...
+ * @since n.e.x.t
+ */
+
+// Set constant to be able to later check for whether this file was loaded.
+define( 'WP_PLUGIN_CHECK_PERFLAB_OBJECT_CACHE_DROPIN_VERSION', 1 );
+
+function plugin_check_initialize_runner() {
+	if ( defined( 'WP_PLUGIN_CHECK_DISABLE_OBJECT_CACHE_DROPIN' ) && WP_PLUGIN_CHECK_DISABLE_OBJECT_CACHE_DROPIN ) {
+		return;
+	}
```
felixarntz: See above, can we remove this?

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -60,4 +78,39 @@ public static function get_plugin_basename_from_input( $plugin_slug ) {
...
+	 */
+	public static function initialize_runner() {
+		$runners = array(
+			new CLI_Runner(),
+			new AJAX_Runner(),
+		);
+
+		foreach ( $runners as $runner ) {
+			if ( $runner->is_plugin_check() ) {
+				static::$cleanup = $runner->prepare();
```
felixarntz, line 96: This cleanup isn't used anywhere and also isn't accessible from the outside. I'm not sure how to best handle this yet, we should probably revisit once we are closer to having the full architecture.

Some ideas:
* Maybe we add a method to this class to run the cleanup on demand from somewhere else.
* Or maybe we immediately hook it on the `shutdown` action here so that it will run automatically.

Maybe best to add a `TODO` comment somewhere here or open a follow up issue so that we don't forget about completing this part.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,89 @@
...
+ * @since n.e.x.t
+ */
+class Runtime_Environment_Setup {
+
+	/**
+	 * Sets up the WordPress environment for runtime checks
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
```
felixarntz: `@return void` should not be present per WP coding/docs standards.
```suggestion
```

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Cleans up the runtime environment setup.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
```
felixarntz: See above.
```suggestion
```

includes/Utilities/Plugin_Request_Utility.php
felixarntz, line 96: One thing that this PR is missing but is needed per the ACs is:
> If any check requested is a `Runtime_Check`, the `prepare()` method will update the wpdb prefix with [wpdb::set_prefix](https://developer.wordpress.org/reference/classes/wpdb/set_prefix/) to match the demo tables prefix.

So far you're setting the prefix correctly when doing the overall environment preparation, but that runs in a separate request and overrides the prefix only for setting up the tables. We still need to add the logic to modify the DB prefix for any plugin check request that includes running >=1 runtime check, which is probably best to integrate into the `Abstract_Check_Runner` class (since it applies to both CLI and AJAX).

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,89 @@
...
+				'plugincheck',
+				'demo@plugincheck.test',
+				false
+			);
+		}
+
+		$wpdb->set_prefix( $prefix );
+
+		// Create the object-cache.php file.
+		if ( $wp_filesystem || WP_Filesystem() ) {
```
felixarntz, line 49: Before doing this, we should do a check for `WP_PLUGIN_CHECK_PERFLAB_OBJECT_CACHE_DROPIN_VERSION` - if that constant is set (and set to whatever is the current version per the `object-cache.copy.php` file), we don't need to even try placing it since it would already be present.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,89 @@
...
+		$tables = $wpdb->tables();
+
+		foreach ( $tables as $table ) {
+			$wpdb->query( "DROP TABLE IF EXISTS `$table`" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		}
+
+		$wpdb->set_prefix( $prefix );
+
+		// Remove the object-cache.php file.
+		if ( $wp_filesystem || WP_Filesystem() ) {
```
felixarntz, line 83: Same as above: If `WP_PLUGIN_CHECK_PERFLAB_OBJECT_CACHE_DROPIN_VERSION` is not set here, we don't need to even attempt removing this file since we know it's not present.

felixarntz (changes requested):
@jjgrainger Overall this looks very solid, just one critical (yet small) thing missing here. I left a few other smaller points of feedback.

tests/Checker/Runtime_Environment_Setup_Tests.php
jjgrainger: Thanks @joemcgill I've added additional tests here for when an object-cache.php drop-in file exists.

includes/Utilities/Plugin_Request_Utility.php
jjgrainger: Thanks @felixarntz I've added this to the `Abstract_Check_Runner` prepare method.

jjgrainger: Thanks @felixarntz I've updated the `Abstract_Check_Runner` with the database prefix

jjgrainger: Thanks @felixarntz I've created a new issue to address this here https://github.com/10up/plugin-check/issues/85

tests/Checker/Runtime_Environment_Setup_Tests.php
jjgrainger: Thanks @joemcgill I've added additional tests to cover this

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,97 @@
...
+				'demo@plugincheck.test',
+				false
+			);
+		}
+
+		// Restore the old prefix.
+		$wpdb->set_prefix( $old_prefix );
+
+		// Return early if the plugin check object cache already exists.
+		if ( defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) && WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION ) {
```
joemcgill, line 44: Does this need to check for a specific version number, in case we need to replace an older version, or is that handled elsewhere?

includes/Checker/Runtime_Environment_Setup.php
joemcgill: If we've confirmed above that the file exists, can we safely delete the file, or do we still need to confirm that the `object-cache.php` file matches the current version? Do we also need to handle the use case where an old version somehow got leftover that may not match, but should still be removed?

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -0,0 +1,97 @@
...
+		$tables     = $wpdb->tables();
+
+		foreach ( $tables as $table ) {
+			$wpdb->query( "DROP TABLE IF EXISTS `$table`" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		}
+
+		// Restore the old prefix.
+		$wpdb->set_prefix( $old_prefix );
+
+		// Return early if the plugin check object cache does not exists.
```
felixarntz:
```suggestion
		// Return early if the plugin check object cache does not exist.
```

tests/Checker/AJAX_Runner_Tests.php
```diff
@@ -10,6 +10,14 @@
 
 class AJAX_Runner_Tests extends WP_UnitTestCase {
 
+	public function tear_down() {
+		// Force reset the database prefix after runner prepare method called.
+		global $wpdb;
+		$wpdb->set_prefix( 'wp_' );
```
felixarntz: Can we use the `$table_prefix` global or (even better) access the original prefix in a `set_up()` method? We shouldn't assume that the regular prefix is `wp_`.

tests/Checker/CLI_Runner_Tests.php
```diff
@@ -10,6 +10,14 @@
 
 class CLI_Runner_Tests extends WP_UnitTestCase {
 
+	public function tear_down() {
+		// Force reset the database prefix after runner prepare method called.
+		global $wpdb;
+		$wpdb->set_prefix( 'wp_' );
```
felixarntz: See above.

felixarntz (commented):
@jjgrainger A few small last things here, but this looks almost good to go.

joemcgill (approved):
I left a couple questions, but not blocking unless you think they need to be addressed. The updates look good to me 👍🏻

includes/Checker/Runtime_Environment_Setup.php
felixarntz: @joemcgill My take is that we can ignore this for now. If/once we make changes to the file, we can update the logic, but just from a pragmatic perspective, I'm not sure we need to introduce version specific logic just yet.

There's a good chance we'll need it at some point, but generally the concept of the drop-in is to be as lightweight as possible to require as little changes as possible over time.

includes/Checker/Runtime_Environment_Setup.php
felixarntz: @joemcgill See my above reply to https://github.com/10up/plugin-check/pull/80#discussion_r1113795096 - though here this may be less relevant since we should remove the file here anytime as long as it's ours, regardless of which version it is.

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Updated

tests/Checker/AJAX_Runner_Tests.php
jjgrainger: @felixarntz I opted for the `$table_prefix` and updated both instances.

includes/Checker/Runtime_Environment_Setup.php
joemcgill: Thanks, @felixarntz. Then my approval stands as is. Resolving this convo.

felixarntz (approved):
@jjgrainger Thanks, LGTM!

PR #88, opened 2023-02-22
Fix PHPUnit warnings and echos

jjgrainger:
### Description of the Change
Cleans up some minor warnings and strings being echoed when running PHPUnit tests.

### DIFF
--- a/tests/testdata/plugins/test-plugin-with-errors/i18-usage-error-file.php
+++ b/tests/testdata/plugins/test-plugin-with-errors/i18-usage-error-file.php
@@ -5,11 +5,11 @@
 
 $city = 'Surat';
 
-printf(
+sprintf(
 	__( 'Your city is %s.', 'test-plugin-check-errors' ),
 	$city
 );
 
 $text_domain = 'test-plugin-check-errors';
 
-esc_html_e( 'Hello World!', $text_domain );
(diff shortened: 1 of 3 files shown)


tests/testdata/plugins/test-plugin-with-errors/i18-usage-error-file.php
```diff
@@ -5,11 +5,11 @@
 
 $city = 'Surat';
 
-printf(
+sprintf(
```
joemcgill, line 8: It would be useful to have an inline comment above each of these errors to explain what error(s) they are trying to demonstrate.

joemcgill (approved):
One suggestion, but seems fine to me. The other approach you could use if you _needed_ to test echoed output is to collect them to an output buffer so they don't get printed to the screen.

felixarntz (approved):
@jjgrainger LGTM, only pending @joemcgill's feedback.

tests/testdata/plugins/test-plugin-with-errors/i18-usage-error-file.php
jjgrainger: Thanks @joemcgill I'm updating these test plugins in #17 I'll add comments in that PR.

PR #89, opened 2023-02-23
Add 'check this plugin' link on the plugin list table

### LINKED ISSUE #36: Introduce a “Check this plugin” link in the Plugins list table
## Description

The plugin will add a new “Check this plugin” link to the WP Admin > Plugins list table. The link will go to the new Tools page introduced by the plugin. The link will have a plugin file (e.g., akismet/akismet.php) as a request parameter so that the Plugin Check can recognize from the request that it needs to start checks on it. 

All the processes after the link have been recognized will be the same as when the user clicks on the “Check it!” link as outlined in the AJAX functionality.

## Acceptance Criteria

- `filter_plugin_action_links()` method should exist in the `Admin_Page` class (see this class details on https://github.com/10up/plugin-check/issues/32 )
- Add `plugin_action_links` filter in the `add_hooks()` and append the `filter_plugin_action_links()` method in that hook. like below:
   
   ```add_filter( 'plugin_action_links', array( $this, 'filter_plugin_action_links' ), 10, 2 );```
   
- `filter_plugin_action_links()` method should take two arguments, `$actions` and `$plugin_file`.

- Based on the `$plugin_file`'s argument, it will return the anchor (`<a>`) link tag with query the string `page=plugin-check&plugin=$plugin_file` to point to the tools page.

- The link text will be `Check this plugin`, so the end code of the `filter_plugin_action_links()` method will look something like below:

   ```
   public function filter_plugin_action_links( $actions, $plugin_file ) {

		$actions[] = '<a href="' . esc_url( admin_url() . 'tools.php?page=plugin-check&plugin=' . $plugin_file ) . '">'.esc_html__( 'Check this plugin' , 'plugin-check').'</a>';

		return $actions;
	}
   ```
- Once the user lands on the `Plugin Check` page, the plugin's dropdown should be populated based on the `plugin` param of the query string.

## Test Coverage

**Automated Tests:**
- Write a test case to check whether the `plugin_action_links` filter is registering.
- Write a test case for the `filter_plugin_action_links()` method that it is generating the proper URL or not.

**Manual Tests:**
- Login to the admin area, go to the plugin list and check if the `Check this plugin` link is available.
- Click on any plugin's `Check this plugin` link and check the user redirecting to the `Plugin Check` page and populating the plugin list dropdown with the proper plugin name.

vishalkakadiya:
Add 'check this plugin' link on the plugin list table

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/36

Closes https://github.com/10up/plugin-check/issues/36

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -72,6 +72,8 @@ class Admin_Page {
 	/**
 	 * Add "check this plugin" link in the plugins list table.
 	 *
+	 * @since n.e.x.t
+	 *
 	 * @param array  $actions     List of actions.
 	 * @param string $plugin_file Plugin main file.
 	 * @return array
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -76,13 +76,13 @@ class Admin_Page {
 	}
 
 	/**
-	 * Add "check this plugin" link in the plugins list table.
+	 * Adds "check this plugin" link in the plugins list table.
 	 *
 	 * @since n.e.x.t
 	 *
 	 * @param array  $actions     List of actions.
 	 * @param string $plugin_file Plugin main file.
-	 * @return array
+	 * @return array The modified list of actions.
 	 */
 	public function filter_plugin_action_links( $actions, $plugin_file ) {
 
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -19,6 +19,7 @@ class Admin_Page {
 	 */
 	public function add_hooks() {
 		add_action( 'admin_menu', array( $this, 'add_page' ) );
+		add_filter( 'plugin_action_links', array( $this, 'filter_plugin_action_links' ), 10, 2 );
 	}
 
 	/**
@@ -63,6 +64,29 @@ class Admin_Page {
 
 		$available_plugins = $this->get_available_plugins();
 
+		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_STRING );
+
 		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
+	}
+
+	/**
+	 * Add "check this plugin" link in the plugins list table.
+	 *
+	 * @param array  $actions     List of actions.
+	 * @param string $plugin_file Plugin main file.
+	 * @return array
+	 */
+	public function filter_plugin_action_links( $actions, $plugin_file ) {
+
+		if ( is_plugin_active( $plugin_file ) ) {
+
+			$actions[] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( admin_url() . 'tools.php?page=plugin-check&plugin=' . $plugin_file ),
+				esc_html__( 'Check this plugin', 'plugin-check' )
+			);
+		}
+
+		return $actions;
 	}
 }
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -78,7 +78,7 @@ class Admin_Page {
 	 */
 	public function filter_plugin_action_links( $actions, $plugin_file ) {
 
-		if ( is_plugin_active( $plugin_file ) ) {
+		if ( current_user_can( 'activate_plugins' ) ) {
 
 			$actions[] = sprintf(
 				'<a href="%1$s">%2$s</a>',
--- a/templates/admin-page.php
+++ b/templates/admin-page.php
@@ -25,7 +25,7 @@
 				<select id="plugin-check__plugins" name="plugin_check_plugins">
 					<option><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
 					<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
-						<option value="<?php echo esc_attr( $plugin_basename ); ?>">
+						<option value="<?php echo esc_attr( $plugin_basename ); ?>" <?php selected( $selected_plugin_basename, $plugin_basename ); ?>>
 							<?php echo esc_html( $available_plugin['Name'] ); ?>
 						</option>
 					<?php } ?>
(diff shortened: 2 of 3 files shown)


vishalkakadiya:
@felixarntz @jjgrainger I have raised the PR for the #36, but this PR depends upon the #77 to be merged, thank you!

includes/Admin/Admin_Page.php
```diff
@@ -63,6 +64,29 @@ public function render_page() {
...
+	/**
+	 * Add "check this plugin" link in the plugins list table.
+	 *
+	 * @param array  $actions     List of actions.
+	 * @param string $plugin_file Plugin main file.
+	 * @return array
+	 */
+	public function filter_plugin_action_links( $actions, $plugin_file ) {
+
+		if ( is_plugin_active( $plugin_file ) ) {
```
felixarntz: This condition isn't needed, we should also be able to test plugins that aren't active.

The only critical differentiation that we had previously discussed is: Only for _active_ plugins we should run the runtime checks.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -63,4 +64,24 @@ public function test_render_page() {
...
+
+		$this->assertEquals(
+			sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( admin_url() . 'tools.php?page=plugin-check&plugin=' . $base_file ),
+				esc_html__( 'Check this plugin', 'plugin-check' )
+			),
+			$action_links[0]
+		);
+	}
```
felixarntz, line 147: See my comment above: You can simplify this test since the link should appear on every plugin, even when it is not active.

includes/Admin/Admin_Page.php
```diff
@@ -63,6 +64,29 @@ public function render_page() {
...
+
+			$actions[] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( admin_url() . 'tools.php?page=plugin-check&plugin=' . $plugin_file ),
+				esc_html__( 'Check this plugin', 'plugin-check' )
+			);
```
felixarntz, line 95: Maybe just to be safe, we should make sure to only add this link if the user has access to the plugin check page (i.e. has the `activate_plugins` capability, see https://github.com/10up/plugin-check/pull/77/files#diff-8f134a3d2e0f406fc9691e5163d91e0fabef5e4446a7f4c20da8cb412ef56b3eR31).

templates/admin-page.php
```diff
@@ -25,7 +25,7 @@
 				<select id="plugin-check__plugins" name="plugin_check_plugins">
 					<option><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
 					<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
-						<option value="<?php echo esc_attr( $plugin_basename ); ?>">
+						<option value="<?php echo esc_attr( $plugin_basename ); ?>" <?php selected( $selected_plugin_basename, $plugin_basename ); ?>>
```
felixarntz: Nit-pick: The `selected()` function will add a space itself if needed, so you can omit it here.
```suggestion
						<option value="<?php echo esc_attr( $plugin_basename ); ?>"<?php selected( $selected_plugin_basename, $plugin_basename ); ?>>
```

felixarntz (commented):
@vishalkakadiya Almost LGTM, just a few small things here.

Of course we still need to merge #77 first.

includes/Admin/Admin_Page.php
vishalkakadiya: Got it, thank you! 👍

includes/Admin/Admin_Page.php
vishalkakadiya: Done

templates/admin-page.php
vishalkakadiya: Done

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Done

vishalkakadiya:
@felixarntz I have addressed all your feedback. Thank you!

jjgrainger (approved):
Thanks @vishalkakadiya looks good!

includes/Admin/Admin_Page.php
```diff
@@ -69,6 +70,31 @@ private function get_available_plugins() {
 	public function render_page() {
 		$available_plugins = $this->get_available_plugins();
 
+		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_STRING );
+
 		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/admin-page.php';
 	}
+
+	/**
+	 * Add "check this plugin" link in the plugins list table.
```
felixarntz:
```suggestion
	 * Adds "check this plugin" link in the plugins list table.
```

includes/Admin/Admin_Page.php
```diff
@@ -69,6 +70,31 @@ private function get_available_plugins() {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array  $actions     List of actions.
+	 * @param string $plugin_file Plugin main file.
+	 * @return array
```
felixarntz:
```suggestion
	 * @return array The modified list of actions.
```

tests/Admin/Admin_Page_Tests.php
```diff
@@ -118,4 +119,33 @@ public function test_render_page_with_no_plugins() {
...
 		$this->assertStringNotContainsString( '<select id="plugin-check__plugins"', $output );
 	}
+
+	public function test_filter_plugin_action_links() {
+
+		$base_file = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
+
+		/** Editor check */
+		$editor = self::factory()->user->create( array( 'role' => 'editor' ) );
+		wp_set_current_user( $editor );
```
felixarntz: No need to create an editor user here. If we just leave this out so no user is set, the `activate_plugins` capability is not granted either, so that works and makes the test a bit faster.

felixarntz (commented):
@vishalkakadiya LGTM! Just a few nit-picks, but I approved this already..

includes/Admin/Admin_Page.php
vishalkakadiya: Done

includes/Admin/Admin_Page.php
vishalkakadiya: Done

tests/Admin/Admin_Page_Tests.php
vishalkakadiya: Done

vishalkakadiya:
@felixarntz I have addressed the feedback on it. Thank you!

felixarntz (approved):
Thanks @vishalkakadiya!

Issue #84, opened 2023-02-21
Handle running checks in loop and return output


vishalkakadiya:
## Description
Handle running checks one by one and return output.

This issue is dependent on the following:
- #34 
- #86 
- #90 

## Acceptance Criteria
- Update the `run_checks()` method in the `Admin_AJAX` class to handle running checks and returning a response.
    - The `run_checks` method will first verify the request using the nonce passed and If invalid will return an error response.
    - Then the `AJAX_Runner` instance is gathered by calling the `Plugin_Request_Utility::get_runner()`.
    - If the `Plugin_Request_Utility::get_runner()` returns `null` a new instance of the `AJAX_Runner` class should be created.
        - As part of creating the `AJAX_Runner` instance, the `set_check_slugs()` and `set_plugin_slug ()` methods should be called passing in the appropriate values taken from the request. 
    - Then the `AJAX_Runners` `run()` method is called and the `Check_Results` instance returned is stored in a variable.
    - If the `run()` method throws an exception, the error message is captured and returned in an error response to the client. The `admin-plugin-check.js` file will handle the error and display it on the admin page for debugging. Any further check requests will not be sent. 
    - The results are then passed to the protected `format_results()` method in the `Admin_AJAX` class. This method will handle formatting all results into an array of messages, with each message being an object with the following properties.
        - `message` - The error or warning message.
        - `type` - Either `error` or `warning`.
        - `standard` - The standard or check name.
        - `file` - The filename and path
        - `line_column` - The line and column numbers the error/warning occurred.
    - A response is returned containing the following information
        - `success` - `true` or `false` as to whether the check run was successful.
        - `results` - The array of messages returned by `format_results()`. Can be an empty array.
        - `check` - The slug of the check that was run.
        - `error` - `false` if there were no critical errors when running checks, `false` if the check run was successful.
- Update the `plugin-check-admin.js` to send requests to the `admin-ajax.php` endpoint to run checks.
    - The `checks_to_run` returned by the `plugin_check_get_checks_to_run` action will be used to send requests for each individual check.
    - The first check is removed from the array and sent in the request.
    - The AJAX request will send the following:
        - A array containing a single check slug to be run
        - The plugin name to check.
        - A valid nonce.
    - On a successful response:
        - The messages returned will be injected into the admin page using the templates setup in #35 
        - The next check in the `checks_to_run` array  is removed and used in the next request. This process will repeat until the `checks_to_run` is empty and the last check is run.
[...]

PR #91, opened 2023-02-27
Add plugin check ajax check request

### LINKED ISSUE #34: Create AJAX function and triggers and handling of progress indicator for the user
When the user triggers the checks by clicking on the “Check it” button, an admin AJAX request should be triggered to initiate the checks, and each check will run sequentially with each request in limited parameters. The response to the AJAX request will be to output checks with their respective results into the table format below the page options.

The AJAX callback function should always check for the nonces to verify the requests and should go ahead according to the verification.

There should be a visible indication of requests being made on the page for the user to see that the checks are in progress. After the successful completion of all checks, there should be an indication for the user that checks are complete and results are populated into the table. When the request has an error when running checks, it should indicate to the user that there has been some error while running the checks.

## Acceptance Criteria

- Class `Admin_AJAX` file should exist at `includes/Admin`
- Take output of the `add_management_page()` function available in the `add_page()` method of the `Admin_Page` class in to `$hook` variable.  (**Admin_Page** class introduced in https://github.com/10up/plugin-check/issues/32)
- Add hook on `load-{$hook}` and attach the `initialize_page` method to it in the `Admin_Page` class.
- `initialize_page()` method will have the `admin_enqueue_scripts` hook and callback method will be `enqueue_scripts`.
- Register new protected variable called `$admin_ajax` in the `Admin_Page` class.
- Add `constructor()` method in the `Admin_Page` class, and instantiate `Admin_AJAX` class in that.
- Add the `get_nonce()` method in the `Admin_AJAX` class. It will return the nonce, which can be used in the `Admin_Page` class.
- `Admin_Page` class should have the `enqueue_scripts()` method, which will register a new custom JS with the name `plugin-check-admin.js`, and this JS file will be placed at the top level of the plugin's directory at `plugin-check/assets`.
- The `plugin-check-admin.js` script should expose two PHP variables: the `admin-ajax` URL and `nonce` (which we get from `Admin_AJAX->get_nonce()`).
- `Admin_AJAX` will have `add_hooks()` method.
- `Admin_AJAX` will have `run_check()` method.
- `Admin_AJAX->add_hooks()` method will hook the ajax call with the `run_check()` method for only logged-in users.
- When the admin clicks on the `Check it!` button (button details available at https://github.com/10up/plugin-check/issues/32), the below process should start:
    1. JS sends an AJAX `POST` request to `admin-ajax.php` with the `plugin_check_run_check` action.
    2. `plugin_check_run_check` action will call the `run_check()` method.
        - The `run_check()` method gets the variables as below:
             - `nonce` - Basically, it is nonce to verify it with the `check_ajax_referer()` function.
        - The `run_check()` method will now respond to the success message if the nonce is valid else, return an error message in the JSON format.
    3. JS will get the details of the success or error messages and display them in the console.

## Test Coverage

**Admin_AJAX** class
- Add a test for the `Admin_Page->initialize_page()` method to check JS script only loads on the plugin-check.php only.
- Add a test for the `Admin_AJAX->add_hooks()` method.
- Add a test case for the `Admin_AJAX->get_nonce()` method.
- Add a test for the `Admin_AJAX->run_check()` method.

**Admin_Page** class
- Verified that the `plugin-check-admin.js` script is enqueued only on the `plugin-check` page.

vishalkakadiya:
Add plugin check ajax check request when user clicks on `Check it!` button.

- Addresses the acceptance criteria in https://github.com/10up/plugin-check/issues/34

Closes https://github.com/10up/plugin-check/issues/34


includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,63 @@
...
+ * @since n.e.x.t
+ */
+class Admin_AJAX {
+
+	/**
+	 * Nonce key.
+	 *
+	 * @var string
+	 */
+	private $nonce_key = '95854-random-admin-check-plugin-check-3845962';
```
joemcgill: Since this class is meant to be instantiated by classes implementing an AJAX checks, this class's constructor should except a nonce key that can be used here, in order to ensure that each nonce is unique rather than possibly reusing the same nonce key for multiple checks. Ideally, nonce keys would be somewhat descriptive so they're easier to debug when they fail.

includes/Admin/Admin_Page.php
```diff
@@ -29,13 +48,44 @@ public function add_hooks() {
...
+	 * @since n.e.x.t
+	 */
+	public function enqueue_scripts() {
+		wp_enqueue_script( 'plugin-check-admin', WP_PLUGIN_CHECK_PLUGIN_DIR_URL . 'assets/js/plugin-check-admin.js', array(), '1.0.0', true );
+
+		wp_add_inline_script(
+			'plugin-check-admin',
+			'const PLUGIN_CHECK = ' . json_encode(
+				array(
+					'ajaxUrl' => admin_url( 'admin-ajax.php' ),
```
joemcgill: If this is meant to only run in the admin, then you can make use of the `ajaxurl` global that is already defined rather than adding it here as well.

templates/admin-page.php
```diff
@@ -31,7 +31,7 @@
 					<?php } ?>
 				</select>
 
-				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" />
+				<input type="button" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="pc_check_it" />
```
joemcgill: Rather than changing the submit type, could you call `event.preventDefault()` in your event listener? I think it's also better to listen for the submit event for the whole form, rather than a click event on this button, which wouldn't be triggered by a keyboard event, like someone hitting "enter" while focused on the form.

templates/admin-page.php
```diff
@@ -17,12 +17,12 @@
...
 			<form>
 				<h2>
-					<label class="title" for="plugin-check__plugins">
+					<label class="title" for="pc_plugins">
 						<?php esc_html_e( 'Check the Plugin', 'plugin-check' ); ?>
 					</label>
 				</h2>
 
-				<select id="plugin-check__plugins" name="plugin_check_plugins">
+				<select id="pc_plugins" name="plugin_check_plugins">
```
joemcgill: Is there a reason why these IDs need to change?

joemcgill (changes requested):

templates/admin-page.php
vishalkakadiya: @joemcgill Not any specific reason, but it is changed for simplicity, as we have button ID `pc_check_it`, so same way, I have changed the ID of the select box, as the short IDs are good to handle in the JS file. And I made this template, and that ID is not yet used anywhere, so we can change it, as it will not break anything. 🙂

includes/Admin/Admin_AJAX.php
vishalkakadiya: Make sense, I have updated this now. Thank you!

includes/Admin/Admin_Page.php
vishalkakadiya: Fixed now.

templates/admin-page.php
vishalkakadiya: Updated it now.

vishalkakadiya:
@joemcgill  I have addressed your feedback on this PR and left one comment. You can review it now, thanks!

joemcgill (approved):
Thanks for the updates 👍🏻

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,75 @@
...
+		$this->nonce_key = $nonce_key;
+	}
+
+	/**
+	 * Initializes hooks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function add_hooks() {
+		add_action( 'wp_ajax_plugin_check_run_check', array( $this, 'run_check' ) );
```
jjgrainger: Based on discussions with @felixarntz in #84 this should be renamed to `wp_ajax_plugin_check_run_checks` and the method here to `run_checks`

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,75 @@
...
+
+		return wp_create_nonce( $this->nonce_key );
+	}
+
+	/**
+	 * Run check.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function run_check() {
```
jjgrainger: Method name should be updated to `run_checks` based on discussions in #84

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,45 @@
...
+	const pluginsList   = document.getElementById( 'pc_plugins' );
+
+	checkItButton.addEventListener( 'click', (e) => {
+
+		e.preventDefault();
+
+		const pluginCheckData = new FormData();
+
+		// Collect the data to pass along for generating a check results.
+		pluginCheckData.append( 'action', 'plugin_check_run_check' );
```
jjgrainger: This is to be updated to `plugin_check_run_checks` based on #84

tests/Admin/Admin_AJAX_Tests.php
```diff
@@ -0,0 +1,32 @@
...
+	protected $admin_ajax;
+
+	public function set_up() {
+		parent::set_up();
+		$this->admin_ajax = new Admin_AJAX( 'run-check-via-check-it-button' );
+	}
+
+	public function test_add_hooks() {
+		$this->admin_ajax->add_hooks();
+		$this->assertEquals( 10, has_action( 'wp_ajax_plugin_check_run_check', array( $this->admin_ajax, 'run_check' ) ) );
```
jjgrainger: Again, updating the action to `wp_ajax_plugin_check_run_checks`

jjgrainger (changes requested):
@vishalkakadiya looks good, just left some minor changes to the action/method name based on discussions in #84

jjgrainger (approved):

jjgrainger:
Hi @felixarntz I've made some minor changes to the method/action names. Please can you review and merge if approved. Thanks!

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * Sets the nonce key for verification.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $nonce_key Nonce key.
+	 */
+	public function __construct( $nonce_key ) {
+		$this->nonce_key = $nonce_key;
+	}
```
felixarntz: Why do we need this constructor parameter and property? That seems unnecessary to me.

I think we can make whichever string we want to choose a `const` in this class.

includes/Admin/Admin_Page.php
```diff
@@ -15,13 +15,32 @@
...
+	 */
+	protected $admin_ajax;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function __construct() {
+		$this->admin_ajax = new Admin_AJAX( 'run-check-via-check-it-button' );
```
felixarntz: See above.

includes/Admin/Admin_Page.php
```diff
@@ -30,13 +49,44 @@ public function add_hooks() {
...
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+	}
+
+	/**
+	 * Loads the check's script.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function enqueue_scripts() {
+		wp_enqueue_script( 'plugin-check-admin', WP_PLUGIN_CHECK_PLUGIN_DIR_URL . 'assets/js/plugin-check-admin.js', array(), '1.0.0', true );
```
felixarntz: The version number on assets is used for cache invalidation, so following the WP core approach we should set this to the current plugin version just in case: `WP_PLUGIN_CHECK_VERSION`

templates/admin-page.php
felixarntz: Same question from my end, I think the original IDs here were more clear and had less of a chance for conflict. There's a higher chance that some other plugin uses the `pc_` prefix than another plugin using the `plugin-check_` prefix. Not a big deal, but I would advise to change it back. Maybe we should make it a bit more clear also what the element is for, e.g. `plugin-check__plugins-dropdown`.

templates/admin-page.php
```diff
@@ -31,7 +31,7 @@
 					<?php } ?>
 				</select>
 
-				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" />
+				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="pc_check_it" />
```
felixarntz: See above. More importantly though, let's not use `check_it`, that's not really a good description of the element. More something like "submit" or "submit button". How about `plugin-check__submit`?

includes/Admin/Admin_Page.php
```diff
@@ -30,13 +49,44 @@ public function add_hooks() {
...
-		add_management_page(
+		$hook = add_management_page(
 			__( 'Plugin Check', 'plugin-check' ),
 			__( 'Plugin Check', 'plugin-check' ),
 			'activate_plugins',
 			'plugin-check',
 			array( $this, 'render_page' )
 		);
+
+		add_action( "load-{$hook}", array( $this, 'initialize_page' ) );
```
felixarntz, line 62: Not strictly necessary, but a good practice would be to return the `$hook` from this method. This way you can cover the last bit here in the `test_add_page()` unit test (which is currently not tested).

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,43 @@
+( function ( data ) {
+	const checkItButton = document.getElementById( 'pc_check_it' );
+	const pluginsList   = document.getElementById( 'pc_plugins' );
```
felixarntz: For both of these, it's always a good practice to check whether those elements are actually present on the page. Of course we can assume they are, but there's always a chance that something goes wrong on the page (e.g. due to another plugin), and then we don't want to be causing a JS error.

I think if one of these returns `null`, we should return early. Maybe we can also do a quick `console.error` to inform that the element is missing.

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,43 @@
...
+			{
+				method: 'POST',
+				credentials: 'same-origin',
+				body: pluginCheckData
+			}
+		)
+		.then(
+			( response ) => {
+				if ( ! response.ok ) {
+					throw new Error(`[${response.message}]`);
```
felixarntz: Missing spaces here according to WP coding standards.

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,43 @@
...
+				return response.json();
+			}
+		)
+		.then(
+			( data ) => {
+				console.log( data.data.message );
+			}
+		)
+		.catch(
+			( error ) => { console.log( error ); }
```
felixarntz: Nit-pick: Better to use `console.error`.

Also, per my other comment, we should check if a response was still returned from the server. If yes, we can extract the first error message from it and display it (`response.data[ 0 ].message`). This needs to be optional though, as an error could also occur just because e.g. there is no network connection, in which case there won't be a response.

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * Run checks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function run_checks() {
+		$nonce = filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_STRING );
+
+		if ( ! wp_verify_nonce( $nonce, $this->nonce_key ) ) {
+			wp_send_json_error();
```
felixarntz: We should pass an actual error message here. Maybe pass a `new WP_Error( ... )` to `wp_send_json_error()` as that function will automatically put the data into the error response in a consistent format.

We also need to send an error status code. Probably a `403` works well here.

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		wp_send_json_success(
+			array(
+				'message' => 'Verified!',
```
felixarntz: Let's use translated user-facing strings. We'll probably replace this one eventually, but that's a practice we should always stick to.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -50,6 +49,11 @@ public function test_add_page() {
 		$this->assertEquals( 'tools.php', $parent_pages['plugin-check'] );
```
felixarntz, line 48: See above, by using the `$hook` as return value, we can also assert here that the hook to initialize the page was added.

felixarntz (changes requested):
@jjgrainger @vishalkakadiya Left some feedback here. Overall this looks solid, but there are a few things to clean up.

jjgrainger:
Thanks @felixarntz I've addressed the changes you've requested. This is ready for another review.

templates/admin-page.php
```diff
@@ -13,32 +13,32 @@
...
-				<select id="plugin-check__plugins" name="plugin_check_plugins">
+				<select id="plugin-check__plugins-dropdown" name="plugin_check_plugins">
 					<option><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
-					<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
+					<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) : ?>
 						<option value="<?php echo esc_attr( $plugin_basename ); ?>"<?php selected( $selected_plugin_basename, $plugin_basename ); ?>>
 							<?php echo esc_html( $available_plugin['Name'] ); ?>
 						</option>
-					<?php } ?>
+					<?php endforeach; ?>
```
felixarntz: Please revert this, we should use proper curly braces instead of that old syntax, which even in templates is discouraged at this point.

templates/admin-page.php
```diff
@@ -13,32 +13,32 @@
 
 	<div class="plugin-check-content">
 
-		<?php if ( ! empty( $available_plugins ) ) { ?>
+		<?php if ( ! empty( $available_plugins ) ) : ?>
```
felixarntz: See my other comment, this should be kept as is.

templates/admin-page.php
```diff
@@ -13,32 +13,32 @@
...
+				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="plugin-check__submit" />
 			</form>
 
-		<?php } else { ?>
+		<?php else : ?>
 
 			<h2><?php esc_html_e( 'No plugins available.', 'plugin-check' ); ?></h2>
 
-		<?php } ?>
+		<?php endif; ?>
```
felixarntz: See above, this should not be changed.

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,66 @@
...
+	 * Run checks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function run_checks() {
+		$nonce = filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_STRING );
+
+		if ( ! wp_verify_nonce( $nonce, self::NONCE_KEY ) ) {
+			wp_send_json_error(
+				new WP_Error( 403, __( 'Invalid nonce', 'plugin-check' ) ),
```
felixarntz: The error code should not be an HTTP status code, but a slug like string that identifies the error.
```suggestion
				new WP_Error( 'invalid-nonce', __( 'Invalid nonce', 'plugin-check' ) ),
```

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,49 @@
...
+				return response.json();
+			}
+		)
+		.then(
+			( data ) => {
+				if ( ! data.success ) {
+					throw new Error( data.data[0].message );
+				}
+
+				console.log('data', data.data.message );
```
felixarntz: WP coding standards violation due to missing space.

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,49 @@
...
+		.then(
+			( response ) => {
+				return response.json();
+			}
+		)
+		.then(
+			( data ) => {
+				if ( ! data.success ) {
+					throw new Error( data.data[0].message );
+				}
```
felixarntz: Have you tested that this works? What if no response is returned, e.g. when no network connection (thus `data` could be `null` or `undefined`)? Let's use defensive coding here to make triple-sure that `data.success` and `data.data[ 0 ].message` are set before accessing it. We should avoid errors in our error handling logic :)

Last but not least: Have you tested that this logic is actually reached when an error happens? Since the response code is a 403, will it get to the code here, or would it immediately run into the `catch` logic without even running this?

felixarntz (commented):
@jjgrainger A few small follow up things, most importantly about the error handling. I'm just unsure whether the approach you have in JS works, so I want to make sure you've tested this.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz I've updated and tested this based on your feedback. To answer some of your questions here:

- From what I understand, if there is no network connection the response is not handled and by the rest of the code and a generic console error is displayed.
- Recent updates check the `response.ok` property before moving on and will throw an error with the status code and text. This now includes the 403 error we return. Originally, the custom invalid nonce error message would be console logged.
- I have tested for different scenarios (500 error, 403 error, returning no data, successful response) and the code is working.

felixarntz: Thanks @jjgrainger for confirming!

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,53 @@
...
+				return response.json();
+			}
+		)
+		.then(
+			( data ) => {
+				if ( ! data.success && ( ! data.data || ! data.data.message ) ) {
+					throw new Error( 'Response contains no data' );
+				}
+
+				console.log( data.data.message );
```
felixarntz: Doesn't this need to use `data.data[ 0 ].message`? You had that here before which I thought was right, but now you got rid of the `[ 0 ]`.

I tested this and it is indeed not working. You need to also still include a condition that covers the error case. If the 403 error from the AJAX action is returned, the `data.data[ 0 ].message` in the error response should be logged via `console.error`. That is not the case right now.

felixarntz (commented):
@jjgrainger The error logging is still not working as expected. You now added another check to ensure the expected success data is present, but at the same time got rid of the (also needed) check that logs the error message if there is one.

To clarify, we need to cover 4 scenarios here:
1. Request successful and `data.data.message` is present.
2. Request seemingly successful and `data.data.message` is not present (least likely to ever happen in reality).
3. Request error and `data.data[ 0 ].message` is present (which means our AJAX endpoint returned an error).
4. Request error and `data.data[ 0 ].message` is not present (which means something out of our control failed, e.g. the user is offline).

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz 

This final console log here is the final success message and all previous errors are handled. Going over the scenarios here:

1. Request successful and `data.data.message` is present - This is this final console log.
2. Request seemingly successful and `data.data.message` is not present (least likely to ever happen in reality). - This is the error thrown above 'Response contains no data', however I can see how this doesn't catch if `success` is true but the message is not set. I can update this.
3. Request error and `data.data[ 0 ].message` is present (which means our AJAX endpoint returned an error) - This is now captured earlier in the response when checking `response.ok`. This will be set to `false` for any error status including the 403 we return.
4. Request error and `data.data[ 0 ].message` is not present (which means something out of our control failed, e.g. the user is offline). - I don't think we can handle the logic with code here, if something like a network failure happens, the response is not handled and a generic console error is logged.

For point 2, I can make the updates here to capture when `success` is true but data/message is undefined.

For point 3,  I can include an additional check as the point in the logic to use the message we return instead of the generic status message?

Let me know if you have any questions.

jjgrainger: Thanks @felixarntz I've updated the code based on your feedback with comments to highlight the different scenarios/errors that are captured here.

mukeshpanchal27 (commented):
@jjgrainger The checked Ajax request run if we don't select any plugin in dropdown can we restrict it so it will only run when the option has value?

templates/admin-page.php
```diff
@@ -31,7 +31,7 @@
 					<?php } ?>
 				</select>
 
-				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" />
+				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="plugin-check__submit" />
```
mukeshpanchal27:
```suggestion
				<input class="button" type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="plugin-check__submit" />
```

Use button class for better design of button.

includes/Admin/Admin_AJAX.php
```diff
@@ -0,0 +1,66 @@
...
+	 * Creates and returns the nonce.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function get_nonce() {
+		return wp_create_nonce( self::NONCE_KEY );
+	}
+
+	/**
+	 * Run checks.
```
mukeshpanchal27:
```suggestion
	 * Runs checks.
```

jjgrainger:
> @jjgrainger The checked Ajax request run if we don't select any plugin in dropdown can we restrict it so it will only run when the option has value?

Thanks @mukeshpanchal27 I agree with the idea here, but I think this is something to address later as we have additional functionality to build for the WP Admin and may want to tackle validation later as part of Milestone 1 Review updates.

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,65 @@
...
+				method: 'POST',
+				credentials: 'same-origin',
+				body: pluginCheckData
+			}
+		)
+		.then(
+			( response ) => {
+				if ( ! response.ok ) {
+					throw new Error( `${response.status}: ${response.statusText}` );
+				}
```
felixarntz: This here is what throws a wrench in things. You already error out due to the bad response code, which makes it impossible to even reach the code below where we expose the more specific JSON error that is present.

I think by removing this our error logging works correctly. Requests that come back with a bad response are automatically logged in the console anyway, so no need to separately log the status code in our logic.

assets/js/plugin-check-admin.js
```diff
@@ -0,0 +1,65 @@
...
+
+				return response.json();
+			}
+		)
+		.then(
+			( data ) => {
+				if ( ! data.success ) {
```
felixarntz, line 40: Before checking this, we should check whether `data` even is an object, particularly when removing the above `response.ok` check. If the response is empty, then this line here would cause a JS error.
```suggestion
				if ( ! data ) {
					throw new Error( 'Response is empty or does not contain valid JSON' );
				}

				if ( ! data.success ) {
```

templates/admin-page.php
```diff
@@ -31,7 +31,7 @@
 					<?php } ?>
 				</select>
 
-				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" />
+				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="plugin-check__submit" class="button" />
```
felixarntz: Maybe make it a primary button, since it is the primary action on the page. We'll revise design later, so this is a nit-pick at this point, but a good practice.
```suggestion
				<input type="submit" value="<?php esc_attr_e( 'Check it!', 'plugin-check' ); ?>" id="plugin-check__submit" class="button button-primary" />
```

felixarntz (commented):
@jjgrainger One last problem with the error logging still remains. I still don't see our error message logged. Did you test this? Did your browser maybe behave differently than mine?

includes/Admin/Admin_Page.php
```diff
@@ -15,28 +15,88 @@
...
+			array(),
+			WP_PLUGIN_CHECK_VERSION,
+			true
+		);
+
+		wp_add_inline_script(
+			'plugin-check-admin',
+			'const PLUGIN_CHECK = ' . json_encode(
+				array(
+					'ajaxUrl' => admin_url( 'admin-ajax.php' ),
```
joemcgill: Nitpick, but I think sending this value is unnecessary since `ajaxurl` is already set as a global in the admin.

jjgrainger:
Thanks @felixarntz I've addressed the feedback here and this is ready for review

felixarntz (approved):
Thanks @jjgrainger, LGTM!

PR #95, opened 2023-02-28
Update Runner classes to validate checks

### LINKED ISSUE #93: Update Runner Classes to accept and validate the checks to run and plugin file
Currently, the `AJAX_Runner` and `CLI_Runner` classes determine the checks to run and the plugin to check early in the WordPress load process. In doing so, these parameters are not collected or validated via the standard WordPress functions available. This issue looks to update both runners to handle these variables via WordPress functions, such as the process request `$args` in the CLI. 

## Acceptance Criteria

- Update the `Abstract_Check_Runner` with the following:
    - Add a constructor used to determine if the class was instantiated early. This is done by checking the `did_action( 'muplugins_loaded' )` - If the MU Plugins have not been loaded yet, the runner was initialized early. This boolean value should be stored in a class property to be used later.
    - A new public `set_check_slugs_to_run( array $check_slugs )` method is created. This accepts an array of check slugs to run that will be manually set during the AJAX/CLI requests. The value passed in stored as a class property to be used later.
    - A new public `set_plugin_basename( string $plugin_basename )` method is created. This accepts the plugin basename to be checked and will be manually set during the AJAX/CLI requests. The value passed in stored as a class property to be used later.
    - Add a protected `get_checks_instance()` method. This method will replace the abstract methods created in the AJAX/CLI runner classes. This method will check if the class was initialized early and if the plugin basename was set manually. If both are true, the class will compare the plugin basename passed with those determined by the request (`determine_plugin_basename()`). If the set values do not match the computed values an exception is thrown. If the values match, the Check instance is created using the plugin basename.
    - Add a protected `get_check_slugs_to_run()` method. This method will replace the abstract methods created in the AJAX/CLI runner classes. This method will check if the class was initialized early and if the check slugs array was set manually. If both are true, the class will compare the checks passed with those determined by the request (`determine_check_slugs_to_run ()`). If the values do not match an exception is thrown. If the values match, the array of check slugs is returned.
    - Rename the abstract `get_checks_instance()` method to `determine_plugin_basename()`
    - Rename the abstract `get_check_slugs_to_run()` method to `determine_check_slugs_to_run()`
- Update the `AJAX_Runner` class with the following:
    - Remove the `get_checks_instance()` method
    - Rename the `get_check_slugs_to_run()` method to `determine_check_slugs_to_run()`
    - Add a protected `determine_plugin_basename()` method. This method will return the basename of the plugin based on the AJAX request sent.
- Update the `CLI_Runner` class with the following:
    - Remove the `get_checks_instance()` method
    - Rename the `get_check_slugs_to_run()` method to `determine_check_slugs_to_run()`
    - Add a protected `determine_plugin_basename()` method. This method will return the basename of the plugin based on the arguments passed in the CLI command.

## Test Coverage
- Add unit tests for the `AJAX_Runner`:
    - Test an exception is thrown when the plugin basename in the request does not match the manually set plugin basename.
    - Test an exception is thrown when the check slugs array in the request does not match the manually set array of check slug.
- Add unit tests for the `CLI_Runner`:
    - Test an exception is thrown when the plugin basename in the cli args does not match the manually set plugin basename.
    - Test an exception is thrown when the check slugs array in the cli args does not match the manually set array of check slug.

jjgrainger:
Addresses acceptance criteria in #93 

Closes #93

### DIFF
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -44,18 +44,18 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Returns the plugin slug based on the request.
+	 * Returns the plugin parameter based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return string The plugin slug.
+	 * @return string The plugin parameter.
 	 *
-	 * @throws Exception Thrown if the plugin slug is invalid.
+	 * @throws Exception Thrown if the plugin parameter is empty.
 	 */
 	protected function get_plugin_param() {
 		if ( ! isset( $_REQUEST['plugin'] ) ) {
 			throw new Exception(
-				__( 'Invalid plugin slug: Plugin slug must not be empty.', 'plugin-check' )
+				__( 'Invalid plugin: Plugin parameter must not be empty.', 'plugin-check' )
 			);
 		}
 
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -8,7 +8,6 @@
 namespace WordPress\Plugin_Check\Checker;
 
 use Exception;
-use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
 /**
  * AJAX Runner class.
@@ -45,27 +44,22 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Creates and returns an instance of the Checks class based on the request.
+	 * Returns the plugin basename based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return Checks An instance of the Checks class.
+	 * @return string The plugin basename.
 	 *
-	 * @throws Exception Thrown if the plugin main file cannot be found based on the AJAX input.
+	 * @throws Exception Thrown if the plugin basename is invalid.
 	 */
-	protected function get_checks_instance() {
-		if ( ! isset( $this->checks ) ) {
-			if ( ! isset( $_REQUEST['plugin'] ) ) {
-				throw new Exception( 'Invalid plugin slug: Plugin slug must not be empty.' );
-			}
-
-			// Get the plugin name from the AJAX request.
-			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $_REQUEST['plugin'] );
-
-			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
+	protected function determine_plugin_basename() {
+		if ( ! isset( $_REQUEST['plugin'] ) ) {
+			throw new Exception(
+				__( 'Invalid plugin slug: Plugin slug must not be empty.', 'plugin-check' )
+			);
 		}
 
-		return $this->checks;
+		return sanitize_text_field( $_REQUEST['plugin'] );
 	}
 
 	/**
@@ -75,7 +69,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	 *
 	 * @return array An array of Check slugs to run.
 	 */
-	protected function get_check_slugs_to_run() {
+	protected function determine_check_slugs_to_run() {
 		$checks = array();
 
 		if ( isset( $_REQUEST['checks'] ) ) {
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -44,15 +44,15 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	}
 
 	/**
-	 * Returns the plugin basename based on the request.
+	 * Returns the plugin slug based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return string The plugin basename.
+	 * @return string The plugin slug.
 	 *
-	 * @throws Exception Thrown if the plugin basename is invalid.
+	 * @throws Exception Thrown if the plugin slug is invalid.
 	 */
-	protected function determine_plugin_basename() {
+	protected function get_plugin_param() {
 		if ( ! isset( $_REQUEST['plugin'] ) ) {
 			throw new Exception(
 				__( 'Invalid plugin slug: Plugin slug must not be empty.', 'plugin-check' )
@@ -69,7 +69,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	 *
 	 * @return array An array of Check slugs to run.
 	 */
-	protected function determine_check_slugs_to_run() {
+	protected function get_check_slugs_param() {
 		$checks = array();
 
 		if ( isset( $_REQUEST['checks'] ) ) {
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -7,6 +7,8 @@
 
 namespace WordPress\Plugin_Check\Checker;
 
+use Exception;
+use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
 
 /**
@@ -41,6 +43,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	protected $plugin_basename;
 
 	/**
+	 * An instance of the Checks class.
+	 *
+	 * @since n.e.x.t
+	 * @var Checks
+	 */
+	protected $checks;
+
+	/**
 	 * Determines if the current request is intended for the plugin checker.
 	 *
 	 * @since n.e.x.t
@@ -50,13 +60,13 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	abstract public function is_plugin_check();
 
 	/**
-	 * Creates and returns an instance of the Checks class based on the request.
+	 * Returns the plugin basename based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
 	 * @return Checks An instance of the Checks class.
 	 */
-	abstract protected function get_checks_instance();
+	abstract protected function determine_plugin_basename();
 
 	/**
 	 * Returns an array of Check slugs to run based on the request.
@@ -65,7 +75,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 *
 	 * @return array An array of Check slugs.
 	 */
-	abstract protected function get_check_slugs_to_run();
+	abstract protected function determine_check_slugs_to_run();
 
 	/**
 	 * Sets whether the runner class was initialized early.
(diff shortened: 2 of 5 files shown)


includes/Checker/AJAX_Runner.php
```diff
@@ -45,27 +44,22 @@ public function is_plugin_check() {
...
-		if ( ! isset( $this->checks ) ) {
-			if ( ! isset( $_REQUEST['plugin'] ) ) {
-				throw new Exception( 'Invalid plugin slug: Plugin slug must not be empty.' );
-			}
-
-			// Get the plugin name from the AJAX request.
-			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $_REQUEST['plugin'] );
-
-			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
+	protected function determine_plugin_basename() {
```
felixarntz: This method name is now misleading based on what it is supposed to do. Since you removed the usage of the `get_plugin_basename_from_input` call, this method now returns simply the value of the parameter, without sanitizing it as the plugin basename (i.e. it could also be just a plugin slug).

Either we need to bring that call back into here to _actually_ return a plugin basename, or we need to rename this method to be something like `get_plugin_param()`. Maybe the latter is indeed cleaner, i.e. keeping the logic itself like you have it, since it means we only have in the child class what is actually different between AJAX and CLI.

includes/Checker/AJAX_Runner.php
```diff
@@ -75,7 +69,7 @@ protected function get_checks_instance() {
 	 *
 	 * @return array An array of Check slugs to run.
 	 */
-	protected function get_check_slugs_to_run() {
+	protected function determine_check_slugs_to_run() {
```
felixarntz: This looks right, but now based on the above comment, maybe an even better name would be `get_check_slugs_param()`? This clarifies that the value is just the parameter value, not necessarily a valid list of checks (i.e. validation would happen outside of these methods).

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -26,13 +60,13 @@ abstract class Abstract_Check_Runner implements Check_Runner {
...
 	/**
-	 * Creates and returns an instance of the Checks class based on the request.
+	 * Returns the plugin basename based on the request.
 	 *
 	 * @since n.e.x.t
 	 *
 	 * @return Checks An instance of the Checks class.
 	 */
-	abstract protected function get_checks_instance();
+	abstract protected function determine_plugin_basename();
```
felixarntz: See above regarding the name of this method.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -41,7 +75,38 @@ abstract protected function get_checks_instance();
 	 *
 	 * @return array An array of Check slugs.
 	 */
-	abstract protected function get_check_slugs_to_run();
+	abstract protected function determine_check_slugs_to_run();
```
felixarntz: See above regarding the name of this method.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -41,7 +75,38 @@ abstract protected function get_checks_instance();
...
+	}
+
+	/**
+	 * Sets the check slugs to be run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $check_slugs An array of check slugs to be run.
+	 */
+	public function set_check_slugs_to_run( array $check_slugs ) {
```
felixarntz: Let's shorten this to `set_check_slugs()` the "to run" bit isn't really helpful.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -41,7 +75,38 @@ abstract protected function get_checks_instance();
...
+		$this->check_slugs = $check_slugs;
```
felixarntz, line 108: Before setting this, we should validate in here if the class was initialized early; validation should happen before setting a value, not only when it is being accessed.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -41,7 +75,38 @@ abstract protected function get_checks_instance();
...
+
+	/**
+	 * Sets the plugin basename to be checked.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_basename The plugin basename to be checked.
+	 */
+	public function set_plugin_basename( $plugin_basename ) {
+		$this->plugin_basename = $plugin_basename;
```
felixarntz: See above, we should validate in here, not only when getting the value.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -167,4 +232,61 @@ protected function get_checks_to_run() {
...
+		$check_slugs = $this->determine_check_slugs_to_run();
+
+		if ( $this->initialized_early && isset( $this->check_slugs ) ) {
+			// Compare the plugin basename to see if there was an error.
+			if ( $this->check_slugs !== $check_slugs ) {
+				throw new Exception(
+					__( 'Invalid checks: The checks to run do not match the original request.', 'plugin-check' )
+				);
+			}
+		}
```
felixarntz: See above, this validation shouldn't happen here. If `$this->check_slugs` is already set, we should simply return it.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -167,4 +232,61 @@ protected function get_checks_to_run() {
...
+		if ( $this->initialized_early && isset( $this->plugin_basename ) ) {
+			// Compare the plugin basename to see if there was an error.
+			if ( $this->plugin_basename !== $plugin_basename ) {
+				throw new Exception(
+					__( 'Invalid plugin basename: The plugin basename does not match the original request.', 'plugin-check' )
+				);
+			}
+		}
+
+		$plugin_file  = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin_basename );
```
felixarntz: There are a few problems in this logic:
* As mentioned above, the validation shouldn't happen here, but rather when the plugin basename is set.
* The `determine_plugin_basename()` does not actually return a plugin basename, it just returns the parameter value, without yet having sanitized it to be a plugin basename. This makes sense for that function (see also the renaming suggestions above), but it means the logic here is incorrect, as we need to make it a plugin basename via `Plugin_Request_Utility::get_plugin_basename_from_input()` _before_ comparing it to the plugin basename set.
* We should only even call the abstract method if `$this->plugin_basename` is not set. If it's set, we can just use that value, no need to run the extra logic.

Nit-pick: Please also clarify the variable names: What you have here as `$plugin_basename` is just the raw parameter value (could be a plugin slug, or plugin basename, or even an empty string), and what you have here as `$plugin_file` is actually the `$plugin_basename`.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -167,4 +232,61 @@ protected function get_checks_to_run() {
...
+	/**
+	 * Returns the check slugs to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of check slugs to run.
+	 *
+	 * @throws Exception Thrown if the checks do not match those in the original request.
+	 */
+	protected function get_check_slugs_to_run() {
```
felixarntz: Let's also shorten this accordingly to `get_check_slugs()`.

includes/Checker/CLI_Runner.php
```diff
@@ -52,20 +52,21 @@ public function is_plugin_check() {
...
 	 *
-	 * @throws Exception Thrown if the plugin main file cannot be found based on the CLI input.
+	 * @throws Exception Thrown if the plugin basename is invalid.
 	 */
-	protected function get_checks_instance() {
-		if ( ! isset( $this->checks ) ) {
-			// Get the plugin name from the command line arguments.
-			$plugin_slug = isset( $_SERVER['argv'][3] ) ? $_SERVER['argv'][3] : '';
-			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin_slug );
+	protected function determine_plugin_basename() {
```
felixarntz: See above regarding method name.

includes/Checker/CLI_Runner.php
```diff
@@ -75,7 +76,7 @@ protected function get_checks_instance() {
 	 *
 	 * @return array An array of Check slugs to run.
 	 */
-	protected function get_check_slugs_to_run() {
+	protected function determine_check_slugs_to_run() {
```
felixarntz: See above regarding method name.

felixarntz (changes requested):
@jjgrainger This looks close, there are a few things that need to be fixed though around sanitization and validation. Also a few method renaming suggestions that just came up in my review based on the logic you're using, I like your idea of using the abstract methods only to get the parameter value but not more.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -41,7 +75,60 @@ abstract protected function get_checks_instance();
...
+	/**
+	 * Sets the plugin slug to be checked.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_slug The plugin slug to be checked.
+	 *
+	 * @throws Exception Thrown if the plugin main file does not match the original request.
+	 */
+	public function set_plugin_slug( $plugin_slug ) {
```
felixarntz: Currently the logic in this checks against the plugin parameter, which can be either a plugin slug or a plugin basename, so the name of this method, the parameter, and also the internal property name is misleading.

We should probably align with the rest of the new method names here to clarify that this is simply the `plugin` param. In other words, let's call this `set_plugin( $plugin )`, and let's clarify on both the method parameter and the `$this->plugin` property declaration that the value can be either a plugin slug or a plugin basename.

This effectively means wherever we use this value we need to run it through `Plugin_Request_Utility::get_plugin_basename_from_input` to sanitize it to a basename. I think that's okay since it only happens in one place anyway.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -167,4 +254,40 @@ protected function get_checks_to_run() {
...
+	 * @return Checks An instance of the Checks class.
+	 *
+	 * @throws Exception Thrown if the plugin slug is invalid.
+	 */
+	protected function get_checks_instance() {
+		if ( isset( $this->checks ) ) {
+			return $this->checks;
+		}
+
+		$plugin_slug     = isset( $this->plugin_slug ) ? $this->plugin_slug : $this->get_plugin_param();
```
felixarntz: See above regarding naming, let's use `$plugin` instead of `$plugin_slug` to account for the ambiguity of what this value can contain (it's not necessarily a slug).

includes/Checker/CLI_Runner.php
```diff
@@ -48,24 +48,25 @@ public function is_plugin_check() {
...
 
-			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
+		if ( empty( $plugin_slug ) ) {
+			throw new Exception(
+				__( 'Invalid plugin slug: Plugin slug must not be empty.', 'plugin-check' )
+			);
 		}
 
-		return $this->checks;
+		return $plugin_slug;
```
felixarntz: See above regarding naming of this. Also can you similarly adjust the exception message to refer to "plugin parameter" rather than "plugin slug"?

includes/Checker/AJAX_Runner.php
```diff
@@ -45,27 +44,22 @@ public function is_plugin_check() {
...
-			}
-
-			// Get the plugin name from the AJAX request.
-			$plugin_file = Plugin_Request_Utility::get_plugin_basename_from_input( $_REQUEST['plugin'] );
-
-			$this->checks = new Checks( WP_PLUGIN_DIR . '/' . $plugin_file );
+	protected function get_plugin_param() {
+		if ( ! isset( $_REQUEST['plugin'] ) ) {
+			throw new Exception(
+				__( 'Invalid plugin slug: Plugin slug must not be empty.', 'plugin-check' )
```
felixarntz: See other similar comment, can you adjust the exception message to refer to "plugin parameter" rather than "plugin slug"?

felixarntz (approved):
@jjgrainger Thanks, the code looks great. A few last naming quirks, but I'll already approve this as following the renaming it can be merged.

vishalkakadiya (approved):
LGTM!

PR #97, opened 2023-03-14
Create get checks to run AJAX endpoint

### LINKED ISSUE #90: Create plugin_check_get_checks_to_run endpoint
An AJAX endpoint will be created to determine what checks to run for the AJAX Runner. This request will be called first in order to determine what checks will be run. The endpoint will accept a list of check slugs and the plugin slug to be checked. If the plugin to check is not active **only** static checks can be run.

## Acceptance Criteria
- Crate a new AJAX endpoint for the action `plugin_check_get_checks_to_run` In the `Admin_Ajax` class.
- A new WordPress action is created for `wp_ajax_plugin_check_get_checks_to_run` that calls a `get_checks_to_run` method on the `Admin_Ajax` class.
    - The `get_checks_to_run` method will first verify the request using the nonce passed. If the nonce is invalid an error response is returned.
    - The check slugs passed in the `$_REQUEST['checks']` will be processed to determine what checks to run by filtering the checks slugs passed against the `Checks` class `get_checks()` method. Only check slugs that exist as keys in the `get_checks()` array will be valid. If the `$_REQUEST['checks']` is empty, all check slugs from the `get_checks()` will be used.
    - Before returning the check slugs the plugin slug passed will be checked to see if the plugin is active. This is important as if the plugin is not active no runtime checks can be run.
        - If `$_REQUEST['checks']` is empty and the plugin is not active, return only the static checks.
        - If `$_REQUEST['checks']` contains any runtime check and the plugin is not active, return an error as that check would be disallowed.
    - The check slugs array or error message is returned in the response.
- Update the `plugin-check-admin.js` so the first request sent after submitting the form is sent to the `admin-ajax.php` endpoint with the `plugin_check_get_checks_to_run` action.
    - The AJAX request will send the following:
        - An array of selected checks. Currently, there are no options to select checks so an empty array will always be sent.
        - The plugin name. Required to create the context for the `Checks` class.
        - A valid nonce.
    - If the request fails, a user-friendly error message is displayed on the admin page and no further AJAX requests will be made.
    - If the request is successful, the array of Check slugs will be stored in a `checks_to_run` variable to be used by future requests.

jjgrainger:
Addresses acceptance criteria for #90 

Closes #90


assets/js/plugin-check-admin.js
```diff
@@ -33,20 +39,62 @@
...
+		)
+		.then(
+			( data ) => {
+				return handleDataErrors(data);
+			}
+		)
+		.then(
+			( data ) => {
 				// If the response is successful and there is no message in the response.
 				if ( ! data.data || ! data.data.message ) {
```
joemcgill, line 96: Should this get consolidated to the `handleDataErrors` function?

assets/js/plugin-check-admin.js
```diff
@@ -33,20 +39,62 @@
...
+				pluginChecker.plugin = data.data.plugin;
+				pluginChecker.checks = data.data.checks;
+
+				runChecks();
+			}
+		)
+		.catch(
+			( error ) => { console.error( error ) }
+		);
+	} );
```
felixarntz: Can we move all this into a function to get the checks to run like you already did with the other code? Now that this file becomes more complex, we should start working with logical units.

I also think it would be best to return a `Promise` from the new function with the resulting data to decouple the two functions. This new function shouldn't set the data on the `pluginChecker` const, and it shouldn't call `runChecks()`. It should only return the data, and then the event listener can take care of the specific bits to connect the two.

includes/Admin/Admin_AJAX.php
```diff
@@ -43,24 +47,111 @@ public function get_nonce() {
...
+			}
+
+			// If all checks are requested, filter out any runtime checks.
+			$checks_to_run = array_filter(
+				$checks_to_run,
+				function ( $check ) {
+					return ! $check instanceof Runtime_Check;
+				}
+			);
+		}
```
felixarntz: This logic here is a bit confusing to follow. Wouldn't it be simpler to split the two paths clearly between `if ( ! empty( $checks ) ) { ... } else { ... }`? That way you avoid the additional check for `empty( $checks )` which is part of what makes this logic hard to follow.

includes/Admin/Admin_AJAX.php
```diff
@@ -43,24 +47,111 @@ public function get_nonce() {
...
+	 * @since n.e.x.t
+	 */
+	protected function verify_nonce() {
+		$nonce = filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_STRING );
+
+		if ( ! wp_verify_nonce( $nonce, self::NONCE_KEY ) ) {
+			wp_send_json_error(
+				new WP_Error( 'invalid-nonce', __( 'Invalid nonce', 'plugin-check' ) ),
+				403
+			);
```
felixarntz: Nit-pick, but sending the JSON response would be more appropriate to handle in the main AJAX methods, as it decouples this specific logic from sending JSON. You could return `bool|WP_Error` from this method here, and then where it is called, it would use `wp_send_json_error()` if the returned value `is_wp_error()`.

includes/Admin/Admin_AJAX.php
```diff
@@ -43,24 +47,111 @@ public function get_nonce() {
...
 		);
 	}
+
+	/**
+	 * Verify the nonce passed in the request.
+	 *
+	 * @since n.e.x.t
+	 */
+	protected function verify_nonce() {
+		$nonce = filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_STRING );
```
felixarntz: Similarly here, this is directly coupled to the AJAX request, so it would be better to keep it in the individual AJAX callback methods. You could pass `$nonce` as a parameter to this method to keep it decoupled.

felixarntz (commented):
@jjgrainger This looks great so far. A few suggestions for decoupling the functions/methods more.

includes/Admin/Admin_AJAX.php
```diff
@@ -43,24 +47,111 @@ public function get_nonce() {
...
+	}
+
+	/**
+	 * Run checks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function run_checks() {
+		// Verify the nonce before continuing.
+		$this->verify_nonce();
```
joemcgill: Nice catch

assets/js/plugin-check-admin.js
```diff
@@ -8,15 +8,21 @@
 		return;
 	}
 
-	checkItButton.addEventListener( 'click', (e) => {
-		e.preventDefault();
+	// Create a state object for the plugin checker.
+	const pluginChecker = {
+		plugin: false,
+		checks_to_run: [],
+	};
```
joemcgill: Rather than storing data from the first AJAX response to send to use during the `runChecks` function, I wonder if we could pass these properties directly to the `runChecks` function as params?

joemcgill (approved):
I've got a couple comments, but this is generally looking good. I'm not sure I totally get my head around why this process needs to happen in separate AJAX calls, rather than all at once when the plugin is being checked, but I don't see that as a blocker to continuing to push forward with this approach. Nice work.

felixarntz (approved):
Thanks @jjgrainger, LGTM!

PR #100, opened 2023-03-16
CLI Runtime environment setup/cleanup

### LINKED ISSUE #87: CLI Runtime Environment Setup/Cleanup
This issue will implement the runtime environment setup and cleanup for CLI requests. This will utilise the `Runtime_Environment_Setup` class to setup the WP environment before checks are run, and cleanup the environment once all checks are complete.

## Acceptance Criteria

- Add a new `before_wp_load` hook with `WP_CLI::add_hook()` to implement the object-cache.php drop-in file.
    - The hook will manually (not using the WP filesystem) add the object-cache.php drop-in file to the wp-content folder.
- Create a `cli.php` file in the plugin root.
    - This file will be used to setup the `Plugin_Check_Command` both early and late in the WordPress load process.
    - The file will first check if the `Plugin_Check_Command` class exists.
        - If it does not this file is running before WordPress has loaded and the command will need to be setup manually.
            - This is done by loading the plugin autoload file so the class is available.
            - A new `Plugin_Context` for the plugin checker is created and passed to the `Plugin_Check_Command` constructor.
            - The `Plugin_Check_Command` is then registered to the WordPress CLI.
        - If the `Plugin_Check_Command` class does exist.
             - A check is made to see if the `$context` variable exists, which will be set in the `Plugin_Main::add_hooks()` method.
             - A new instance of the `Plugin_Check_Command` is created passing the `$context` to the constructor.
             - The `Plugin_Check_Command` is then registered to the WordPress CLI.
- Update the `Plugin_Main::add_hooks()` method to `require_once` the `cli.php` file, instead of adding the command directly to WP_CLI.
- Update the `wp plugin check` to run the `Runtime_Environment_Setup::setup()` method at the start of the commands run process if any runtime check is to be performed. This is to setup the database tables required for runtime checks.
- Update the `wp plugin check` to run the `Runtime_Environment_Setup::cleanup()` method at the end of the commands run process. This will remove the database tables and object-cache.php drop-in file if they exists.
- When running the `wp plugin check` command the `--require` flag can be used to include the `cli.php` file early to create the object-cache.php file and register the command.

jjgrainger:
Addresses acceptance criteria for #87 

Closes #87

### DIFF
--- a/cli.php
+++ b/cli.php
@@ -12,39 +12,37 @@ if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 	return;
 }
 
-define( 'WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH', dirname( __FILE__ ) );
-
-// Check if the plugin autoloading is setup.
+// Check if the plugin autoloading is set up.
 if ( ! class_exists( 'WordPress\Plugin_Check\CLI\Plugin_Check_Command' ) ) {
 	// Check the autoload file exists.
-	if ( ! file_exists( WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/vendor/autoload.php' ) ) {
-		\WP_CLI::error( 'Plugin Check autoloaded not found.' );
+	if ( ! file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
+		WP_CLI::error( 'Plugin Check autoloaded not found.' );
 		return;
 	}
 
 	// Load the Composer autoloader.
-	require_once WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/vendor/autoload.php';
+	require_once __DIR__ . '/vendor/autoload.php';
 }
 
 if ( ! isset( $context ) ) {
-	$context = new WordPress\Plugin_Check\Plugin_Context( WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/plugin-check.php' );
+	$context = new WordPress\Plugin_Check\Plugin_Context( __DIR__ . '/plugin-check.php' );
 }
 
 // Create the CLI command instance and add to WP CLI.
 $plugin_command = new WordPress\Plugin_Check\CLI\Plugin_Check_Command( $context );
-\WP_CLI::add_command( 'plugin', $plugin_command );
+WP_CLI::add_command( 'plugin', $plugin_command );
 
-
-// Add hook to setup the object-cache.php drop-in file.
-\WP_CLI::add_hook( 'before_wp_load', 'wp_plugin_check_cli_object_cache' );
 
 /**
- * Sets up the object-cache.php drop-in file.
+ * Adds hook to set up the object-cache.php drop-in file.
  *
  * @since n.e.x.t
  */
-function wp_plugin_check_cli_object_cache() {
-	if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
-		copy(  WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+WP_CLI::add_hook(
+	'before_wp_load',
+	function() {
+		if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+			copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+		}
 	}
-}
+);
--- a/cli.php
+++ b/cli.php
@@ -41,8 +41,18 @@ WP_CLI::add_command( 'plugin', $plugin_command );
 WP_CLI::add_hook(
 	'before_wp_load',
 	function() {
-		if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
-			copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+		if ( empty( $_SERVER['argv'] ) || 3 > count( $_SERVER['argv'] ) ) {
+			return;
+		}
+
+		if (
+			'wp' === substr( $_SERVER['argv'][0], -2 ) &&
+			'plugin' === $_SERVER['argv'][1] &&
+			'check' === $_SERVER['argv'][2]
+		) {
+			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+				copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+			}
 		}
 	}
 );
--- a/cli.php
+++ b/cli.php
@@ -8,6 +8,9 @@
  * @since n.e.x.t
  */
 
+use WordPress\Plugin_Check\Plugin_Context;
+use WordPress\Plugin_Check\CLI\Plugin_Check_Command;
+
 if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 	return;
 }
@@ -16,7 +19,7 @@ if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 if ( ! class_exists( 'WordPress\Plugin_Check\CLI\Plugin_Check_Command' ) ) {
 	// Check the autoload file exists.
 	if ( ! file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
-		WP_CLI::error( 'Plugin Check autoloaded not found.' );
+		WP_CLI::error( 'Plugin Check autoloader not found.' );
 		return;
 	}
 
@@ -25,11 +28,11 @@ if ( ! class_exists( 'WordPress\Plugin_Check\CLI\Plugin_Check_Command' ) ) {
 }
 
 if ( ! isset( $context ) ) {
-	$context = new WordPress\Plugin_Check\Plugin_Context( __DIR__ . '/plugin-check.php' );
+	$context = new Plugin_Context( __DIR__ . '/plugin-check.php' );
 }
 
 // Create the CLI command instance and add to WP CLI.
-$plugin_command = new WordPress\Plugin_Check\CLI\Plugin_Check_Command( $context );
+$plugin_command = new Plugin_Check_Command( $context );
 WP_CLI::add_command( 'plugin', $plugin_command );
 
 
@@ -51,7 +54,9 @@ WP_CLI::add_hook(
 			'check' === $_SERVER['argv'][2]
 		) {
 			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
-				copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+				if ( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
+					WP_CLI::error( 'Unable to copy object-cache.php file.' );
+				}
 			}
 		}
 	}
--- a/cli.php
+++ b/cli.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Sets up the CLI command early in the WordPress load process.
+ *
+ * This is necessary to setup the environment to perform runtime checks.
+ *
+ * @package plugin-check
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
+	return;
+}
+
+define( 'WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH', dirname( __FILE__ ) );
+
+// Check if the plugin autoloading is setup.
+if ( ! class_exists( 'WordPress\Plugin_Check\CLI\Plugin_Check_Command' ) ) {
+	// Check the autoload file exists.
+	if ( ! file_exists( WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/vendor/autoload.php' ) ) {
+		\WP_CLI::error( 'Plugin Check autoloaded not found.' );
+		return;
+	}
+
+	// Load the Composer autoloader.
+	require_once WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/vendor/autoload.php';
+}
+
+if ( ! isset( $context ) ) {
+	$context = new WordPress\Plugin_Check\Plugin_Context( WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/plugin-check.php' );
+}
+
+// Create the CLI command instance and add to WP CLI.
+$plugin_command = new WordPress\Plugin_Check\CLI\Plugin_Check_Command( $context );
+\WP_CLI::add_command( 'plugin', $plugin_command );
+
+
+// Add hook to setup the object-cache.php drop-in file.
+\WP_CLI::add_hook( 'before_wp_load', 'wp_plugin_check_cli_object_cache' );
+
+/**
+ * Sets up the object-cache.php drop-in file.
+ *
+ * @since n.e.x.t
+ */
+function wp_plugin_check_cli_object_cache() {
+	if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+		copy(  WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+	}
+}
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -10,7 +10,6 @@ namespace WordPress\Plugin_Check\Admin;
 use WP_Error;
 use Exception;
 use WordPress\Plugin_Check\Checker\AJAX_Runner;
-use WordPress\Plugin_Check\Checker\Checks;
 use WordPress\Plugin_Check\Checker\Runtime_Check;
 use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 /**
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -60,7 +60,7 @@ class Admin_AJAX {
 			wp_send_json_error( $valid_nonce, 403 );
 		}
 
-		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
+		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
 
 		$runner = Plugin_Request_Utility::get_runner();
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -69,7 +69,7 @@ class Admin_AJAX {
 
 		try {
 			$checks_to_run = $runner->get_checks_to_run();
-		} catch( Exception $error ) {
+		} catch ( Exception $error ) {
 			wp_send_json_error(
 				new WP_Error( 'invalid-checks', $error->getMessage() ),
 				403
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -9,6 +9,7 @@ namespace WordPress\Plugin_Check\Admin;
 
 use WP_Error;
 use Exception;
+use WordPress\Plugin_Check\Checker\AJAX_Runner;
 use WordPress\Plugin_Check\Checker\Checks;
 use WordPress\Plugin_Check\Checker\Runtime_Check;
 use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
@@ -59,53 +60,20 @@ class Admin_AJAX {
 			wp_send_json_error( $valid_nonce, 403 );
 		}
 
-		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
+		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
 
-		// Attempt to get the plugin basename based on the request.
+		$runner = new AJAX_Runner();
+		$runner->set_check_slugs( $checks );
+		$runner->set_plugin( $plugin );
+
 		try {
-			$plugin_basename = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin );
-		} catch ( Exception $error ) {
+			$checks_to_run = $runner->get_checks_to_run();
+		} catch( Exception $error ) {
 			wp_send_json_error(
-				new WP_Error( 'invalid-plugin', $error->getMessage() ),
+				new WP_Error( 'invalid-checks', $error->getMessage() ),
 				403
 			);
-		}
-
-		$plugin_active = is_plugin_active( $plugin_basename );
-
-		// Create the checks instance.
-		$checks_instance = new Checks( WP_PLUGIN_DIR . '/' . $plugin_basename );
-		$all_checks      = $checks_instance->get_checks();
-
-		// If specific checks are requested to run.
-		if ( ! empty( $checks ) ) {
-			// Get the check instances based on the requested checks.
-			$checks_to_run = array_intersect_key( $all_checks, array_flip( $checks ) );
-
-			// Return an error if at least 1 runtime check is requested to run against an inactive plugin.
-			if ( ! $plugin_active && $this->has_runtime_check( $checks_to_run ) ) {
-				wp_send_json_error(
-					new WP_Error(
-						'inactive-plugin',
-						__( 'Runtime checks cannot be run against inactive plugins.', 'plugin-check' )
-					),
-					403
-				);
-			}
-		} else {
-			// Run all checks for the plugin.
-			$checks_to_run = $all_checks;
-
-			// Only run static checks if the plugin is inactive.
-			if ( ! $plugin_active ) {
-				$checks_to_run = array_filter(
-					$checks_to_run,
-					function ( $check ) {
-						return ! $check instanceof Runtime_Check;
-					}
-				);
-			}
 		}
 
 		wp_send_json_success(
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -63,12 +63,26 @@ class Admin_AJAX {
 		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
 
-		$runner = new AJAX_Runner();
-		$runner->set_check_slugs( $checks );
-		$runner->set_plugin( $plugin );
+		$runner = Plugin_Request_Utility::get_runner();
+
+		if ( is_null( $runner ) ) {
+			$runner = new AJAX_Runner();
+		}
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
+			wp_send_json_error(
+				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ) ),
+				403
+			);
+		}
 
 		try {
-			$checks_to_run = $runner->get_checks_to_run();
+			$runner->set_check_slugs( $checks );
+			$runner->set_plugin( $plugin );
+
+			$plugin_basename = $runner->get_plugin_basename();
+			$checks_to_run   = $runner->get_checks_to_run();
 		} catch ( Exception $error ) {
 			wp_send_json_error(
 				new WP_Error( 'invalid-checks', $error->getMessage() ),
@@ -78,7 +92,7 @@ class Admin_AJAX {
 
 		wp_send_json_success(
 			array(
-				'plugin' => $plugin,
+				'plugin' => $plugin_basename,
 				'checks' => array_keys( $checks_to_run ),
 			)
 		);
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -131,7 +131,7 @@ class Plugin_Check_Command {
 		}
 
 		if ( $this->has_runtime_check( $checks_to_run ) ) {
-			WP_CLI::line( __( 'Setup runtime environment.', 'plugin-check' ) );
+			WP_CLI::line( __( 'Setting up runtime environment.', 'plugin-check' ) );
 			$runtime_setup = new Runtime_Environment_Setup();
 			$runtime_setup->setup();
 		}
@@ -144,7 +144,7 @@ class Plugin_Check_Command {
 
 			if ( isset( $runtime_setup ) ) {
 				$runtime_setup->cleanup();
-				WP_CLI::line( __( 'Cleanup runtime environment.', 'plugin-check' ) );
+				WP_CLI::line( __( 'Cleaning up runtime environment.', 'plugin-check' ) );
 			}
 
 			WP_CLI::error( $error->getMessage() );
@@ -154,7 +154,7 @@ class Plugin_Check_Command {
 
 		if ( isset( $runtime_setup ) ) {
 			$runtime_setup->cleanup();
-			WP_CLI::line( __( 'Cleanup runtime environment.', 'plugin-check' ) );
+			WP_CLI::line( __( 'Cleaning up runtime environment.', 'plugin-check' ) );
 		}
 
 		// Get errors and warnings from the results.
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -164,7 +164,7 @@ class Plugin_Check_Command {
 		// Run checks against the plugin.
 		try {
 			$runner->set_plugin( $plugin );
-			$runner->set_check_slugs( $checks );
+			$runner->set_check_slugs( array_keys( $checks_to_run ) );
 			$result = $runner->run();
 		} catch ( Exception $error ) {
 			Plugin_Request_Utility::destroy_runner();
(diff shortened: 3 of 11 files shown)


jjgrainger:
Hi @felixarntz 

I've setup a draft PR to review the changes here. I have the command working with the Runtime setup, however I think there is an opportunity to refactor some elements of the code here, in particular the latest additions to the `Plugin_Check_Command`.

Currently this uses the same logic found in the `Admin_AJAX::get_checks_to_run()` method to determine the checks to be run, if those checks can be run and filtering out runtime checks if the plugin is inactive.

I wasn't sure how best to solve this but my current thinking is we could potentially move a lot of this logic over to the `Abstract_Check_Runner`?
- The `get_checks_to_run` method could handle the additional logic to either remove runtime checks from the array or throw an exception if the plugin is inactive based on the checks requested.
- Rename the `requires_universal_preparations` method to `has_runtime_check` and make it public so it can be used in the `Plugin_Check_Command` to determine if we need to use the `Runtime_Environment_Setup`
- Update the logic for the `Admin_AJAX` to use a `AJAX_Runner` instance to handle returning the check slugs to run in the `get_checks_to_run` method.

Let me know your thoughts here. Thanks

includes/CLI/Plugin_Check_Command.php
```diff
@@ -103,6 +106,46 @@ public function check( $args, $assoc_args ) {
 		$plugin = isset( $args[0] ) ? $args[0] : '';
 		$checks = wp_parse_list( $options['checks'] );
 
+		try {
+			// Attempt to get the plugin basename based on the request.
+			$plugin_basename = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin );
+			$checks_instance = new Checks( WP_PLUGIN_DIR . '/' . $plugin_basename );
+			$all_checks      = $checks_instance->get_checks();
+			$plugin_active   = is_plugin_active( $plugin_basename );
```
spacedmonkey: What about network plugins? Maybe use  `is_plugin_active_for_network`?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -103,6 +106,46 @@ public function check( $args, $assoc_args ) {
...
+				}
+			} else {
+				// Run all checks for the plugin.
+				$checks_to_run = $all_checks;
+
+				// Only run static checks if the plugin is inactive.
+				if ( ! $plugin_active ) {
+					$checks_to_run = array_filter(
+						$checks_to_run,
+						function ( $check ) {
```
spacedmonkey: static function

felixarntz:
@jjgrainger Thanks for pointing those out!

I agree with your points 1 and 3, it would be great if you could incorporate those changes.

Regarding point 2, I'm not sure about that one yet, it doesn't seem as appropriate, although I also agree with you we should avoid the duplicate logic. Maybe for now let's keep that one as is (i.e. acknowledge that we'll for now have the duplicate logic for the "has runtime check" concept), and we can revisit this as part of the architecture review.

includes/CLI/Plugin_Check_Command.php
jjgrainger: Thanks @spacedmonkey multisite support is something we'll be looking to add in the next milestone. For now we're just working on single site support.

jjgrainger:
Thanks @felixarntz I've made the updates for 1 and 3. This is now ready for review.

cli.php
```diff
@@ -0,0 +1,50 @@
...
+
+// Add hook to setup the object-cache.php drop-in file.
+\WP_CLI::add_hook( 'before_wp_load', 'wp_plugin_check_cli_object_cache' );
+
+/**
+ * Sets up the object-cache.php drop-in file.
+ *
+ * @since n.e.x.t
+ */
+function wp_plugin_check_cli_object_cache() {
```
felixarntz: We should preferably avoid declaring anything in this file. I would suggest you make this a closure for the hook, no need to have a named function.

cli.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @package plugin-check
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
+	return;
+}
+
+define( 'WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH', dirname( __FILE__ ) );
```
felixarntz: No need to define a constant for this, we can simply use `__DIR__`.

cli.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Sets up the object-cache.php drop-in file.
+ *
+ * @since n.e.x.t
+ */
+function wp_plugin_check_cli_object_cache() {
+	if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+		copy(  WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
+	}
```
felixarntz: This code now runs unconditionally regardless of which command is invoked. I think at a minimum, we should run it only when it's the `wp plugin check` command.

We could hack this in here by manually checking `$_SERVER['argv']` like we do elsewhere, but maybe there is a cleaner way through a command-specific hook. I am not entirely sure whether any of the command-specific hooks would be executed before WordPress loads, but it's worth trying. Please review https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-add-hook/ and https://make.wordpress.org/cli/handbook/guides/commands-cookbook/. Maybe we can do something with the `before_invoke` or `before_add_command` hooks? Or would those already run too late? Would be great if you could explore that.

cli.php
```diff
@@ -0,0 +1,50 @@
...
+	require_once WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/vendor/autoload.php';
+}
+
+if ( ! isset( $context ) ) {
+	$context = new WordPress\Plugin_Check\Plugin_Context( WP_PLUGIN_CHECK_CLI_PLUGIN_DIR_PATH . '/plugin-check.php' );
+}
+
+// Create the CLI command instance and add to WP CLI.
+$plugin_command = new WordPress\Plugin_Check\CLI\Plugin_Check_Command( $context );
+\WP_CLI::add_command( 'plugin', $plugin_command );
```
felixarntz: Nit-pick, but we are in global namespace here, so no need for the backslash. This applies throughout the file.

includes/Admin/Admin_AJAX.php
```diff
@@ -59,55 +60,22 @@ public function get_checks_to_run() {
 			wp_send_json_error( $valid_nonce, 403 );
 		}
 
-		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
+		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
```
felixarntz: I don't think changing this parameter to be a string makes sense. Just because in CLI we can only provide a string we shouldn't force the same limitation to apply to our AJAX handlers. Providing an array of checks makes more sense. Furthermore, updating this only here will also create a bug due to inconsistency with how the parameter is handled in `AJAX_Runner`.

I think we could easily call the `wp_parse_list` on the CLI `--checks` parameter only. Or am I missing something, what is your rationale for changing this here?

includes/Admin/Admin_AJAX.php
```diff
@@ -59,55 +60,22 @@ public function get_checks_to_run() {
...
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
 
-		// Attempt to get the plugin basename based on the request.
+		$runner = new AJAX_Runner();
```
felixarntz: We shouldn't unconditionally instantiate this; we need to use the same approach we're taking elsewhere (e.g. in the CLI command below), first call `Plugin_Request_Utility::get_runner()` to see if the runner was already instantiated etc.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -118,15 +121,42 @@ public function check( $args, $assoc_args ) {
 			);
 		}
 
-		// Run checks against the plugin.
 		try {
-			$runner->set_plugin( $plugin );
 			$runner->set_check_slugs( $checks );
+			$runner->set_plugin( $plugin );
```
felixarntz, line 125: Why was the order of this reversed? Seems random? Or any particular reason for it?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -118,15 +121,42 @@ public function check( $args, $assoc_args ) {
...
+
+			$checks_to_run = $runner->get_checks_to_run();
+		} catch ( Exception $error ) {
+			WP_CLI::error( $error->getMessage() );
+		}
+
+		if ( $this->has_runtime_check( $checks_to_run ) ) {
+			WP_CLI::line( __( 'Setup runtime environment.', 'plugin-check' ) );
```
```suggestion
			WP_CLI::line( __( 'Setting up runtime environment.', 'plugin-check' ) );
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -118,15 +121,42 @@ public function check( $args, $assoc_args ) {
...
+			}
+
 			WP_CLI::error( $error->getMessage() );
 		}
 
+		Plugin_Request_Utility::destroy_runner();
+
+		if ( isset( $runtime_setup ) ) {
+			$runtime_setup->cleanup();
+			WP_CLI::line( __( 'Cleanup runtime environment.', 'plugin-check' ) );
```
```suggestion
			WP_CLI::line( __( 'Cleaning up runtime environment.', 'plugin-check' ) );
```

includes/Admin/Admin_AJAX.php
```diff
@@ -59,55 +60,22 @@ public function get_checks_to_run() {
...
-					function ( $check ) {
-						return ! $check instanceof Runtime_Check;
-					}
-				);
-			}
-		}
-
 		wp_send_json_success(
 			array(
 				'plugin' => $plugin,
```
felixarntz: Should we use `AJAX_Runner::get_plugin_basename()` here instead of the raw unsanitized parameter?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -50,6 +50,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	protected $checks;
 
+	/**
+	 * The plugin slug or basename to check.
```
felixarntz: This is not correct. This will always be the plugin basename, never the plugin slug. The "plugin slug or basename" part only applies to the raw unsanitized parameter, i.e. the `$plugin` property, so I think it's worth keeping that in there to clarify.

includes/Traits/URL_Aware.php
```diff
@@ -154,7 +154,27 @@ protected function go_to( $url ) {
...
+			if ( $t->publicly_queryable && ! empty( $t->query_var ) ) {
+				$GLOBALS['wp']->add_query_var( $t->query_var );
+			}
+		}
+
+		foreach ( get_post_types( array(), 'objects' ) as $t ) {
+			if ( is_post_type_viewable( $t ) && ! empty( $t->query_var ) ) {
+				$GLOBALS['wp']->add_query_var( $t->query_var );
+			}
+		}
```
felixarntz, line 177: This change seems unrelated. Can you explain what this does and why it's needed? Is there any overlap with the main change here?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -243,16 +251,38 @@ private function get_shared_preparations( array $checks ) {
...
+	 *
+	 * @throws Exception Thrown exception when a runtime check is requested and the plugin inactive.
 	 */
-	protected function get_checks_to_run() {
-		$check_slugs = $this->get_check_slugs();
-		$all_checks  = $this->get_checks_instance()->get_checks();
-
-		if ( empty( $check_slugs ) ) {
-			return $all_checks;
+	public function get_checks_to_run() {
```
felixarntz, line 257: One thing that this method is missing (and FWIW the code has been missing before as well) is to check that the early initialization / runtime environment preparation is even possible. If for example the `object-cache.php` file cannot be placed, we won't be able to prepare the environment, in which case we also cannot do runtime checks.

So I think we should check here if the runner instance was _not_ initialized early, and if it wasn't, we need to exclude all runtime checks. In the case the check slugs were provided that include a runtime check, we should then also throw an exception to explain why those runtime checks cannot be run.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -290,4 +319,20 @@ protected function get_check_slugs() {
 
 		return $this->get_check_slugs_param();
 	}
+
+	/**
+	 * Returns the plugin basename.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string
```
felixarntz: Nit-pick, but please add a return description for what this is.

includes/Plugin_Main.php
```diff
@@ -54,12 +54,11 @@ public function context() {
 	 * @since n.e.x.t
 	 */
 	public function add_hooks() {
+		// Setup the CLI command.
+		$context = $this->context();
+		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'cli.php';
```
felixarntz: I know it's strictly not necessary since the file checks it too, but I think it's a good practice to also include the `WP_CLI` constant check here. We don't need to load the file if we're not in a CLI command execution.

felixarntz (commented):
@jjgrainger Mostly looks good including the refactoring, however a few things need to be fixed.

cli.php
jjgrainger: Thanks @felixarntz updated

cli.php
jjgrainger: Updated

cli.php
jjgrainger: Thanks @felixarntz 

I've explored using some of the other hooks however, from what I have seen, any that include the command name in the hook run too late for adding this file. I've gone with the approach of manually checking the command before adding the file.

cli.php
jjgrainger: Removed throughout, thanks

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks, updated

includes/CLI/Plugin_Check_Command.php
jjgrainger: @felixarntz this wasn't deliberate and it doesn't matter which order these are set. But, this change does match the same order in the admin AJAX now.

includes/CLI/Plugin_Check_Command.php
jjgrainger: Updated

includes/CLI/Plugin_Check_Command.php
jjgrainger: Updated

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks, updated

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @felixarntz That makes sense to me as it validates the plugin parameter early. I've made this change, but to note this did require making the `AJAX_Runner::get_plugin_basename()` public.

includes/Traits/URL_Aware.php
jjgrainger: I was experiencing errors when the `_cleanup_query_vars()` function call was used when testing manually. From what I can see the `_cleanup_query_vars()` is a test utility class which explains why this was working in unit tests. The contents here matches that of the function (and poc).

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I noticed this too. I've updated this method with the additional check.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Updated

includes/Plugin_Main.php
jjgrainger: Thanks @felixarntz, I agree and have updated.

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz It seems that the array in the `FormData` passed in the request was being converted into a comma separated string, which when running manual tests were breaking here which lead to this update.

I agree, and think there is a way to ensure an array is passed with the `FormData`. For now I have reverted this change and I can do the same and make the necessary updates to get this working (as I've also seen the `AJAX_Runner` uses `wp_parse_list` too).

cli.php
felixarntz: @jjgrainger Sounds good, thanks!

includes/Admin/Admin_AJAX.php
felixarntz: That's a fair point. But yeah, since it's definitely possible to pass query or body parameters as an array (via `param[]=...` notation), I would imagine this can be done via `FormData` as well.

includes/Admin/Admin_AJAX.php
felixarntz: @jjgrainger Makes sense, thanks!

includes/Traits/URL_Aware.php
felixarntz: @jjgrainger Good catch, thanks!

felixarntz (approved):
Thanks @jjgrainger, LGTM!

cli.php
```diff
@@ -0,0 +1,58 @@
...
+			return;
+		}
+
+		if (
+			'wp' === substr( $_SERVER['argv'][0], -2 ) &&
+			'plugin' === $_SERVER['argv'][1] &&
+			'check' === $_SERVER['argv'][2]
+		) {
+			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+				copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' );
```
spacedmonkey:
```suggestion
				copy(  __DIR__ . '/object-cache.copy.php', WP_CONTENT_DIR . '/object-cache.php' );
```

spacedmonkey: I think we should have some level of error handling, if this file fails to copy. wp-content directory can be set to not writable.

mukeshpanchal27: Similar to Performance Lab can we define variable and use in `if` condition and `copy` command?

`$dropin_path = WP_CONTENT_DIR . '/object-cache.php';`

includes/CLI/Plugin_Check_Command.php
```diff
@@ -317,4 +347,22 @@ protected function display_results( $formatter, $file_name, $file_results ) {
 		WP_CLI::line();
 		WP_CLI::line();
 	}
+
+	/**
+	 * Check for a Runtime_Check in a list of checks
```
```suggestion
	 * Checks for a Runtime_Check in a list of checks.
```

cli.php
```diff
@@ -0,0 +1,58 @@
...
+
+if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
+	return;
+}
+
+// Check if the plugin autoloading is set up.
+if ( ! class_exists( 'WordPress\Plugin_Check\CLI\Plugin_Check_Command' ) ) {
+	// Check the autoload file exists.
+	if ( ! file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
+		WP_CLI::error( 'Plugin Check autoloaded not found.' );
```
mukeshpanchal27:
```suggestion
		WP_CLI::error(
			__( 'Plugin Check autoloaded not found.', 'plugin-check' )
		);
```

mukeshpanchal27 (commented):
Thanks @jjgrainger, Left nit-pick.

cli.php
jjgrainger: Thanks @spacedmonkey @mukeshpanchal27 

This file is required early in the load process so the `WP_CONTENT_DIR` constant isn't available, so we can't use that here.

In terms of error handling, if this file is not copied we capture this later in the CLI command process and respond back with an error. For context, the object-cache.php file is used to initialize the `CLI_Runner` class early for runtime checks. If the file is not copied, the `CLI_Runner` is not initialized early and we cannot perform runtime checks so we return an error in the command.

I agree we can probably do more to harden this logic but I think we could tackle this next week as part of the milestone 1 review.

cli.php
```diff
@@ -0,0 +1,58 @@
...
+		return;
+	}
+
+	// Load the Composer autoloader.
+	require_once __DIR__ . '/vendor/autoload.php';
+}
+
+if ( ! isset( $context ) ) {
+	$context = new WordPress\Plugin_Check\Plugin_Context( __DIR__ . '/plugin-check.php' );
```
spacedmonkey: Question: is there a reason why these are not imported at the top of the file this?

```
use WordPress\Plugin_Check\Plugin_Context;
```

cli.php
spacedmonkey: Are you sure about this? Developers can and do define in `WP_CONTENT_DIR` in their wp-config file. [Consider](https://github.com/roots/bedrock/blob/09627b150fc323c5dd98d5e88f5d4789dd0ebf8c/config/application.php#L70) bedrock. Could we check to see if the const exists and fallback to `ABSPATH . 'wp-content`?

tests/Utilities/Plugin_Request_Utility_Tests.php
```diff
@@ -85,6 +85,9 @@ function( $checks ) {
 			}
 		);
 
+		$muplugins_loaded = $wp_actions['muplugins_loaded'];
+		unset( $wp_actions['muplugins_loaded'] );
```
spacedmonkey, line 89: This feels like it should be in the `set_up` method.

tests/Utilities/Plugin_Request_Utility_Tests.php
```diff
@@ -97,6 +100,7 @@ function( $checks ) {
 		$runner  = Plugin_Request_Utility::get_runner();
 
 		unset( $_SERVER['argv'] );
+		$wp_actions['muplugins_loaded'] = $muplugins_loaded;
```
spacedmonkey, line 103: This feels like this should in the `tear_down` method

includes/Checker/CLI_Runner.php
```diff
@@ -37,7 +37,7 @@ public function is_plugin_check() {
 		}
 
 		if (
-			'wp' === $_SERVER['argv'][0] &&
+			'wp' === substr( $_SERVER['argv'][0], -2 ) &&
 			'plugin' === $_SERVER['argv'][1] &&
 			'check' === $_SERVER['argv'][2]
 		) {
```
spacedmonkey, line 43: This code seems to be repeated. It worth moving this to a util.

cli.php
spacedmonkey:
```suggestion
				if( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
				      WP_CLI::error(
			              __( 'Unable to copy object-cache.php file.', 'plugin-check' )
		              );
				}
```

Could we do something simple like this?

includes/Admin/Admin_AJAX.php
```diff
@@ -62,55 +63,36 @@ public function get_checks_to_run() {
...
+
+		if ( is_null( $runner ) ) {
+			$runner = new AJAX_Runner();
+		}
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
 			wp_send_json_error(
-				new WP_Error( 'invalid-plugin', $error->getMessage() ),
+				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ) ),
```
spacedmonkey, line 74:
```suggestion
				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ), array( 'status' => 403 ) ),
```

Nitpic, you can pass status to WP_Error object.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -243,16 +251,46 @@ private function get_shared_preparations( array $checks ) {
...
+		} else {
+			// Run all checks for the plugin.
+			$checks_to_run = $all_checks;
+
+			// Only run static checks if the plugin is inactive or the runtime environment was not set up.
+			if ( ! $plugin_active || ! $this->initialized_early ) {
+				$checks_to_run = array_filter(
+					$checks_to_run,
+					function ( $check ) {
+						return ! $check instanceof Runtime_Check;
```
spacedmonkey, line 287: This line is repeated
https://github.com/10up/plugin-check/pull/100/files#diff-ad151f5d51d23b5a7893d014606fa6d982f5efce1a005d4070a44957badfb35eR207

A little util method could be used here.

tests/Checker/AJAX_Runner_Tests.php
```diff
@@ -61,9 +63,14 @@ function( $checks ) {
 			}
 		);
 
+		$muplugins_loaded = $wp_actions['muplugins_loaded'];
+		unset( $wp_actions['muplugins_loaded'] );
```
spacedmonkey, line 67: As this same code is repeated, maybe 
1. Create a trait, to reuse code. 
2. Move set_up / tear_down.

includes/Admin/Admin_AJAX.php
spacedmonkey: Same 
https://github.com/10up/plugin-check/pull/100/files#diff-9664c81d5e62051fb173b90454641ac9013c9397129f7e6adbb75f97978cad20R88

spacedmonkey:
I think this is close, I have some nicpics, otherwise this is looking good.

cli.php
felixarntz: @jjgrainger @mukeshpanchal27 Please also fix this to say "Plugin Check autoloader ..." instead of "Plugin Check autoloaded ..."

tests/Utilities/Plugin_Request_Utility_Tests.php
felixarntz: @spacedmonkey I think this is specific to this test and needs to be here to force the code to think that the action wasn't run yet. So I think this is correct.

tests/Utilities/Plugin_Request_Utility_Tests.php
felixarntz: See my reply in https://github.com/10up/plugin-check/pull/100#discussion_r1145234889.

cli.php
jjgrainger: Thanks @spacedmonkey 

For context, we use this file with the command like so `wp plugin check hello.php --require=./wp-content/plugins/plugin-check/cli.php` so this file is run before any of WordPress has been loaded. We're using a specific [WP CLI hook](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-add-hook/) `before_wp_load` to run this logic ahead of the WP load process to ensure the runtime environment is setup.

I've tested using the `WP_CONTENT_DIR` and translation functions and both times these are not available.

The same would be true for bedrock as it `requires_once` it's [own config files](https://github.com/roots/bedrock/blob/09627b150fc323c5dd98d5e88f5d4789dd0ebf8c/web/wp-config.php#L8) as part of the `wp-config.php`. 

I agree there is more that can be done here, not only testing if the `WP_CONTENT_DIR` exists, but additional tests for whether the `/wp-content` directory exists as developers can change the folder name and location. As part of the milestone 1 review we're expecting to run this against a standard WordPress installation so we can be confident this will work for that. As for the rest I think we should tackle this as part of the architectural review next week where we have time to explore and discuss solutions.

Thanks @mukeshpanchal27 on your comment on creating a `$dropin_path = WP_CONTENT_DIR . '/object-cache.php';`, we're preferably trying to avoid declaring anything in the file and I think the `WP_CONTENT_DIR` comments above also apply. Let's review this as part of next weeks milestone.

cli.php
jjgrainger: @mukeshpanchal27 @felixarntz I will update to `autoloader` here, but see my other comment on how WordPress is not loaded yet so the translation functions are not available yet and can't be used. https://github.com/10up/plugin-check/pull/100#discussion_r1146027688

Let's review this as part of the milestone 1 review to see how and if we would want to resolve this.

tests/Utilities/Plugin_Request_Utility_Tests.php
jjgrainger: @spacedmonkey @felixarntz Felix's comment is correct, we're forcing the code to think it's initialized early. We also don't use this for all tests in this file so I don't think adding this to a `set_up` method would be the best solution here.

As we've progressed we've identified repeated code that we could refactor but I think it would be best to tackle this once the wider architecture is complete so we can tackle the refactoring as a whole.

includes/Checker/CLI_Runner.php
jjgrainger: Thanks @spacedmonkey Yes, we've noticed this too and other elements of the codebase as well that is repeated. See my other comment https://github.com/10up/plugin-check/pull/100#discussion_r1146032399 I think it's best to tackle some of this refactoring as part of next weeks review once the wider architecture is complete.

cli.php
jjgrainger: @spacedmonkey @felixarntz I've made this update, however I've omitted the use of the translsation function as it's not available yet - https://github.com/10up/plugin-check/pull/100#discussion_r1146027688

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @spacedmonkey as you've identified we've been consistent with the approach here so this change would require refactoring across multiple instances. With the current goal being to complete the architecture ahead of the milestone 1 review that starts next week, I would rather discuss and address changes like this then.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: @spacedmonkey I agree, but I think we should tackle this once the wider architecture is complete so we can see where/how this is used across the codebase before refactoring.

tests/Checker/AJAX_Runner_Tests.php
jjgrainger: Thanks @spacedmonkey I agree, but lets tackle this type of refactoring as part of the milestone 1 review once the architecture is complete.

includes/Checker/Abstract_Check_Runner.php
spacedmonkey: Okay, please leave a comment, so we can revise this later.

includes/Checker/CLI_Runner.php
spacedmonkey: We should fix this later, add a comment / todo in the code to fix.

tests/Checker/AJAX_Runner_Tests.php
spacedmonkey: This one, I don't agree with. I don't like this how it is. Adding a trait is simple enough, let's just do this one now.

tests/Utilities/Plugin_Request_Utility_Tests.php
spacedmonkey: Okay, this is repeated code, it should be part of a trait.

includes/Admin/Admin_AJAX.php
spacedmonkey: Anything you are not planning to fix in this PR, then add a comment in the code.

includes/CLI/Plugin_Check_Command.php
spacedmonkey: ?

cli.php
spacedmonkey: Is it worth doing this. 

```
$content_dir = defined( 'WP_CONTENT_DIR' ) ? WP_CONTENT_DIR : ABSPATH . 'wp-content';
```

That way you could define it at the start of the file yourself and get around these issues?

tests/Utilities/Plugin_Request_Utility_Tests.php
jjgrainger: @spacedmonkey we've agreed that we're planning to address refactoring elements like this as part of the milestone review so that we don't refactor in isolation and  the wider team can contribute as part of this. As this is non-blocking I think this can wait until then.

includes/Checker/CLI_Runner.php
jjgrainger: @spacedmonkey we have not done this with other areas we would like to refactor and I don't think adding comments now will add any value. We'll be reviewing items like this as part of the wider review, for now we want to be reviewing and addressing anything critical.

cli.php
jjgrainger: @spacedmonkey there is more work that needs to be done here as part of the refactoring regarding developers being able to change the folder name and location. We'd need more time to discuss and decide on a best path forward as part of the milestone 1 review. I don't think making your suggested change helps to solve anything here

cli.php
felixarntz: @jjgrainger Ah that makes sense. In that case, I agree, let's not use `__()`.

includes/Admin/Admin_AJAX.php
felixarntz: @spacedmonkey This one I'm not sure about, what would be the benefit of passing the status code to `WP_Error`? I know the REST API uses that, but this is a simple JSON response in an AJAX request, so I don't think this would do anything?

spacedmonkey: This is a nicpic. Ignore me.

tests/Checker/AJAX_Runner_Tests.php
spacedmonkey: Break out issue. https://github.com/10up/plugin-check/issues/109

tests/Utilities/Plugin_Request_Utility_Tests.php
spacedmonkey: We would handle this [here](https://github.com/10up/plugin-check/issues/109).

cli.php
```diff
@@ -0,0 +1,63 @@
...
+	'before_wp_load',
+	function() {
+		if ( empty( $_SERVER['argv'] ) || 3 > count( $_SERVER['argv'] ) ) {
+			return;
+		}
+
+		if (
+			'wp' === substr( $_SERVER['argv'][0], -2 ) &&
+			'plugin' === $_SERVER['argv'][1] &&
+			'check' === $_SERVER['argv'][2]
```
spacedmonkey, line 54: See https://github.com/10up/plugin-check/issues/111

includes/Checker/CLI_Runner.php
spacedmonkey: See https://github.com/10up/plugin-check/issues/111

includes/CLI/Plugin_Check_Command.php
spacedmonkey: See https://github.com/10up/plugin-check/issues/112

spacedmonkey (approved):
Created a number of follow on issues. But this is good to merge now.

felixarntz:
Thanks @spacedmonkey!

PR #101, opened 2023-03-17
Create AJAX runtime environment setup/cleanup

### LINKED ISSUE #86: AJAX Runtime Environment Setup/Cleanup
This issue will implement the runtime environment setup and cleanup for AJAX requests. This will utilise the `Runtime_Environment_Setup` class to setup the WP environment before checks are run, and cleanup the environment once all checks are complete.

- #90 

### Create a Runtime Setup AJAX endpoint
- In the `Admin_Ajax` should create a new AJAX endpoint for the action `plugin_check_setup_environment`
- A new action is created for `wp_ajax_plugin_check_setup_environment` that calls a `setup_environment` method on the `Admin_Ajax` class.
- A public `has_runtime_check` method exists in the `Abstract_Check_Runner` class. This method will test the checks returned by `get_checks_to_run()`. If there is is at least 1 `Runtime_Check` it will return `true`, otherwise `false`.
- The `setup_environment` method.
    - This method will get an instance of the `AJAX_Runner` class via the `Plugin_Request_Utility::get_runner()` method.
        - If the `Plugin_Request_Utility::get_runner()` returns `null` a new instance of the `AJAX_Runner` class should be created.
    - Run the `set_check_slugs()` and `set_plugin_slug ()` methods  on the `AJAX_Runner` instance passing in the the appropriate values taken from the request. 
    - The method will test if any of the checks requested are Runtime checks by calling the `AJAX_Runner` `has_runtime_check()` method. If this returns `true` the `Runtime_Environment_Setup` class will be instantiated and it's `setup()` method called.
    - The method will then return a JSON response for the AJAX request with the following structure `{ "success": true|false, "message": (string) success/error message }`
        - An error response will be returned if setup had a critical error passing `success` as `false` and the message of the error that occured.
        - If the setup was successful, a success response will be returned with `success` as `true` and the message `"runtime environment setup successful"`
        - If the setup did not need to run, a success response will be returned with `success` as `true` and the message `"no runtime checks, runtime environment was not setup"`
- The `plugin-check-admin.js` file will be updated to first send an AJAX request to the `admin-ajax.php` endpoint with the `plugin_check_setup_environment` action before any check requests are sent.
- The request will also pass the plugin and checks to be run to the endpoint in order to determine if the runtime environment needs to be setup.
- If there is an error response, a user friendly error message will be displayed on the admin page and no further AJAX requests will be sent to attempt to run checks,

### Create a Runtime Cleanup AJAX endpoint
- In the `Admin_Ajax` should create a new AJAX endpoint for the action `plugin_check_cleanup_environment`
- A new action is created for `wp_ajax_plugin_check_cleanup_environment` that calls a `cleanup_environment` method on the `Admin_Ajax` class.
- The `cleanup_environment` method:
    - This method will test if the runtime environment had been prepared by checking if the any of the tables with the `wppc_` prefix exist.
    - If the `wppc_` tables exist, the `Runtime_Environment_Setup` class will be instantiated and it's `cleanup()` method called.
    - The method will then return a JSON response for the AJAX request with the following structure `{ "success": true|false, "message": (string) success/error message }`
        - An error response will be returned if cleanup had a critical error passing `success` as `false` and the message of the error that occured.
        - If the setup was successful, a success response will be returned with `success` as `true` and the message `"runtime environment cleanup successful"`
        - If the setup did not need to run, a success response will be returned with `success` as `true` and the message `"Runtime environment was not prepared, cleanup was not run."`
- The `plugin-check-admin.js` file will be updated to send a final AJAX request to the `admin-ajax.php` endpoint with the `plugin_check_cleanup_environment` action after all checks have been run.
- The request will also pass the plugin and checks to be run to the endpoint in order to determine if the runtime environment needs to be cleaned up.

jjgrainger:
- Addresses acceptance criteria for #86 

Closes #86

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -12,9 +12,9 @@
 		e.preventDefault();
 
 		getChecksToRun()
-		.then( setupEnvironment )
+		.then( setUpEnvironment )
 		.then( runChecks )
-		.then( cleanupEnvironment )
+		.then( cleanUpEnvironment )
 		.then(
 			( data ) => {
 				console.log( data.message );
@@ -32,7 +32,7 @@
 	 *
 	 * @since n.e.x.t
 	 */
-	function setupEnvironment( data ) {
+	function setUpEnvironment( data ) {
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', data.plugin );
@@ -74,7 +74,7 @@
 	 *
 	 * @since n.e.x.t
 	 */
-	function cleanupEnvironment( data ) {
+	function cleanUpEnvironment( data ) {
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'action', 'plugin_check_clean_up_environment' );
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -39,7 +39,7 @@
 		pluginCheckData.append( 'action', 'plugin_check_set_up_environment' );
 
 		for (var i = 0; i < data.checks.length; i++) {
-			pluginCheckData.append('checks[]', data.checks[i]);
+			pluginCheckData.append( 'checks[]', data.checks[ i ] );
 		}
 
 		return fetch(
@@ -116,7 +116,6 @@
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', pluginsList.value );
-		pluginCheckData.append( 'checks[]', [] );
 		pluginCheckData.append( 'action', 'plugin_check_get_checks_to_run' );
 
 		return fetch(
@@ -157,7 +156,7 @@
 		pluginCheckData.append( 'action', 'plugin_check_run_checks' );
 
 		for (var i = 0; i < data.checks.length; i++) {
-			pluginCheckData.append('checks[]', data.checks[i]);
+			pluginCheckData.append( 'checks[]', data.checks[ i ] );
 		}
 
 		return fetch(
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -11,7 +11,8 @@
 	checkItButton.addEventListener( 'click', ( e ) => {
 		e.preventDefault();
 
-		getChecksToRun()
+		setupEnvironment()
+		.then( getChecksToRun )
 		.then( runChecks )
 		.then(
 			( data ) => {
@@ -24,6 +25,46 @@
 			}
 		);
 	} );
+
+	/**
+	 * Setup the runtime environment if needed.
+	 *
+	 * @since n.e.x.t
+	 */
+	function setupEnvironment() {
+		// Setup the form data to post.
+		const pluginCheckData = new FormData();
+		pluginCheckData.append( 'nonce', pluginCheck.nonce );
+		pluginCheckData.append( 'plugin', pluginsList.value );
+		pluginCheckData.append( 'checks', [] );
+		pluginCheckData.append( 'action', 'plugin_check_setup_environment' );
+
+		return fetch(
+			ajaxurl,
+			{
+				method: 'POST',
+				credentials: 'same-origin',
+				body: pluginCheckData
+			}
+		)
+		.then(
+			( response ) => {
+				return response.json();
+			}
+		)
+		.then( handleDataErrors )
+		.then(
+			( data ) => {
+				if ( ! data.data || ! data.data.message ) {
+					throw new Error( 'Response contains no data.' );
+				}
+
+				console.log( data.data.message );
+
+				return data.data;
+			}
+		);
+	}
 
 	/**
 	 * Get the Checks to run.
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -11,9 +11,10 @@
 	checkItButton.addEventListener( 'click', ( e ) => {
 		e.preventDefault();
 
-		setupEnvironment()
-		.then( getChecksToRun )
+		getChecksToRun()
+		.then( setupEnvironment )
 		.then( runChecks )
+		.then( cleanupEnvironment )
 		.then(
 			( data ) => {
 				console.log( data.message );
@@ -31,12 +32,12 @@
 	 *
 	 * @since n.e.x.t
 	 */
-	function setupEnvironment() {
+	function setupEnvironment( data ) {
 		// Setup the form data to post.
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
-		pluginCheckData.append( 'plugin', pluginsList.value );
-		pluginCheckData.append( 'checks', [] );
+		pluginCheckData.append( 'plugin', data.plugin );
+		pluginCheckData.append( 'checks', data.checks );
 		pluginCheckData.append( 'action', 'plugin_check_setup_environment' );
 
 		return fetch(
@@ -65,6 +66,44 @@
 			}
 		);
 	}
+
+	/**
+	 * Cleanup the runtime environment.
+	 *
+	 * @since n.e.x.t
+	 */
+		function cleanupEnvironment( data ) {
+			const pluginCheckData = new FormData();
+			pluginCheckData.append( 'nonce', pluginCheck.nonce );
+			pluginCheckData.append( 'action', 'plugin_check_cleanup_environment' );
+
+			return fetch(
+				ajaxurl,
+				{
+					method: 'POST',
+					credentials: 'same-origin',
+					body: pluginCheckData
+				}
+			)
+			.then(
+				( response ) => {
+					return response.json();
+				}
+			)
+			.then( handleDataErrors )
+			.then(
+				( data ) => {
+					if ( ! data.data || ! data.data.message ) {
+						throw new Error( 'Response contains no data.' );
+					}
+
+					console.log( data.data.message );
+
+					return data.data;
+				}
+			);
+		}
+
 
 	/**
 	 * Get the Checks to run.
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -72,37 +72,37 @@
 	 *
 	 * @since n.e.x.t
 	 */
-		function cleanupEnvironment( data ) {
-			const pluginCheckData = new FormData();
-			pluginCheckData.append( 'nonce', pluginCheck.nonce );
-			pluginCheckData.append( 'action', 'plugin_check_cleanup_environment' );
+	function cleanupEnvironment( data ) {
+		const pluginCheckData = new FormData();
+		pluginCheckData.append( 'nonce', pluginCheck.nonce );
+		pluginCheckData.append( 'action', 'plugin_check_cleanup_environment' );
 
-			return fetch(
-				ajaxurl,
-				{
-					method: 'POST',
-					credentials: 'same-origin',
-					body: pluginCheckData
+		return fetch(
+			ajaxurl,
+			{
+				method: 'POST',
+				credentials: 'same-origin',
+				body: pluginCheckData
+			}
+		)
+		.then(
+			( response ) => {
+				return response.json();
+			}
+		)
+		.then( handleDataErrors )
+		.then(
+			( data ) => {
+				if ( ! data.data || ! data.data.message ) {
+					throw new Error( 'Response contains no data.' );
 				}
-			)
-			.then(
-				( response ) => {
-					return response.json();
-				}
-			)
-			.then( handleDataErrors )
-			.then(
-				( data ) => {
-					if ( ! data.data || ! data.data.message ) {
-						throw new Error( 'Response contains no data.' );
-					}
 
-					console.log( data.data.message );
+				console.log( data.data.message );
 
-					return data.data;
-				}
-			);
-		}
+				return data.data;
+			}
+		);
+	}
 
 
 	/**
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -38,7 +38,7 @@
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', data.plugin );
 		pluginCheckData.append( 'checks', data.checks );
-		pluginCheckData.append( 'action', 'plugin_check_setup_environment' );
+		pluginCheckData.append( 'action', 'plugin_check_set_up_environment' );
 
 		return fetch(
 			ajaxurl,
@@ -75,7 +75,7 @@
 	function cleanupEnvironment( data ) {
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
-		pluginCheckData.append( 'action', 'plugin_check_cleanup_environment' );
+		pluginCheckData.append( 'action', 'plugin_check_clean_up_environment' );
 
 		return fetch(
 			ajaxurl,
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -33,12 +33,14 @@
 	 * @since n.e.x.t
 	 */
 	function setupEnvironment( data ) {
-		// Setup the form data to post.
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', data.plugin );
-		pluginCheckData.append( 'checks', data.checks );
 		pluginCheckData.append( 'action', 'plugin_check_set_up_environment' );
+
+		for (var i = 0; i < data.checks.length; i++) {
+			pluginCheckData.append('checks[]', data.checks[i]);
+		}
 
 		return fetch(
 			ajaxurl,
@@ -111,11 +113,10 @@
 	 * @since n.e.x.t
 	 */
 	function getChecksToRun() {
-		// Collect the data to pass along for generating a check results.
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', pluginsList.value );
-		pluginCheckData.append( 'checks', [] );
+		pluginCheckData.append( 'checks[]', [] );
 		pluginCheckData.append( 'action', 'plugin_check_get_checks_to_run' );
 
 		return fetch(
@@ -153,8 +154,11 @@
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', data.plugin );
-		pluginCheckData.append( 'checks', data.checks );
 		pluginCheckData.append( 'action', 'plugin_check_run_checks' );
+
+		for (var i = 0; i < data.checks.length; i++) {
+			pluginCheckData.append('checks[]', data.checks[i]);
+		}
 
 		return fetch(
 			ajaxurl,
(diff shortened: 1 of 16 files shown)


jjgrainger:
Hi @felixarntz 

After setting this up I've been facing errors where the `wp_get_cache` function do not exist. From what I can tell this is due to setting up and preparing the Runners in the `object-cache.php` file, in particular when getting the plugin parameter and using the `Plugin_Request_Utility::get_plugin_basename_from_input` which calls `get_plugins()` that uses `wp_get_cache`.

The reason these functions are called is because we need to create the `Check_Context` for the `Checks` class in order to return the array of checks to then work out if there is a runtime check and setup the environment.

I'm wondering what the best path forward is here? Whether we refactor the `Checks` class so we can return the array of Checks without needing a `Check_Context` (though I think we would still need to test if the plugin is active based on the logic we are trying to run) or if we simply could `require_once wp-includes/cache.php` before the `get_plugins` call is made?

Happy to discuss this over a call as I feel this is somewhat related to my other comment - https://github.com/10up/plugin-check/pull/100#issuecomment-1472466634

Thanks

felixarntz:
@jjgrainger Good catch!

Maybe we can get around this without any notable refactoring, just by executing that logic later? While we have to call the code in `object-cache.php`, it doesn't necessarily have to be executed right away.

Potentially, we can hook the code into the `muplugins_loaded` action? That action should still be in time for us to do the other things, like filter out plugins etc. If there is an earlier action that we can reasonably use, that would be even better, but I'm not sure there is.

jjgrainger:
Thanks @felixarntz 

I've updated the PR to use the `muplugins_loaded` action to initialize the runner and this is ready for review.

After updating this I came across a different error where the `WP_PLUGIN_CHECK_MAIN_FILE` and `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH` constants were not defined in the `Universal_Runtime_Preparation`.

I've put in a small fix to get around this but I'm not sure this is the best approach but we can discuss as part of the code review.

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
 		return wp_create_nonce( self::NONCE_KEY );
 	}
 
+	/**
+	 * Handles the AJAX request to setup the runtime environment if needed.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function setup_environment() {
```
felixarntz: Nit-pick: "Setup" is a noun, we should use the verb "set up". This also applies to the AJAX action etc.
```suggestion
	public function set_up_environment() {
```

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
...
+			)
+		);
+	}
+
+	/**
+	 * Handles the AJAX request to cleanup the runtime environment.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function cleanup_environment() {
```
felixarntz: Same here: "Cleanup" is a noun, but for method names we should use verbs, i.e. "clean up".
```suggestion
	public function clean_up_environment() {
```

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
...
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
+			wp_send_json_error(
+				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ) ),
+				500
+			);
+		}
+
+		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
```
felixarntz: See my comment in https://github.com/10up/plugin-check/pull/100#discussion_r1142610721, why is this not an array of strings?

includes/Admin/Admin_AJAX.php
```diff
@@ -59,7 +157,7 @@ public function get_checks_to_run() {
 			wp_send_json_error( $valid_nonce, 403 );
 		}
 
-		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
+		$checks = wp_parse_list( filter_input( INPUT_POST, 'checks', FILTER_SANITIZE_STRING ) );
```
felixarntz: See above, I don't think we should change this.

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
...
+			$runtime = new Runtime_Environment_Setup();
+			$runtime->setup();
+			$message = __( 'Runtime environment setup successful.', 'plugin-check' );
+		}
+
+		wp_send_json_success(
+			array(
+				'message' => $message,
+				'plugin'  => $plugin,
+				'checks'  => $checks,
```
felixarntz, line 105: I see you're returning all these values from every AJAX endpoint just to be able to chain the AJAX requests in the JS file. I think that makes for a confusing architecture and requires us to include unnecessary data in the AJAX responses. No need to fix that here since it's already happening in existing code, but I think that will be something we should resolve as part of the architecture review.

assets/js/plugin-check-admin.js
```diff
@@ -12,7 +12,9 @@
 		e.preventDefault();
 
 		getChecksToRun()
+		.then( setupEnvironment )
 		.then( runChecks )
+		.then( cleanupEnvironment )
```
felixarntz: See my comment below, rather than chaining these and thereby forcing every AJAX response to include its request parameters to pass it on to the next AJAX request, I think we should decouple them.

We can then change the JS functions to accept dedicated parameters for the things that each function needs rather than the generic `data` object from the previous AJAX response.

But like my comment above, let's handle this refactoring in a separate PR.

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
...
+			wp_send_json_error( $valid_nonce, 403 );
+		}
+
+		// Set the new prefix.
+		$old_prefix = $wpdb->set_prefix( 'wppc_' );
+
+		$message = __( 'Runtime environment was not prepared, cleanup was not run.', 'plugin-check' );
+
+		// Test if the runtime environment tables exist.
+		if ( 'wppc_posts' === $wpdb->get_var( "SHOW TABLES LIKE 'wppc_posts'" ) ) {
```
felixarntz: Maybe we should also check for the constant that our `object-cache.php` drop-in sets. Just to make sure that we also run the clean up if e.g. the `object-cache.php` drop-in was placed but for some reason the database tables were not created successfully.
```suggestion
		if ( 'wppc_posts' === $wpdb->get_var( "SHOW TABLES LIKE 'wppc_posts'" ) || defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) ) {
```

felixarntz (commented):
@jjgrainger Left a few comments, some of them similar to what I mentioned on #100.

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks, updated

includes/Admin/Admin_AJAX.php
jjgrainger: Updated

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz I responded on https://github.com/10up/plugin-check/pull/100#discussion_r1143636914 but to clarify I have found a solution to ensuring an array is passed in the JS `FormData`. This required appending each checks to the array so it is passed to the AJAX endpoint as expected.

I've also updated the `AJAX_Runner` and unit tests which were not setup correctly.

includes/Admin/Admin_AJAX.php
jjgrainger: Updated

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz I agree and have updated

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @felixarntz I agree and let's tackle this as part of the architectural review.

assets/js/plugin-check-admin.js
```diff
@@ -25,17 +27,96 @@
 		);
 	} );
 
+	/**
+	 * Setup the runtime environment if needed.
+	 *
+	 * @since n.e.x.t
+	 */
+	function setupEnvironment( data ) {
```
felixarntz: Nit-pick, but per my previous comment, this will also need to be changed accordingly:
```suggestion
	function setUpEnvironment( data ) {
```

assets/js/plugin-check-admin.js
```diff
@@ -25,17 +27,96 @@
...
+			}
+		);
+	}
+
+	/**
+	 * Cleanup the runtime environment.
+	 *
+	 * @since n.e.x.t
+	 */
+	function cleanupEnvironment( data ) {
```
felixarntz:
```suggestion
	function cleanUpEnvironment( data ) {
```

assets/js/plugin-check-admin.js
```diff
@@ -25,17 +27,96 @@
...
 	 *
 	 * @since n.e.x.t
 	 */
 	function getChecksToRun() {
-		// Collect the data to pass along for generating a check results.
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', pluginsList.value );
-		pluginCheckData.append( 'checks', [] );
+		pluginCheckData.append( 'checks[]', [] );
```
felixarntz: Doesn't this parameter need to not be set at all for it to be interpreted as an empty array? If that causes an error, we may need to adjust the server-side logic to ensure that the `checks` parameter is not _required_ (i.e. can be an empty array).

But setting this here to an empty array technically means this is an array with one element that is itself an array. So that wouldn't be right.

assets/js/plugin-check-admin.js
```diff
@@ -73,9 +154,12 @@
 		const pluginCheckData = new FormData();
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', data.plugin );
-		pluginCheckData.append( 'checks', data.checks );
 		pluginCheckData.append( 'action', 'plugin_check_run_checks' );
 
+		for (var i = 0; i < data.checks.length; i++) {
+			pluginCheckData.append('checks[]', data.checks[i]);
+		}
```
felixarntz: Please pay attention to correct spacing per WP coding standards. We may want to set us JS linting via ESLint for this repo (maybe open an issue for that?).
```suggestion
		for ( var i = 0; i < data.checks.length; i++ ) {
			pluginCheckData.append( 'checks[]', data.checks[ i ] );
		}
```

felixarntz (commented):
@jjgrainger Almost good to go, just one thing that is off in the `FormData` set, and a few nit-picks.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz updated and created a new issue here https://github.com/10up/plugin-check/issues/102

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz I've made changes here so that the `checks` parameter isn't required.

includes/Admin/Admin_AJAX.php
```diff
@@ -76,7 +76,7 @@ public function set_up_environment() {
 			);
 		}
 
-		$checks = array_filter( filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY ) );
+		$checks = isset( $_REQUEST['checks'] ) ? array_filter( $_REQUEST['checks'] ) : array();
```
felixarntz: I think this change also needs to be made to the `AJAX_Runner`?

On another note, is there no way to use `filter_input()` for this? Does `filter_input()` fail if the value is not provided? Maybe there's a flag?

felixarntz (commented):
@jjgrainger One last question :)

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz This change matches what we have for checks in the `AJAX_Runner` as part of this PR.

Looking into this further we could use the `FILTER_FORCE_ARRAY` flag to force the parameter to an array. I can revert this back to using `filter_input` and update both the `get_check_slugs_param` and `get_plugin_param` to also use the `filter_input`.

felixarntz: @jjgrainger Apologies, I must have missed the change in `AJAX_Runner` 🤦 

Would be great if you could update the code to use `filter_input()` again (with `FILTER_FORCE_ARRAY` like you're saying, assuming that also works if nothing is provided for the parameter). Not a blocker, so I am already approving this. Thanks!

felixarntz (approved):
Thanks @jjgrainger!

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @felixarntz this works and I have made the update

assets/js/plugin-check-admin.js
```diff
@@ -25,17 +27,95 @@
...
+			( data ) => {
+				if ( ! data.data || ! data.data.message ) {
+					throw new Error( 'Response contains no data.' );
+				}
+
+				console.log( data.data.message );
+
+				return data.data;
+			}
+		);
```
spacedmonkey, line 69: This code is repeated. Make it into a function or re-usage somehow.

assets/js/plugin-check-admin.js
```diff
@@ -25,17 +27,95 @@
...
+
+		return fetch(
+			ajaxurl,
+			{
+				method: 'POST',
+				credentials: 'same-origin',
+				body: pluginCheckData
+			}
+		)
+		.then(
```
spacedmonkey, line 90: This is a lot of thens. We can use async / await calls if possible, it would make this a lot more readable.

includes/Checker/AJAX_Runner.php
```diff
@@ -70,13 +70,6 @@ protected function get_plugin_param() {
...
 	protected function get_check_slugs_param() {
-		$checks = array();
-
-		if ( isset( $_REQUEST['checks'] ) ) {
-			// Checks are passed as a comma separated string.
-			$checks = wp_parse_list( $_REQUEST['checks'] );
-		}
-
-		return $checks;
+		return filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
```
spacedmonkey, line 73: Can you explain this change?

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -82,7 +82,14 @@ public function prepare() {
 	public function filter_active_plugins( $active_plugins ) {
 		if ( is_array( $active_plugins ) && in_array( $this->plugin_basename, $active_plugins, true ) ) {
 
-			$plugin_base_file = plugin_basename( WP_PLUGIN_CHECK_MAIN_FILE );
+			if ( defined( 'WP_PLUGIN_CHECK_MAIN_FILE' ) ) {
+				$plugin_check_file = WP_PLUGIN_CHECK_MAIN_FILE;
+			} else {
+				$plugins_dir       = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : WP_CONTENT_DIR . '/plugins';
```
spacedmonkey, line 88: Why would `WP_PLUGIN_DIR` not be defined?

includes/Checker/Preparations/Universal_Runtime_Preparation.php
```diff
@@ -54,7 +54,14 @@ public function prepare() {
 
 		$cleanup_functions = array();
 
-		$use_minimal_theme_preparation = new Use_Minimal_Theme_Preparation( 'wp-empty-theme', WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/test-content/themes' );
+		if ( ! defined( 'WP_PLUGIN_CHECK_PLUGIN_DIR_PATH' ) ) {
+			$plugins_dir  = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : WP_CONTENT_DIR . '/plugins';
+			$theme_folder = $plugins_dir . '/plugin-check/test-content/themes';
+		} else {
+			$theme_folder = WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'test-content/themes';
+		}
```
spacedmonkey, line 62: This logic is inverted compared to https://github.com/10up/plugin-check/pull/101/files#diff-35a6c4633cd89c9ad66535db741d064ff3f730cf69ea87d347d8e250544fff1eR85

assets/js/plugin-check-admin.js
jjgrainger: @spacedmonkey this is somewhat related to comments in @felixarntz earlier review (https://github.com/10up/plugin-check/pull/101#pullrequestreview-1349321441). There is additional work to be done as part of the remaining issues for milestone 1 so this will change.

assets/js/plugin-check-admin.js
jjgrainger: @spacedmonkey this is somewhat related to comments in @felixarntz earlier review (https://github.com/10up/plugin-check/pull/101#pullrequestreview-1349321441) and were we have agreed to address this in the future in a separate PR.

includes/Checker/AJAX_Runner.php
jjgrainger: @spacedmonkey this came from changes made within the `Admin_AJAX` file too. (see https://github.com/10up/plugin-check/pull/101#discussion_r1144957847). These need to match in order for the validation element in the `Abstract_Check_Runner` to work correctly.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
jjgrainger: @spacedmonkey the preparations are run when the `object-cache.php` file is loaded. This happens when the `WP_CONTENT_DIR` has been defined, but before the default `WP_PLUGIN_DIR` is, so it may not be set at this point in the load process.

includes/Admin/Admin_AJAX.php
```diff
@@ -46,6 +50,100 @@ public function get_nonce() {
...
 	}
 
+	/**
+	 * Handles the AJAX request to setup the runtime environment if needed.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function set_up_environment() {
+		// Verify the nonce before continuing.
+		$valid_nonce = $this->verify_nonce( filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_STRING ) );
```
joemcgill, line 60: Noting this for future reference, `FILTER_SANITIZE_STRING` has been deprecated in PHP 8.1 and will begin logging warnings, so we should replace this with a different alternative. See the changelog in: https://www.php.net/manual/en/filter.filters.sanitize.php

felixarntz: @joemcgill Can you open a new issue for addressing this throughout the codebase?

joemcgill: Created https://github.com/10up/plugin-check/issues/110.

assets/js/plugin-check-admin.js
spacedmonkey: Created issue. https://github.com/10up/plugin-check/issues/113

assets/js/plugin-check-admin.js
spacedmonkey: See https://github.com/10up/plugin-check/issues/113

spacedmonkey (approved):
LGTM

felixarntz:
@jjgrainger Now that #100 was merged, there's a merge conflict here, could you look into that please?

jjgrainger:
Thanks @felixarntz @spacedmonkey  I've addressed the conflicts and will merge now

PR #104, opened 2023-03-24
Set up JS linting for repo

spacedmonkey:
### Description of the Change

As documented in the WordPress handbook, wp-scripts, as lots of handy linters and checkers. [See](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#setup). Copy and pasted the format npm command in our project and run it. This is the result. Going forward developers can use this command to lint js/md/yml files. 

Closes #102

### Changelog Entry

> Added - New feature
> Changed - Existing functionality
> Deprecated - Soon-to-be removed feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @spacedmonkey

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #114, opened 2023-03-28
AJAX Requests to loop over checks to run

### LINKED ISSUE #84: Handle running checks in loop and return output

jjgrainger:
Addresses acceptance criteria in #84 

Closes #84


jjgrainger:
Set up as a draft so once #104 is merged I can update with the JS formatting here

assets/js/plugin-check-admin.js
```diff
@@ -149,15 +149,24 @@
...
 	 * @since n.e.x.t
 	 */
-	function runChecks( data ) {
+	async function runChecks( data ) {
+		const results = [];
+
+		for (var i = 0; i < data.checks.length; i++) {
+			let result = await runCheck( data.plugin, data.checks[ i ] );
+			results.push( result );
+		}
```
spacedmonkey: As a rule, it is a not good to do a series of promises in a loop like this. Do these NEED to run it order. What if one of the promises fail?

spacedmonkey: I would recommend using `Promise.all` here. 

Try something like this. 

```
const promises = [];
for (var i = 0; i < data.checks.length; i++) {
	promises.push(  runCheck( data.plugin, data.checks[ i ] ) );
}
const results = await Promise.all( promises );
```

includes/Admin/Admin_AJAX.php
```diff
@@ -248,4 +279,107 @@ protected function has_runtime_check( array $checks ) {
...
+			}
+
+			$file_results = $this->flatten_file_results( $file_name, $file_errors, $file_warnings );
+			$all_results  = array_merge( $all_results, $file_results );
+		}
+
+		// If there are any files left with only warnings, print those next.
+		foreach ( $warnings as $file_name => $file_warnings ) {
+			$file_results = $this->flatten_file_results( $file_name, array(), $file_warnings );
+			$all_results  = array_merge( $all_results, $file_results );
```
spacedmonkey: It can be expensive to call `array_merge`. Maybe just add `$file_results` to an array and then fatten it at the end.

includes/Admin/Admin_AJAX.php
```diff
@@ -248,4 +279,107 @@ protected function has_runtime_check( array $checks ) {
...
+				}
+				if ( $a['line'] > $b['line'] ) {
+					return 1;
+				}
+				if ( $a['column'] < $b['column'] ) {
+					return -1;
+				}
+				if ( $a['column'] > $b['column'] ) {
+					return 1;
+				}
```
spacedmonkey: These seems repeated, how about something like this. 

```
	if ( $a['line'] < $b['line'] || $a['column'] < $b['column'] ) {
					return -1;
	}
```

felixarntz: I don't think that would be the same logic, we have to _first_ consider both conditions around the `line` before we look at `column`.

includes/Admin/Admin_AJAX.php
```diff
@@ -208,9 +209,39 @@ public function run_checks() {
...
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
+			wp_send_json_error(
+				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ) ),
+				500
+			);
+		}
+
+		$checks = isset( $_REQUEST['checks'] ) ? array_filter( $_REQUEST['checks'] ) : array();
```
felixarntz: Why is that done differently here from above? I would think we should use the same logic everywhere for this parameter.
```suggestion
		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
		$checks = is_null( $checks ) ? array() : $checks;
```

assets/js/plugin-check-admin.js
felixarntz: Hmm, not sure about that one here. Normally I would agree with you, but I think it's a better user experience to show the check results in a consistent order, so circling through the checks makes sense to me in this context.

If we use `Promise.all`, they will be handled in a random order. That would only be a good idea if we wanted to output all results together at the end, but I think it's a good idea to output results one by one as it gives the user feedback already as soon as there are results.

assets/js/plugin-check-admin.js
```diff
@@ -149,15 +149,24 @@
...
+
+		console.log( results );
```
felixarntz: See my comment above, if you're doing this, then I would agree with @spacedmonkey's feedback. However, I would suggest we render the output from each check as soon as we have it. In other words, we should prepare for that here, i.e. `console.log` every individual result right away rather than putting them all in an array and then outputting them.

includes/Admin/Admin_AJAX.php
```diff
@@ -208,9 +209,39 @@ public function run_checks() {
...
+				new WP_Error( 'invalid-request', $error->getMessage() ),
+				400
+			);
+		}
+
 		wp_send_json_success(
 			array(
-				'message' => __( 'Verified!', 'plugin-check' ),
+				'results' => $this->format_results( $results ),
+				'checks'  => $checks,
```
felixarntz: Do we need to return `checks` from here? Why is that needed?

includes/Admin/Admin_AJAX.php
```diff
@@ -248,4 +279,107 @@ protected function has_runtime_check( array $checks ) {
...
+				unset( $warnings[ $file_name ] );
+			}
+
+			$file_results = $this->flatten_file_results( $file_name, $file_errors, $file_warnings );
+			$all_results  = array_merge( $all_results, $file_results );
+		}
+
+		// If there are any files left with only warnings, print those next.
+		foreach ( $warnings as $file_name => $file_warnings ) {
+			$file_results = $this->flatten_file_results( $file_name, array(), $file_warnings );
```
felixarntz: Why do we need to "flatten" the results? I don't see why that's needed. If it is helpful for the output, we should probably do that in the other issue in JS, but I'm not sure why we would do that.

I'm not even sure we need to do anything to the formatting here. Shouldn't this be the responsibility of the JS? The way this data is best formatted depends on the way it is being rendered, and that should be decoupled from the AJAX endpoint. I think we could probably avoid all this logic and instead simply return `errors` and `warnings` with the raw data in the AJAX response.

includes/Admin/Admin_AJAX.php
```diff
@@ -208,9 +209,39 @@ public function run_checks() {
...
+			wp_send_json_error(
+				new WP_Error( 'invalid-request', $error->getMessage() ),
+				400
+			);
+		}
+
 		wp_send_json_success(
 			array(
-				'message' => __( 'Verified!', 'plugin-check' ),
+				'results' => $this->format_results( $results ),
```
felixarntz: See my other comment: I don't think we need to format to cater for a specific opinionated output format, we can return the raw data here and then handle formatting as needed in JS, according to how we use the template.
```suggestion
				'errors    => $results->get_errors(),
				'warnings' => $results->get_warnings(),
```

felixarntz (commented):
Thanks @jjgrainger, mostly looks good. My main feedback is that I don't think we should do all the formatting in the AJAX endpoint as it seems unnecessary here and could happen in JS depending on the template usage.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @spacedmonkey @felixarntz 

Another point to raise here is that we would want these to run sequentially. We wouldn't want to have 2 runtime checks running in parallel, where both their preparations are implemented at the same time, as this may cause incorrect results returned by checks.

I have updated this to log the results as they're returned.

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @spacedmonkey I agree here, although the updates to handle the formatting on the JS side now has meant this code has been removed.

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz good catch! Not sure why this was reverted but have updated this to match.

assets/js/plugin-check-admin.js
jjgrainger: Updated to log the individual results right away.

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz this was part of the AC's but I agree that we should only return whats required in the response and we can always add things as and when we need it. I've removed the `checks` for now.

includes/Admin/Admin_AJAX.php
jjgrainger: @felixarntz this was part of the AC's to format the results in a particular way. However, I agree that we should decouple the logic here and return the raw `errors` and `warnings`. The formatting methods have been removed.

includes/Admin/Admin_AJAX.php
jjgrainger: Thanks @felixarntz updated.

jjgrainger:
Thanks @spacedmonkey & @felixarntz

I've made updates based on the feedback and is ready for you to review.

package-lock.json
```diff
@@ -12237,4 +12237,4 @@
       "dev": true
     }
   }
-}
+}
```
felixarntz, line 12240: This change seems unrelated?

felixarntz (approved):
@jjgrainger Thank you for the updates, LGTM!

One tiny thing unrelated to the actual PR.

assets/js/plugin-check-admin.js
```diff
@@ -129,19 +132,33 @@
...
 	 * Run Checks.
 	 *
+	 * @since n.e.x.t
+	 *
 	 * @param {Object} data The response data.
+	 */
+	async function runChecks( data ) {
+		for ( let i = 0; i < data.checks.length; i++ ) {
+			const result = await runCheck( data.plugin, data.checks[ i ] );
+			console.log( result );
```
spacedmonkey:
```suggestion
            try {
			   const result = await runCheck( data.plugin, data.checks[ i ] );
			   console.log( result );
			} catch (e) {
              // Ignore for now.
             }   
```

My point about error handling seems to have been missed.

spacedmonkey (changes requested):
Requested one change and I will approve. We are nearly there.

package-lock.json
jjgrainger: @felixarntz this would have come about when resolving merge conflicts in when bringing in the JS linting. Looks like my editor automatically removed the newline at end of file which matches the `.editorconfig` we have.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @spacedmonkey this has been updated.

spacedmonkey (approved):
Looking good to me.

PR #115, opened 2023-03-30
Show check results on admin page

### LINKED ISSUE #35: Show check results on the page
Results/errors returned from each check should be shown in the tables in the Admin area on the `Dashboard >> Tools >> Plugin Check` page.

- #34 
- #86 
- #90 
- #84 

## Acceptance Criteria

**NOTE:** Read description of #34 before checking below steps.

- `render_result_row()` should available in the `Admin_Page` class (this class is introduced in https://github.com/10up/plugin-check/issues/32)
- Add `constructor()` method in the `Admin_AJAX` class which will take the`$render_callback` variable as a constructor parameter. (`Admin_AJAX` class is introduced in https://github.com/10up/plugin-check/issues/34)
- `Admin_Page` class may/should have the `open_result_table()` and `close_result_table()` methods to format the result into the table format.
- `template-table-open.php` file template should exist at `includes/Admin/Templates` and should have the table opening tags.
- `template-table-close.php` file template should exist at `includes/Admin/Templates` and should have the table closing tags.
- `template-result-row.php` file template should exist at `includes/Admin/Templates` - this file will handle the formatting of the error/warning message into the table's row (`<tr>`) format using the `script[type="text/template"]` tags. The row format's columns can be as below:
   - Error/warning message
   - Error Type (error/warning)
   - Error/Warning Standard name (e.g., WordPress.XSS.EscapeOutput.OutputNotEscaped, and so on. )
   - File name and path - on which error/warning occurs
   - Line and Column numbers - on which error/warning occurs
- `open_result_table()` method should includes the `template-table-open.php` template in it.
- `render_result_row()` method should includes the `template-result-row.php` template in it.
- `close_result_table()` method should includes the `template-table-close.php` template in it.
- `Admin_Page` class will use the `initialize_page()` method(see https://github.com/10up/plugin-check/issues/32). This method should call all three template functions to create the template markup in the admin footer with the `admin_footer` action.
    1. `open_result_table()`
    2. `render_result_row()`
    3. `close_result_table()`

**NOTE:** On the fresh admin screen, the table not exist on the page, and once results are returned, the table will be created and rendered, but that will be handled in #84.

## Tests Coverage

**Automated Tests:**
- Write a test case for the `Admin_Page::render_result_row()`.
- Write a test case for the `Admin_Page::open_result_table()` and `Admin_Page::close_result_table()`.
- Write a test case to test the row template markup.

**Manual Tests:**
- Once the user clicks the `Check it!` button, check if the proper HTML table and table row are showing up and appending on the screen.

jjgrainger:
Addresses acceptance criteria in #35 

Closes #35

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -1,6 +1,7 @@
 ( function ( pluginCheck ) {
 	const checkItButton = document.getElementById( 'plugin-check__submit' );
 	const resultsContainer = document.getElementById( 'plugin-check__results' );
+	const spinner = document.getElementById( 'plugin-check__spinner' );
 	const pluginsList = document.getElementById(
 		'plugin-check__plugins-dropdown'
 	);
@@ -16,6 +17,7 @@
 
 		// Empty the results container.
 		resultsContainer.innerText = '';
+		spinner.classList.add( 'is-active' );
 
 		getChecksToRun()
 			.then( setUpEnvironment )
@@ -25,6 +27,7 @@
 				console.log( data.message );
 
 				resultsContainer.innerHTML += 'Checks complete';
+				spinner.classList.remove( 'is-active' );
 			} )
 			.catch( ( error ) => {
 				console.error( error );
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -16,8 +16,7 @@
 	checkItButton.addEventListener( 'click', ( e ) => {
 		e.preventDefault();
 
-		// Empty the results container.
-		resultsContainer.innerText = '';
+		resetResults();
 		spinner.classList.add( 'is-active' );
 
 		getChecksToRun()
@@ -26,14 +25,22 @@
 			.then( cleanUpEnvironment )
 			.then( ( data ) => {
 				console.log( data.message );
-
-				resultsContainer.innerHTML += 'Checks complete';
 				spinner.classList.remove( 'is-active' );
 			} )
 			.catch( ( error ) => {
 				console.error( error );
 			} );
 	} );
+
+	/**
+	 * Reset the results container.
+	 *
+	 * @since n.e.x.t
+	 */
+	function resetResults() {
+		// Empty the results container.
+		resultsContainer.innerText = '';
+	}
 
 	/**
 	 * Setup the runtime environment if needed.
@@ -97,8 +104,6 @@
 				if ( ! responseData.data || ! responseData.data.message ) {
 					throw new Error( 'Response contains no data.' );
 				}
-
-				console.log( responseData.data.message );
 
 				return responseData.data;
 			} );
@@ -242,6 +247,8 @@
 		for ( const file in warnings ) {
 			renderFileResults( file, [], warnings[ file ] );
 		}
+
+		resultsContainer.innerHTML += '<p>' + wp.i18n.__( 'Checks complete', 'plugin-check' ) + '</p>';
 	}
 
 	/**
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -249,7 +249,6 @@
 	 */
 	function renderFileResults( file, errors, warnings ) {
 		const tableTemplate = wp.template( 'plugin-check-results-table' );
-		const rowTemplate = wp.template( 'plugin-check-results-row' );
 		const index = Date.now();
 
 		// Render the file table.
@@ -258,37 +257,34 @@
 			'plugin-check__results-body-' + index
 		);
 
-		// Loop over each result by the line, column and messages.
-		for ( const line in errors ) {
-			for ( const column in errors[ line ] ) {
-				for ( let i = 0; i < errors[ line ][ column ].length; i++ ) {
-					const message = errors[ line ][ column ][ i ].message;
-					const code = errors[ line ][ column ][ i ].code;
+		// Render results to the table.
+		renderResultRows( 'ERROR', errors, resultsTable );
+		renderResultRows( 'WARNING', warnings, resultsTable );
+	}
 
-					resultsTable.innerHTML += rowTemplate( {
-						line,
-						column,
-						file,
-						type: 'ERROR',
-						message,
-						code,
-					} );
-				}
-			}
-		}
+	/**
+	 * Renders a result row onto the file table.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} type    The result type. Either ERROR or WARNING.
+	 * @param {Object} results The results object.
+	 * @param {Object} table   The HTML table to append a result row to.
+	 */
+	function renderResultRows( type, results, table ) {
+		const rowTemplate = wp.template( 'plugin-check-results-row' );
 
 		// Loop over each result by the line, column and messages.
-		for ( const line in warnings ) {
-			for ( const column in warnings[ line ] ) {
-				for ( let i = 0; i < warnings[ line ][ column ].length; i++ ) {
-					const message = warnings[ line ][ column ][ i ].message;
-					const code = warnings[ line ][ column ][ i ].code;
+		for ( const line in results ) {
+			for ( const column in results[ line ] ) {
+				for ( let i = 0; i < results[ line ][ column ].length; i++ ) {
+					const message = results[ line ][ column ][ i ].message;
+					const code = results[ line ][ column ][ i ].code;
 
-					resultsTable.innerHTML += rowTemplate( {
+					table.innerHTML += rowTemplate( {
 						line,
 						column,
-						file,
-						type: 'WARNING',
+						type,
 						message,
 						code,
 					} );
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -23,6 +23,8 @@
 			.then( cleanUpEnvironment )
 			.then( ( data ) => {
 				console.log( data.message );
+
+				resultsContainer.innerHTML += 'Checks complete';
 			} )
 			.catch( ( error ) => {
 				console.error( error );
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -248,7 +248,8 @@
 			renderFileResults( file, [], warnings[ file ] );
 		}
 
-		resultsContainer.innerHTML += '<p>' + wp.i18n.__( 'Checks complete', 'plugin-check' ) + '</p>';
+		resultsContainer.innerHTML +=
+			'<p>' + wp.i18n.__( 'Checks complete', 'plugin-check' ) + '</p>';
 	}
 
 	/**
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -5,9 +5,10 @@
 	const pluginsList = document.getElementById(
 		'plugin-check__plugins-dropdown'
 	);
+	const templates = {};
 
 	// Return early if the elements cannot be found on the page.
-	if ( ! checkItButton || ! pluginsList || ! resultsContainer ) {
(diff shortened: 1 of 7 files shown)


templates/results-table.php
```diff
@@ -0,0 +1,32 @@
...
+				<td>
+					FILE:
+				</td>
+				<td colspan="4">
+					{{ data.file }}
+				</td>
+			</tr>
+			<tr>
+				<td>
+					Line
```
spacedmonkey: Should these be translated?

assets/js/plugin-check-admin.js
```diff
@@ -207,4 +211,89 @@
...
+						line,
+						column,
+						file,
+						type: 'ERROR',
+						message,
+						code,
+					} );
+				}
+			}
+		}
```
spacedmonkey: Repeated code again. There only two params that are different here are `errors` and the row type 'ERROR'. 

This should be very simple to change to a function.

spacedmonkey:
@jjgrainger There is no styling on the table. Is this expected?

Screenshot 2023-03-31 at 14 48 59

templates/results-table.php
```diff
@@ -0,0 +1,32 @@
+<script type="text/template" id="tmpl-plugin-check-results-table">
+	<table id="plugin-check__results-table-{{data.index}}" class="plugin-check__results-table">
```
spacedmonkey:
```suggestion
	<table id="plugin-check__results-table-{{data.index}}" class="widefat plugin-check__results-table">
```

templates/results-table.php
```diff
@@ -0,0 +1,32 @@
...
+		<thead>
+			<tr>
+				<td>
+					FILE:
+				</td>
+				<td colspan="4">
+					{{ data.file }}
+				</td>
+			</tr>
```
spacedmonkey: I am not sure that this makes sense as a table header. Maybe H tag makes more tense here.

spacedmonkey:
Screenshot 2023-03-31 at 14 53 06

When I click on a plugin without an error or warning, I see no message at all. It seems like the plugin checker has done nothing here. This maybe hard to test, if plugin even works.

includes/Admin/Admin_Page.php
```diff
@@ -154,4 +157,14 @@ public function filter_plugin_action_links( $actions, $plugin_file ) {
...
 	}
+
+	/**
+	 * Render the results table templates in the footer.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function admin_footer() {
+		include_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-table.php';
+		include_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-row.php';
```
spacedmonkey: Why include over require?

templates/results-table.php
```diff
@@ -0,0 +1,32 @@
...
+				<td>
+					Code
+				</td>
+				<td>
+					Message
+				</td>
+			</tr>
+		</thead>
+		<tbody id="plugin-check__results-body-{{data.index}}"></tbody>
+	</table>
```
spacedmonkey: Maybe we could add a <br /> or other spacer after each table.

templates/results-table.php
jjgrainger: Yes! Thanks for flagging. I've updated these.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @spacedmonkey I've created a function for this.

templates/results-table.php
jjgrainger: Thanks @spacedmonkey updated this with your suggestion

templates/results-table.php
jjgrainger: Thanks @spacedmonkey updated to a H tag and this looks better with the table styling too.

includes/Admin/Admin_Page.php
jjgrainger: Updated to require

templates/results-table.php
jjgrainger: Thanks @spacedmonkey I've added a line break after each table. This along with using a heading for the filename gives it spacing to make it easier to read.

jjgrainger:
Thanks @spacedmonkey 

I've updated the PR based on your feedback. As for your comment...

> When I click on a plugin without an error or warning, I see no message at all. It seems like the plugin checker has done nothing here. This maybe hard to test, if plugin even works.

I agree, and for now I've added a simple 'Checks complete' message once the process is done. I think it would be good to discuss what output through the process we'd want to display as part of the milestone review. Can you create an issue for this one?

spacedmonkey:
Screenshot 2023-03-31 at 18 10 39

This is looking a lot better! 

Couple of things. If I run this a couple of times, I see complete message twice, I don't know why this is. 

Another nitpic, but would not take long to implement, when you click the button, if you a spinner could show, then go away when request is complete that would be amazing. Jetpack took a while to process and did not know when it was complete. To add a spinner in WP is very simple. 

Next to the submit button add this. 

```
<span class="spinner"></span>
```
It will look like this. 
Screenshot 2023-03-31 at 18 17 30


Add a html id to it. When the button is clicked, add `is-active` class to the id and remove when the requests are complete. Should not take more than a couple of minutes to add and will make it clear to the user that something is happening.

assets/js/plugin-check-admin.js
```diff
@@ -207,4 +213,85 @@
...
+	 * Renders the file results table.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} file     The file name for the results.
+	 * @param {Object} errors   The file errors.
+	 * @param {Object} warnings The file warnings.
+	 */
+	function renderFileResults( file, errors, warnings ) {
+		const tableTemplate = wp.template( 'plugin-check-results-table' );
```
felixarntz: Parsing the template here will result in the same template being parsed several times, we should avoid that.

templates/results-table.php
```diff
@@ -0,0 +1,26 @@
+<script type="text/template" id="tmpl-plugin-check-results-table">
+	<h4><?php esc_html_e( 'FILE:', 'plugin-check' ); ?> {{ data.file }}</h4>
+	<table id="plugin-check__results-table-{{data.index}}" class="widefat plugin-check__results-table">
```
spacedmonkey:
```suggestion
	<table id="plugin-check__results-table-{{data.index}}" class="widefat striped plugin-check__results-table">
```

assets/js/plugin-check-admin.js
```diff
@@ -207,4 +213,85 @@
...
+	 * Renders a result row onto the file table.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} type    The result type. Either ERROR or WARNING.
+	 * @param {Object} results The results object.
+	 * @param {Object} table   The HTML table to append a result row to.
+	 */
+	function renderResultRows( type, results, table ) {
+		const rowTemplate = wp.template( 'plugin-check-results-row' );
```
felixarntz: Same here, since this function is also called several times.

I think to fix that in both instance, we should introduce a general function that handles the template rendering generically and is smart about it, something like:
```js
const templates = {};
function renderTemplate( templateSlug, data ) {
	if ( ! templates[ templateSlug ] ) {
		templates[ templateSlug ] = wp.template( templateSlug );
	}
	const template = templates[ templateSlug ];
	return template( data );
}
```

assets/js/plugin-check-admin.js
```diff
@@ -1,24 +1,30 @@
...
+		resultsContainer.innerText = '';
+
 		getChecksToRun()
 			.then( setUpEnvironment )
 			.then( runChecks )
 			.then( cleanUpEnvironment )
 			.then( ( data ) => {
 				console.log( data.message );
+
+				resultsContainer.innerHTML += 'Checks complete';
```
felixarntz: This should be a translated string, also we should probably put it at least in some minimal markup, like a `<p>` tag.

Last but not least, having this line here seems out of context. Shouldn't it rather happen as the last bit of `renderResults()` as it is connected to that function?

assets/js/plugin-check-admin.js
```diff
@@ -1,24 +1,30 @@
...
+	if ( ! checkItButton || ! pluginsList || ! resultsContainer ) {
 		console.error( 'Missing form elements on page' );
 		return;
 	}
 
 	checkItButton.addEventListener( 'click', ( e ) => {
 		e.preventDefault();
 
+		// Empty the results container.
+		resultsContainer.innerText = '';
```
felixarntz: Let's move this into a simple function like `resetResults()`.

spacedmonkey (dismissed):
I think barring the spinner, this is nearly there.

includes/Admin/Admin_Page.php
```diff
@@ -154,4 +157,14 @@ public function filter_plugin_action_links( $actions, $plugin_file ) {
...
 	}
+
+	/**
+	 * Render the results table templates in the footer.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function admin_footer() {
+		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-table.php';
+		require_once WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-row.php';
```
felixarntz: A few things here:
* These files should preferably only include the templates themselves, not the surrounding script tag. For that we should use the `wp_print_inline_script_tag()` function. To pass the output of these templates into the function, we can then use an output buffer.
* No need for `require_once`, this file doesn't contain PHP functions or anything like that and therefore is okay to be included multiple times.
```suggestion
		ob_start();
		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-table.php';
		$results_table_template = ob_get_clean();
		wp_print_inline_script_tag(
			$results_table_template,
			array(
				'id'   => 'tmpl-plugin-check-results-table',
				'type' => 'text/template',
			)
		);

		ob_start();
		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . '/templates/results-row.php';
		$results_row_template = ob_get_clean();
		wp_print_inline_script_tag(
			$results_row_template,
			array(
				'id'   => 'tmpl-plugin-check-results-row',
				'type' => 'text/template',
			)
		);
```

felixarntz (commented):
@jjgrainger Left some feedback here and above, but most of it looks good so far.

assets/js/plugin-check-admin.js
```diff
@@ -207,4 +222,108 @@
...
+			}
+		}
+
+		// Render remaining files with only warnings.
+		for ( const file in warnings ) {
+			renderFileResults( file, [], warnings[ file ] );
+		}
+
+		resultsContainer.innerHTML +=
+			'<p>' + wp.i18n.__( 'Checks complete', 'plugin-check' ) + '</p>';
```
felixarntz: Providing translations to JS is a bit trickier, and since this is the only translated string we have in JS, I would prefer if we handled that in PHP and made it available to JS.

While that may seem a bit overkill, why don't we put this `<p>` tag with the string in another `script[type="text/template"]` and then render it like that? That would then fully decouple the UI from the JS logic.

felixarntz (commented):
@jjgrainger One last comment here, but this is pretty much good to go.

joemcgill (commented):
This is looking good to me. I may have found a small bug that we could improve. It looks like if you click the button more than once wild checks are running, you will end up with multiple instances of the results. Probably the best way to handle this is to disable the button while the spinner is active. 

image

felixarntz (approved):
@jjgrainger Thank you for the updates. Approved, pending the feedback from @joemcgill, it would be great to make the button temporarily `disabled` until the process is complete (or has terminated due to an error).

joemcgill (approved):
The latest commit fixes the double results issue for me. Thanks, @jjgrainger!

PR #116, opened 2023-03-30
Implement PHPStan

### LINKED ISSUE #105: Setup PHPStan in repo

[PHPStan](https://phpstan.org/) is a tool to scan code to spot issues and improve code quality. There is a project called [phpstan-wordpress](https://github.com/szepeviktor/phpstan-wordpress), that is designed to use PHPStan on a WordPress project.

PHPStan can be integrated in CI pipeline and can improve code quality. 

### Designs

_No response_

_No response_

### Code of Conduct

spacedmonkey:
### Description of the Change
Implement PHPStan, a tool to improve code quality. Set the level to 5 for now, as this requires the min number of code changes. This number goes all the way up to 9, but we can up this level later. 

Most of the fixes are related to isset, which can be a little unpredictable to use. Use the null check, is a little cleaner, so stop PHPStan flagging it. Other issues, were related to a return value on action, which will be ignored by WordPress. 

Closes #105

### How to test the Change

PHPStan can be run locally like this. 

```
npm run phpstan
```

or 
```
composer phpstan
```

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

 
### Credits

Props @spacedmonkey

### Checklist:


phpstan.neon.dist
```diff
@@ -0,0 +1,16 @@
+parameters:
+  level: 5
```
spacedmonkey, line 2: This number goes up to 9. This makes the code very strict and harder to maintain, we can up this number at a later time.

includes/Admin/Admin_Page.php
```diff
@@ -60,8 +58,6 @@ public function add_page() {
 		);
 
 		add_action( "load-{$hook}", array( $this, 'initialize_page' ) );
-
-		return $hook;
```
felixarntz, line 66: This change looks wrong to me. Why would it be a bad practice to return `$hook` from here? We want to be able to use the page hook elsewhere.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -39,7 +39,8 @@ public function test_add_page() {
 		wp_set_current_user( $admin_user );
 		set_current_screen( 'dashboard' );
 
-		$page_hook    = $this->admin_page->add_page();
+		$this->admin_page->add_page();
+		$page_hook    = 'admin_page_plugin-check';
```
felixarntz: See above. I don't know if or why PHPStan complains about it, but hard-coding this here is clearly worse than the code was before.

felixarntz (changes requested):
@spacedmonkey Mostly looks good to me, except for one change that doesn't make sense to me.

includes/Admin/Admin_Page.php
jjgrainger: I agree with Felix here, is there a specific reason not to return `$hook` and use a hard coded string instead?

spacedmonkey: Actions do not use the return value for anything. Returning is pointless other than for the test. Adding just for a test is bad practice.

tests/Admin/Admin_Page_Tests.php
spacedmonkey: This change was made for phpstan. I don't see the problem with hard coding this, why would it change?

includes/Admin/Admin_Page.php
spacedmonkey: I can revert this change and mark it add a phpstan ignore in the code. But PHPStan was correct to flag this.

felixarntz: I don't think returning this value is a bad practice, or at least it is in line with how WordPress works, for two reasons:
* It indicates that the admin page was added successfully.
* Any code that wants to do something to the admin page needs this value.

I'd be in favor of adding an ignore statement.

tests/Admin/Admin_Page_Tests.php
felixarntz: Hard-coding is a bad practice, since this value comes from WordPress core. It shouldn't change without a change of code, but by hard-coding the value here we make the test more error-prone. For example, if we change where the admin page is added in the class (e.g. to another admin menu), the original version of this test would just work. But with the value hard-coded here we would also need to update it, which is not optimal.

includes/Admin/Admin_Page.php
spacedmonkey: I reverted it and I don't see an error 😕 . Either way, it is not fixed.

tests/Admin/Admin_Page_Tests.php
spacedmonkey: Fixed.

includes/Admin/Admin_Page.php
```diff
@@ -47,8 +47,6 @@ public function add_hooks() {
 	 * Registers the admin page under the tools menu.
 	 *
 	 * @since n.e.x.t
-	 *
-	 * @return string The hook identifier for the admin page.
```
felixarntz, line 53: This also needs to be added back. We are returning the hook, so it needs to be documented.

felixarntz (commented):
@spacedmonkey One last thing needed here.

felixarntz:
@spacedmonkey Now PHPStan fails again. I think as you initially mentioned, we'll need an ignore statement on that method.

spacedmonkey:
@felixarntz After our chat, I broke that method into two methods and PHPStan seems to like that. Unit tests and lints are now passing. I think the PR is good to go.

includes/Admin/Admin_Page.php
```diff
@@ -37,31 +37,37 @@ public function __construct() {
...
 			array( $this, 'render_page' )
 		);
+	}
 
+	/**
+	 * Registers the admin page under the tools menu.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function admin_menu() {
```
felixarntz: Method names should start with an action verb. Maybe `add_and_initialize_page()`? That's technically what this method does. `admin_menu()` doesn't mean much.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -39,7 +39,8 @@ public function test_add_page() {
 		wp_set_current_user( $admin_user );
 		set_current_screen( 'dashboard' );
 
-		$page_hook    = $this->admin_page->add_page();
+		$page_hook = $this->admin_page->add_page();
+		$this->admin_page->admin_menu();
```
felixarntz: This test change does not look right. Because you now call both methods, it will result in `add_page()` being called twice. The tests pass, but that's still not the correct logic flow.

Can you move the `$page_hook = $this->admin_page->add_page();` line right above the last assertion? This way the method is only called then, at which point all data used in assertions has already been recorded, and we only need the `$page_hook` variable in the last assertion anyway.

felixarntz (commented):
@spacedmonkey Thank you for the update. This is now good in terms of PHPStan, though I think the method name should be changed to something more descriptive, and there's a minor issue with the test.

includes/Admin/Admin_Page.php
```diff
@@ -22,6 +22,14 @@ class Admin_Page {
 	 */
 	protected $admin_ajax;
 
+	/**
+	 * Admin page hook suffix.
+	 *
+	 * @since n.e.x.t
+	 * @var string
+	 */
+	protected $hook_suffix;
```
felixarntz: Why introduce a class property for this? This is not used anywhere.

tests/Admin/Admin_Page_Tests.php
```diff
@@ -39,6 +39,7 @@ public function test_add_page() {
 		wp_set_current_user( $admin_user );
 		set_current_screen( 'dashboard' );
 
+		$this->admin_page->add_and_initialize_page();
 		$page_hook    = $this->admin_page->add_page();
```
felixarntz: See my previous comment, you're still calling both of these methods in a single test, which isn't right.

I mention above that the `$hook_suffix` class property isn't used, but you could make use of it here which would justify its existence more. Instead of calling `add_page()`, use reflection on the `Admin_Page` instance to get the value of the property.

felixarntz (commented):
@spacedmonkey My one test comment from before still hasn't really been addressed - we shouldn't call the method to add the page twice in the same test.

includes/Admin/Admin_Page.php
spacedmonkey: It is used on line 62. It means we can run `add_page` multiple times without it re-add the admin page. This property is a cache and means code is not run twice.

tests/Admin/Admin_Page_Tests.php
spacedmonkey: `$hook_suffix` is used 😕 In `add_page`. I added this class property so we can call `add_page` again here without it run the same code twice. I could maybe convert this to a class reflection to get the class property, but that seems a bit like overkill.

spacedmonkey:
@felixarntz I did action all your feedback. Calling `add_page` twice is now not a problem, as the result is cache in a class property. Please can re-review the code. Specially this commit https://github.com/10up/plugin-check/pull/116/commits/ec90229b67d04a127acd230fd1ab2ef66b6cf5a4

includes/Admin/Admin_Page.php
felixarntz: You mentioned above that we shouldn't introduce a return value just for a test because it's a bad pattern. I'd argue the same applies here: We shouldn't use a class property to avoid duplicate execution of a method just for a test.

The `add_page()` method should be called only from `add_and_initialize_page()`, it shouldn't be called again in the test.

tests/Admin/Admin_Page_Tests.php
felixarntz: See my above reply, using the class property just to avoid duplicate execution because of a test is not a good pattern. Using reflection to get the class property is more appropriate as that way this test works without an incorrect logic flow.

I think instead of line 43 here we should get the value from `Admin_Page::$hook_suffix`, and its usage from the `add_page()` method should be removed.

felixarntz:
@spacedmonkey I replied to your comments.

> Calling `add_page` twice is now not a problem, as the result is cache in a class property.

It's technically not a problem, however I don't think we should be adding the extra condition in production code just because of a test.

includes/Admin/Admin_Page.php
spacedmonkey: How do we get the value if we don't call the add_page method?

felixarntz: We call `add_and_initialize_page()` which calls `add_page()`, so the function will be called. It then sets the `$hook_suffix` class property value and we can get the value from there.

spacedmonkey: You mentioned you wanted developers to be able to get this value. How do you expect them do this?

tests/Admin/Admin_Page_Tests.php
spacedmonkey: The property is protected. So it isn't that easy.

includes/Admin/Admin_Page.php
felixarntz: > You mentioned you wanted developers to be able to get this value. How do you expect them do this?

I don't think that's strictly necessary. If we want external developers to use this value, we should probably add a `get_hook_suffix()` public method that returns the value of the class property. That may not be a bad idea, then we can also use it in the test without using reflection.

spacedmonkey: Okay. You want me to add a getter then? Please confirm, I feel we really going around in central and I want to confirm the path forward.

felixarntz: Yeah, I think a getter is probably the cleanest way for both the production code and tests.

felixarntz: We should initialize this as an empty string, otherwise the `@var string` above is incorrect, as it could also be `null` then.
```suggestion
	protected $hook_suffix = '';
```

includes/Admin/Admin_Page.php
```diff
@@ -39,31 +47,35 @@ public function __construct( Admin_AJAX $admin_ajax ) {
...
 			'plugin-check',
 			array( $this, 'render_page' )
 		);
+	}
 
-		add_action( "load-{$hook}", array( $this, 'initialize_page' ) );
-
-		return $hook;
+	/**
+	 * Registers the admin page under the tools menu.
```
felixarntz:
```suggestion
	 * Adds and initializes the admin page under the tools menu.
```

includes/Admin/Admin_Page.php
```diff
@@ -39,31 +47,35 @@ public function __construct( Admin_AJAX $admin_ajax ) {
...
-		add_action( 'admin_menu', array( $this, 'add_page' ) );
+		add_action( 'admin_menu', array( $this, 'add_and_initialize_page' ) );
 		add_filter( 'plugin_action_links', array( $this, 'filter_plugin_action_links' ), 10, 2 );
 
 		$this->admin_ajax->add_hooks();
 	}
 
 	/**
-	 * Registers the admin page under the tools menu.
+	 * Add the admin page.
```
felixarntz:
```suggestion
	 * Adds the admin page under the tools menu.
```

includes/Admin/Admin_Page.php
```diff
@@ -203,4 +215,15 @@ public function admin_footer() {
 			)
 		);
 	}
+
+	/**
+	 * Get hook_suffix property.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string
```
felixarntz:
```suggestion
	/**
	 * Gets the hook suffix under which the admin page is added.
	 *
	 * @since n.e.x.t
	 *
	 * @return string Hook suffix, or empty string if admin page was not added.
```

tests/Admin/Admin_Page_Tests.php
```diff
@@ -41,7 +41,9 @@ public function test_add_page() {
 		wp_set_current_user( $admin_user );
 		set_current_screen( 'dashboard' );
 
-		$page_hook    = $this->admin_page->add_page();
+		$this->admin_page->add_and_initialize_page();
+		$page_hook = $this->admin_page->get_hook_suffix();
+		$this->assertNotNull( $page_hook );
```
felixarntz: Based on that this value should always be a string:
```suggestion
		$this->assertNotEmpty( $page_hook );
```

felixarntz (commented):
@spacedmonkey A few last nit-picks here on slightly inaccurate documentation. I can quickly fix those things.

felixarntz (approved):
@spacedmonkey LGTM! This is going to help our code follow best practices in the future 👏

felixarntz:
@jjgrainger Could you please give this a review as well?

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
...
+  scanFiles:
+    - plugin-check.php
+    - tests/phpstan/stubs/wp-cli.php
+    - tests/phpstan/stubs/exitexception.php
+    - tests/phpstan/stubs/formatter.php
+    - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
+    - vendor/squizlabs/php_codesniffer/autoload.php
+  dynamicConstantNames:
+    - WP_PLUGIN_CHECK_PLUGIN_DIR_URL
+
```
mukeshpanchal27:
```suggestion
```

mukeshpanchal27 (approved):

PR #143, opened 2023-04-13
Fix PHPCS check when testing a single file plugin

### LINKED ISSUE #135: PHPCS Checks will run against all plugins when testing against a single file plugin
When testing against a single file plugin like Hello Dolly, the `Abstract_PHP_CodeSniffer_Check` will run the sniff against all plugins installed. I believe this is due to the [path passed to PHPCS](https://github.com/10up/plugin-check/blob/trunk/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php#L74) will result in the plugin directory as the plugins file is omitted from the path.

We should look at updated the [`Plugin_Context`](https://github.com/10up/plugin-check/blob/trunk/includes/Plugin_Context.php) or [`Check_Context`](https://github.com/10up/plugin-check/blob/trunk/includes/Checker/Check_Context.php) with logic to determine if this is a single file plugin and return the correct plugin folder or file.

jjgrainger:
Added a `location()` method to the `Plugin_Context` class to return the correct location for running PHPCS sniffs if checking a single file plugin or folder.

- I'm not 100% sure on the method name and open to suggestions.
- Alternatively, the logic could be handled inside the `Abstract_PHP_CodeSniffer_Check` if we think a method is not needed.

Closes #135

### DIFF
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -67,4 +67,22 @@ class Plugin_Context {
 	public function url( $relative_path = '/' ) {
 		return plugin_dir_url( $this->main_file ) . ltrim( $relative_path, '/' );
 	}
+
+	/**
+	 * Returns the plugin location.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The plugin file if single file plugin. Or the plugin folder.
+	 */
+	public function location() {
+		$path = $this->path();
+
+		// Return the plugin path and basename if the path matches the plugin directory.
+		if ( WP_PLUGIN_DIR . '/' === $path ) {
+			return $path . $this->basename();
+		}
+
+		return $path;
+	}
 }
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -71,7 +71,7 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 		// Create the default arguments for PHPCS.
 		$defaults = array(
 			'',
-			$result->plugin()->path( '' ),
+			$result->plugin()->location(),
 			'--report=Json',
 			'--report-width=9999',
 		);
--- a/tests/Plugin_Context_Tests.php
+++ b/tests/Plugin_Context_Tests.php
@@ -34,4 +34,15 @@ class Plugin_Context_Tests extends WP_UnitTestCase {
 	public function test_url_with_parameter() {
 		$this->assertSame( WP_PLUGIN_URL . '/' . $this->plugin_name . '/folder/file.css', $this->plugin_context->url( '/folder/file.css' ) );
 	}
+
+	public function test_location() {
+		$this->assertSame( WP_PLUGIN_DIR . '/' . $this->plugin_name . '/', $this->plugin_context->location() );
+	}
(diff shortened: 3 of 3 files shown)


includes/Plugin_Context.php
```diff
@@ -67,4 +67,22 @@ public function path( $relative_path = '/' ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The plugin file if single file plugin. Or the plugin folder.
+	 */
+	public function location() {
+		$path = $this->path();
+
+		// Return the plugin path and basename if the path matches the plugin directory.
+		if ( WP_PLUGIN_DIR . '/' === $path ) {
```
spacedmonkey, line 82: What about MU plugins?

includes/Plugin_Context.php
```diff
@@ -67,4 +67,22 @@ public function path( $relative_path = '/' ) {
...
+			return $path . $this->basename();
+		}
+
+		return $path;
```
spacedmonkey, line 86: Should this work instead. 

```
return str_replace( $this->main_file, '',  plugin_dir_path( $this->main_file ) );

includes/Plugin_Context.php
felixarntz: We don't support testing MU plugins at this point, so this is okay as is I think. Testing MU plugins will require lots of additional changes, so let's think about that separately in the future.

felixarntz (approved):
@jjgrainger Thanks, LGTM.

Please consider @spacedmonkey's feedback though.

includes/Plugin_Context.php
mukeshpanchal27: @jjgrainger Do we needs to open separate issue for MU plugins?

includes/Plugin_Context.php
mukeshpanchal27: @spacedmonkey The above code did not work for the single file plugin.

mukeshpanchal27 (approved):

includes/Plugin_Context.php
jjgrainger: Thanks @mukeshpanchal27 @spacedmonkey I've opened an issue to discuss MU Plugin support here https://github.com/10up/plugin-check/issues/154

includes/Plugin_Context.php
jjgrainger: Thanks @spacedmonkey 

I'm not sure if I've understood your suggestion entirely, but I don't think the suggestion works in this case.

We use the `plugin_dir_path` in the `path()` method so both `$path` and your suggestion are pretty much the same here. The `plugin_dir_path` also removes the plugin main file and returns the plugins folder so I don't think the `str_replace` would do anything?

Let me know if I've misunderstood your suggestion here.

includes/Plugin_Context.php
felixarntz: Fair point, though at this point testing MU plugins is out of scope. We can consider adding support in the future, but it will involve notably more than just this change, e.g. we have to fix loading order of the code etc. to work for MU plugins which will be even more challenging than it has been for regular plugins.

includes/Plugin_Context.php
spacedmonkey: Nevermind

spacedmonkey (approved):

PR #145, opened 2023-04-14
Use constants for AJAX actions

### LINKED ISSUE #130: Use constants instead of hard-coded strings for AJAX actions

The AJAX actions in the `Admin_AJAX` class currently use hard-coded strings for their names, which are then also referenced hard-coded in JavaScript. We should instead add `const`s for them to the `Admin_AJAX` class.

The `Admin_Page` class should then pass all of these names to the JS file by adding them to the `PLUGIN_CHECK` JS variable.

### Designs

_No response_

_No response_

### Code of Conduct

jjgrainger:
Closes #130

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -94,7 +94,11 @@ class Admin_Page {
 			'plugin-check-admin',
 			'const PLUGIN_CHECK = ' . json_encode(
 				array(
-					'nonce' => $this->admin_ajax->get_nonce(),
+					'nonce'                               => $this->admin_ajax->get_nonce(),
+					'action_get_checks_to_run'            => Admin_AJAX::ACTION_GET_CHECKS_TO_RUN,
+					'action_set_up_runtime_environment'   => Admin_AJAX::ACTION_SET_UP_ENVIRONMENT,
+					'action_run_checks'                   => Admin_AJAX::ACTION_RUN_CHECKS,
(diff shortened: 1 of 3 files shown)


spacedmonkey (approved):

includes/Admin/Admin_Page.php
```diff
@@ -94,7 +94,11 @@ public function enqueue_scripts() {
 			'plugin-check-admin',
 			'const PLUGIN_CHECK = ' . json_encode(
 				array(
-					'nonce' => $this->admin_ajax->get_nonce(),
+					'nonce'                               => $this->admin_ajax->get_nonce(),
+					'action_get_checks_to_run'            => Admin_AJAX::ACTION_GET_CHECKS_TO_RUN,
+					'action_set_up_runtime_environment'   => Admin_AJAX::ACTION_SET_UP_ENVIRONMENT,
+					'action_run_checks'                   => Admin_AJAX::ACTION_RUN_CHECKS,
+					'action_clean_up_runtime_environment' => Admin_AJAX::ACTION_CLEAN_UP_ENVIRONMENT,
```
felixarntz: Better to use camel case for the JS properties as that is in line with WP JS coding standards (e.g. `PLUGIN_CHECK.actionGetChecksToRun`).

felixarntz (approved):
@jjgrainger One small nit-pick here that would be good to address before merge, but generally LGTM.

includes/Admin/Admin_Page.php
jjgrainger: Thanks @felixarntz I've updated these to camel case

PR #151, opened 2023-04-25
Add a check to wp_get_theme to see if theme exists

### LINKED ISSUE #119: Add a check to `wp_get_theme` to see if theme exists
### Describe the bug

Add a check to `wp_get_theme` to see if theme exists, by calling ->exists method. 

### Steps to Reproduce

N / A

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

spacedmonkey:
### Description of the Change
Improve error handling, to check if theme exists. 

Closes #119 

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -125,13 +125,16 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param string $current Current theme.
 	 * @return string The theme name.
+	 *
+	 * @throws Exception Thrown if theme does not exist for some reason.
 	 */
-	public function get_theme_name( $current ) {
+	public function get_theme_name() {
 		$theme = wp_get_theme( $this->theme_slug, $this->themes_dir );
 		if ( ! $theme->exists() ) {
-			return $current;
+			throw new Exception(
+				__( 'Invalid theme: Theme does not exist for some reason.', 'plugin-check' )
+			);
 		}
 		return $theme->display( 'Name' );
 	}
--- a/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
+++ b/includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
@@ -125,10 +125,14 @@ class Use_Minimal_Theme_Preparation implements Preparation {
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param string $current Current theme.
 	 * @return string The theme name.
 	 */
-	public function get_theme_name() {
+	public function get_theme_name( $current ) {
 		$theme = wp_get_theme( $this->theme_slug, $this->themes_dir );
+		if ( ! $theme->exists() ) {
+			return $current;
+		}
 		return $theme->display( 'Name' );
 	}
 
--- a/tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
+++ b/tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
@@ -47,10 +47,19 @@ class Use_Minimal_Theme_Preparation_Tests extends WP_UnitTestCase {
 	public function test_get_theme_name() {
 		$preparation = new Use_Minimal_Theme_Preparation( $this->theme_slug, $this->theme_dir );
 		$cleanup     = $preparation->prepare();
-		$theme_name  = $preparation->get_theme_name( '' );
+		$theme_name  = $preparation->get_theme_name();
 		$cleanup();
 
 		$this->assertSame( $this->theme_name, $theme_name );
+	}
+
+	public function test_get_theme_name_error() {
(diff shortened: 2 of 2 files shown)


includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
```diff
@@ -125,10 +125,14 @@ public function get_theme_slug() {
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param string $current Current theme.
 	 * @return string The theme name.
 	 */
-	public function get_theme_name() {
+	public function get_theme_name( $current ) {
```
felixarntz: What is the purpose of this new parameter? That seems pointless to me as you're just passing an empty string to it, I think it should be removed.

felixarntz (changes requested):
@spacedmonkey The new check looks good, but I think the new parameter shouldn't be there.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
spacedmonkey: This is a callback to a filter. So if the theme doesn't exist, then it should fallback to the current value.

jjgrainger: I agree with @spacedmonkey here, we're using this method for the `pre_option_current_theme` filter and this update falls back to the current theme name if our minimal theme doesn't exist.

The empty string being passed is used in tests where we're expecting the minimal theme name. It may be worth adding another test for when the `Use_Minimal_Theme_Preparation` class uses a theme that does not exist to ensure the fallback is returned.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
```diff
@@ -47,7 +47,7 @@ public function test_prepare() {
 	public function test_get_theme_name() {
 		$preparation = new Use_Minimal_Theme_Preparation( $this->theme_slug, $this->theme_dir );
 		$cleanup     = $preparation->prepare();
-		$theme_name  = $preparation->get_theme_name();
+		$theme_name  = $preparation->get_theme_name( '' );
```
jjgrainger: I think it would be worth including a new test for `get_theme_name` when the theme does not exist. This would test for the new fallback you've added and asserts the method returns the value passed.

jjgrainger (commented):
Thanks @spacedmonkey looks good, just one suggestion on adding a test for the fallback logic you've added

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
felixarntz: That's a fair point on this being a filter callback. However, I don't think we should silently use the regular theme if the theme does not exist.

This class is to _enforce_ the use of our specific minimal test theme: If the test theme for whatever reason does not exist, that means something is wrong with our code. Falling back to the default theme is not the right thing to do as that will result in incorrect runtime data that is affected by the actual theme used on the site.

I think we should throw an exception if the theme does not exist and then get rid of the `$current` parameter again as for this particular filter callback it should never be used.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
felixarntz: See my reply above - while I agree with your explanation of this being a filter callback, the parameter should still be unnecessary.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
spacedmonkey: I guess it depends on how we want to handle this error state. 
1. Default to current theme - do nothing. 
2. Call `wp_die` with error message. 
3. Call `_doing_it_wrong`  with error message and default to current theme. 
4. Throw an exception and trigger fatal error handling. 

This is super edge, so I don't think we should make a big deal of this. I will defer to others thoughts on this.

tests/Checker/Preparations/Use_Minimal_Theme_Preparation_Tests.php
spacedmonkey: Adding a unit test is a good idea. I will wait until we decided on how the error handling should work.

includes/Checker/Preparations/Use_Minimal_Theme_Preparation.php
felixarntz: Because it's indeed super edge case, I think we should just throw an exception and be done with it.

jjgrainger: Thanks @felixarntz I agree with throwing an exception here 👍

spacedmonkey: Now throws `exception` .

jjgrainger (approved):

felixarntz (approved):
Thanks @spacedmonkey!

Issue #111, opened 2023-03-27
Create util function to check if running plugin check command 


spacedmonkey:
### Is your enhancement related to a problem? Please describe.

This [code](https://github.com/10up/plugin-check/blob/391b7eab651d75d48b9cca38d763681dd4f306d5/includes/Checker/CLI_Runner.php#L34-L48) is repeated in https://github.com/10up/plugin-check/pull/100. 

Create a util function method, so code is not reported.  

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #152, opened 2023-04-25
Inject runner classes into other classes.

spacedmonkey:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

- [ ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).


cli.php
```diff
@@ -43,21 +46,16 @@
  */
 WP_CLI::add_hook(
 	'before_wp_load',
-	function() {
-		if ( empty( $_SERVER['argv'] ) || 3 > count( $_SERVER['argv'] ) ) {
+	function() use ( $runner ) {
+		if( !$runner->is_plugin_check() ){
```
jjgrainger, line 50: This approach would resolve https://github.com/10up/plugin-check/issues/111 if we can use the runner classes or it's methods in this way.

cli.php
```diff
@@ -31,8 +33,9 @@
 	$context = new Plugin_Context( __DIR__ . '/plugin-check.php' );
 }
 
+$runner = new CLI_Runner();
```
jjgrainger, line 36: This is a possibility and helps to remove the repeated logic by being able to use `is_plugin_check()` later in this file. However, there are a couple of considerations.

- The `Abstract_Check_Runner` constructor calls `did_action` which would not be available when executed as this file would run before any of WordPress was loaded.
- We use the `did_action( 'muplugins_loaded' );` in the constructor to determine if the runner was instantiated early enough in the WordPress load process in order to perform runtime checks.
- For additional context, we use this action in the `Plugin_Request_Utility` to run the `Universal_Runtime_Preparation` [here](https://github.com/10up/plugin-check/blob/trunk/includes/Utilities/Plugin_Request_Utility.php#L97)

includes/CLI/Plugin_Check_Command.php
```diff
@@ -137,20 +131,18 @@ public function check( $args, $assoc_args ) {
...
-			Plugin_Request_Utility::destroy_runner();
-
 			if ( isset( $runtime_setup ) ) {
 				$runtime_setup->cleanup();
 				WP_CLI::line( __( 'Cleaning up runtime environment.', 'plugin-check' ) );
 			}
 
 			WP_CLI::error( $error->getMessage() );
+		} finally {
+			$this->runner->prepare();
```
jjgrainger, line 143: The prepare method here needs be called well before the `run` method as it handles the `Universal_Runtime_Preparations` needed to perform runtime checks.

These preparations in particular needed to be called early in the WordPress load process. We currently call this in the `muplugins_loaded` action in the `Plugin_Request_Utility`.

object-cache.copy.php
```diff
@@ -33,10 +33,5 @@ function plugin_check_initialize_runner() {
 	}
 
 	require_once $plugin_dir . 'vendor/autoload.php';
-
-	if ( class_exists( 'WordPress\Plugin_Check\Utilities\Plugin_Request_Utility' ) ) {
-		// Initialize the Check Runner class based on the request.
-		WordPress\Plugin_Check\Utilities\Plugin_Request_Utility::initialize_runner();
```
jjgrainger, line 39: We're using the object-cache.php file to call this method early in the WordPress load process so that we can handle any required preparations for runtime checks. It's a bit hacky but the best approach we had to start with. 

This method call is what creates the runner and setups the action to call the `prepare` method for runtime checks. Without this we wouldn't be able to perform runtime checks for both AJAX and CLI requests.

Also, the sole purpose of this file is to call this method, if we could find a way to prepare the runtime environment without we could remove this file and the logic surrounding its creation.

cli.php
```diff
@@ -43,21 +46,16 @@
...
-			'wp' === substr( $_SERVER['argv'][0], -2 ) &&
-			'plugin' === $_SERVER['argv'][1] &&
-			'check' === $_SERVER['argv'][2]
-		) {
-			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
-				if ( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
-					WP_CLI::error( 'Unable to copy object-cache.php file.' );
-				}
+		if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+			if ( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
```
jjgrainger, line 55: Just to highlight the purpose of the object cache here for context, we're using this file to call the `Plugin_Request_Utility::initialize_runner();` in order to run required preparations early in the WP load process to perform runtime checks.

includes/Admin/Admin_AJAX.php
```diff
@@ -94,27 +112,14 @@ public function set_up_environment() {
...
 		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
 		try {
-			$runner->set_check_slugs( $checks );
-			$runner->set_plugin( $plugin );
-			$checks_to_run = $runner->get_checks_to_run();
+			$this->runner->set_check_slugs( $checks );
+			$this->runner->set_plugin( $plugin );
+			$checks_to_run = $this->runner->get_checks_to_run();
```
jjgrainger, line 122: The `get_checks_to_run` method returns checks based on logic for when the runner was initialized (`did_action( 'muplugins_loaded' );`). In this situation where we instantiate and pass the runner to the class I believe the runner would be created too late and this method will only ever return static checks.

jjgrainger (commented):
Thanks @spacedmonkey 

I've reviewed the PR as a means to provide more context on the runner setups.

The reasoning is to create and prepare a test environment early in the WP load process in order to test runtime checks that execute plugin code. To do this we wanted to create a test environment and db table, then prepare the environment by activating specific plugins/themes as well as potentially creating test posts.

To do this we're checking early to see if a runtime check is requested and then make sure that the object-cache.php file is created in order to instantiate and prepare the runner early within the WP load process.

Let me know if you have any follow up questions and myself or @felixarntz can get back to you.

felixarntz:
@spacedmonkey Which issue does this PR address?

While I'm not necessarily against this change, we need to focus on the critical problems that need to be addressed prior to the repository transfer for now. This PR is something new that we haven't discussed before - which normally is fine, but we need to get to a clean cut off point, so I would prefer if you could raise an issue for this first where we can discuss whether and why this is critical to potentially do before.

PR #155, opened 2023-04-26
Fix some file tables display no results in admin page

### LINKED ISSUE #137: Admin Page: Some file tables display no results
> When I run the checker on some plugins like Elementor, a table appears but it has no errors.  Should these appear within the report if no errors?  Might confuse the user. 

Looking into this it appears the tables are created so quickly they can share the same ID ([Which is a timestamp](https://github.com/10up/plugin-check/blob/trunk/assets/js/plugin-check-admin.js#L280) when the table is created). Results are not missing but added to the wrong table (the first table created with the same ID).

The fix would be to generate a truely unique ID for each of the tables rather than relying on a timestamp.

image-2

mukeshpanchal27:
### Description of the Change
Use a unique ID so it will generate the table and the error or notice properly. Use similar code that WordPress core [postbox.js](https://github.com/WordPress/wordpress-develop/blob/trunk/src/js/_enqueues/admin/postbox.js#L390) uses.

Closes #137 

### Checklist:

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -283,7 +283,7 @@
 	 * @param {Object} warnings The file warnings.
 	 */
 	function renderFileResults( file, errors, warnings ) {
-		const index = Date.now();
+		const index = parseInt( Math.random() * 100000, 10 );
 
 		// Render the file table.
 		resultsContainer.innerHTML += renderTemplate(
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -283,7 +283,9 @@
 	 * @param {Object} warnings The file warnings.
 	 */
 	function renderFileResults( file, errors, warnings ) {
-		const index = parseInt( Math.random() * 100000, 10 );
+		const index =
+			Date.now().toString( 36 ) +
+			Math.random().toString( 36 ).substr( 2 );
 
 		// Render the file table.
 		resultsContainer.innerHTML += renderTemplate(


assets/js/plugin-check-admin.js
```diff
@@ -283,7 +283,7 @@
 	 * @param {Object} warnings The file warnings.
 	 */
 	function renderFileResults( file, errors, warnings ) {
-		const index = Date.now();
+		const index = parseInt( Math.random() * 100000, 10 );
```
spacedmonkey: I have used [uuid](https://www.npmjs.com/package/uuid) in the past. But adding this library might overkill. 

This might be another option. 

```
const uid = function(){
    return Date.now().toString(36) + Math.random().toString(36).substr(2);
}
``` 
From https://stackoverflow.com/a/53116778

jjgrainger (dismissed):
Thanks @mukeshpanchal27 Looks good, but I think it's worth taking @spacedmonkey approach here for generating the unique ID here. Please can you take a look.

felixarntz (commented):
@mukeshpanchal27 LGTM, except I agree that using the function @spacedmonkey outlined would be a bit better.

mukeshpanchal27:
Thanks to all for the feedback. I have updated the unique ID approach as suggested. The PR is ready for review.

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks! Addressed in https://github.com/10up/plugin-check/pull/155/commits/b8a9e01f60b884aaf4ed51ae629ea30ecd68afb1

spacedmonkey (approved):

felixarntz (approved):
Thanks @mukeshpanchal27!

jjgrainger (approved):

PR #157, opened 2023-04-28
Refactor to remove repeated code.

spacedmonkey:
### Description of the Change
While reviewing code, I noticed a lot of repeated code in Admin_Ajax. Simple refactor to tidy up the code. 

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

- [ ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).


includes/Admin/Admin_AJAX.php
```diff
@@ -89,27 +89,11 @@ public function get_nonce() {
 	 */
 	public function set_up_environment() {
 		// Verify the nonce before continuing.
-		$valid_request = $this->verify_request( filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_FULL_SPECIAL_CHARS ) );
+		$this->validate_request();
```
jjgrainger: I had originally approached this that way before but it was recommended to decouple this and handle the response outside of the verify/validate method. (see https://github.com/10up/plugin-check/pull/97#discussion_r1136040211)

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +223,70 @@ public function run_checks() {
...
+	 * @since n.e.x.t
+	 *
+	 * @return AJAX_Runner
+	 */
+	protected function get_runner() {
+		$runner = Plugin_Request_Utility::get_runner();
+
+		if ( is_null( $runner ) ) {
+			Plugin_Request_Utility::initialize_runner();
+			$runner = Plugin_Request_Utility::get_runner();
```
jjgrainger: The `initialize_runner` method handles other logic alongside creating the correct runner (and could handle more in the future). As we know we're in the AJAX context I think we should simply instantiate an `AJAX_Runner` as before rather than use the `Plugin_Request_Utility` methods here.

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +223,70 @@ public function run_checks() {
...
+		}
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
+			wp_send_json_error(
+				new WP_Error( 'invalid-runner', __( 'AJAX Runner was not initialized correctly.', 'plugin-check' ) ),
+				500
+			);
```
jjgrainger: Similar to my other comment, we may want to return the `WP_Error` rather than send the JSON response here. (see https://github.com/10up/plugin-check/pull/97#discussion_r1136040211)

Alternatively, we could throw an `Exception` instead and call `get_runner` in the try/catch block?

jjgrainger (changes requested):
Thanks @spacedmonkey looks good, left a couple of comments based on previous feedback in another PR

includes/Admin/Admin_AJAX.php
spacedmonkey: Then what is the point of ` Plugin_Request_Utility::get_runner();` then. Why not just create a `AJAX_Runner` here?

includes/Admin/Admin_AJAX.php
spacedmonkey: This is just the same logic that was there before. Why change it?

includes/Admin/Admin_AJAX.php
spacedmonkey: I don't think those two things are the same. This does change any of the logic, just moves it around.

felixarntz: Moving the logic into reusable functions makes sense, but to @jjgrainger's point we shouldn't couple the new utility methods to send JSON response data. The response should be handled separately by the AJAX callbacks, the new utility functions should rather throw an `Exception` or return `WP_Error` that can then be returned as JSON.

includes/Admin/Admin_AJAX.php
felixarntz: +1 to what @jjgrainger said, this is changing the logic from what it was before. We should create an `AJAX_Runner` here, not call `Plugin_Request_Utility::initialize_runner()` and `Plugin_Request_Utility::get_runner()`, as that also runs other logic.

includes/Admin/Admin_AJAX.php
felixarntz: See https://github.com/10up/plugin-check/pull/157/files#r1180537801, the utility methods should have the responsibility of what their name is, not return a JSON response. For example this method should return the runner, or throw an exception if there is an error. The AJAX callbacks can then use `try { ... } catch` and return a JSON error response in case of an exception.

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +223,70 @@ public function run_checks() {
...
+	 *
+	 * @since n.e.x.t
+	 */
+	protected function validate_request() {
+		$valid_request = $this->verify_request( filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_FULL_SPECIAL_CHARS ) );
+
+		if ( is_wp_error( $valid_request ) ) {
+			wp_send_json_error( $valid_request, 403 );
+		}
+	}
```
felixarntz: See above. This method shouldn't be necessary at all. I would suggest we rather modify the `verify_request()` method to throw an exception instead of returning `WP_Error`. That way we can catch a potential exception with the same code that we would use also for the other methods, like `get_runner()` above.

felixarntz (changes requested):
@spacedmonkey Moving this logic to reusable methods is a great idea, however I agree with @jjgrainger that we should decouple those methods from sending JSON response data. The methods should just return what they're named for, or throw an exception if something goes wrong.

The AJAX callbacks can then catch the exception and return it as a JSON error response.

includes/Admin/Admin_AJAX.php
spacedmonkey: If we do that, that could mean repeated the same code every time this method is called. I am trying avoid lots of boilerplate code here.

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +223,69 @@ public function run_checks() {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return AJAX_Runner
+	 */
+	protected function get_runner() {
+		$runner = Plugin_Request_Utility::get_runner();
+
+		if ( is_null( $runner ) ) {
+			$runner = new AJAX_Runner();
```
spacedmonkey: @felixarntz @jjgrainger By doing this, should we not also be calling the prepare method on this runner?

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +239,48 @@ public function run_checks() {
...
+	 * @since n.e.x.t
+	 *
+	 * @return AJAX_Runner
+	 */
+	protected function get_runner() {
+		$runner = Plugin_Request_Utility::get_runner();
+
+		if ( ! ( $runner instanceof AJAX_Runner ) ) {
+			$runner = new AJAX_Runner();
+		}
```
felixarntz, line 255: This is a functional change we shouldn't make: This method should only instantiate a new `AJAX_Runner` if `Plugin_Request_Utility::get_runner()` returns `null`. Otherwise, if the runner returned from that method itself is not an `AJAX_Runner`, we should throw an exception and then the underlying AJAX callback should terminate with a JSON error, similar to how that works today.

felixarntz (commented):
This looks almost good to go, however one thing needs to be addressed.

includes/Admin/Admin_AJAX.php
spacedmonkey: Why would we want to do that? If it returns null, it is okay to create a new runner right? If `get_runner` returns a CLI_Runner, then error? Why error here? What is the benefit of erroring when we can just create an instance?

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +239,48 @@ public function run_checks() {
 		);
 	}
 
+
```
mukeshpanchal27, line 242:
```suggestion
```
Remove extra line.

includes/Admin/Admin_AJAX.php
```diff
@@ -278,6 +239,48 @@ public function run_checks() {
...
+	/**
+	 * Get array of checks to run.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array
+	 */
+	protected function get_checks() {
+		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
```
mukeshpanchal27, line 268: Similar function present in [AJAX_Runner::get_check_slugs_param()](https://github.com/10up/plugin-check/blob/trunk/includes/Checker/AJAX_Runner.php#L65-L74)

mukeshpanchal27 (commented):
Thanks @spacedmonkey, Can we make util function for `filter_input( INPUT_POST, 'nonce', FILTER_SANITIZE_FULL_SPECIAL_CHARS )` also as it used four time in this file also I left some feedbacks.

PR #160, opened 2023-05-10
Improve result message copy

### LINKED ISSUE #138: Improve the copy when a plugin is valid
When the plugin has no errors and is valid, we should look at improving the copy so that it's more obvious that the plugin is valid.  At the moment, it just says checks complete.  Maybe we add a checkmark with this plugin has passed our checks, or something along those lines. 

image-3

mukeshpanchal27:
The PR improve the result message copy. If the result shows error/notice then it show WordPress error notice and on sucees it show success notice.

**Success notice:**
Screenshot 2023-05-10 at 11 13 48 AM

**Error notice:**
Screenshot 2023-05-10 at 11 15 41 AM

Closes #138 

### Checklist:


assets/js/plugin-check-admin.js
```diff
@@ -265,12 +266,28 @@
...
+			'plugin-check__message'
+		);
+		if ( resultsMessage ) {
+			const resultsMessageChild = resultsMessage.querySelector( 'p' );
+			if ( errorWarningMessage ) {
+				resultsMessage.classList.remove( 'notice-success' );
+				resultsMessage.classList.add( 'notice-error' );
+				resultsMessageChild.innerHTML += ' ' + pluginCheck.errorMessage;
+			} else {
+				resultsMessageChild.innerHTML +=
```
spacedmonkey:
```suggestion
				resultsMessage.classList.add( 'notice-success' );
				resultsMessage.classList.remove( 'notice-error' );
				resultsMessageChild.innerHTML +=
```

If the message goes from a error to success if run multiple times. Let's be overly careful here.

assets/js/plugin-check-admin.js
```diff
@@ -265,12 +266,28 @@
...
+					' ' + pluginCheck.successMessage;
+			}
+		}
```
felixarntz: Instead of all this logic that manually tampers with the output of the `renderTemplate()` function, we should instead adjust that template itself (i.e. the `results-complete.php` file) to use placeholders (e.g. for the class and the additional sentence).

We should then pass the dynamic parts in the `data` parameter to the `renderTemplate()` function.

felixarntz (changes requested):
@mukeshpanchal27 I think there's a much simpler solution to this - we are already using a templating system, so it's better to rely on that rather than manually tweak the DOM.

assets/js/plugin-check-admin.js
```diff
@@ -252,9 +252,10 @@
 	 */
 	function renderResults( results ) {
 		const { errors, warnings } = results;
-
+		let errorWarningMessage = false;
```
jjgrainger: @mukeshpanchal27 could we not check the `errors` or `warnings` length here to see if there are errors? Instead of always updating the variable in for loops?

jjgrainger: On second thought, I think we'd also want to move this logic out to the `runChecks`. This `renderResults` function is run for each individual check, so when more than one check is run the complete message will display multiple times.

assets/js/plugin-check-admin.js
jjgrainger: @mukeshpanchal27 @felixarntz I agree here, lets make use of the template to set the message text and notice class depending on if there are errors or not.

jjgrainger (changes requested):
Thanks @mukeshpanchal27 looking good, just a few changes here

spacedmonkey (approved):

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks @spacedmonkey for the feedback. Based on the other feedback i use other approach to solve the issue. Please take a look. Thanks!

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks both for the feedback. In https://github.com/10up/plugin-check/pull/160/commits/dc862e281c38c9be01984985a9f76bdac7e72575 i have address this feedback. Please take a look. Thanks!

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks @jjgrainger, In https://github.com/10up/plugin-check/pull/160/commits/dc862e281c38c9be01984985a9f76bdac7e72575 i address this feedback. Please take a look.

mukeshpanchal27:
@jjgrainger @felixarntz I've addressed your feedback, please take another look.

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
...
 	 */
 	async function runChecks( data ) {
+		let isSuccessMessage = true;
+		let messageClass = '';
+		let messageText = '';
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			try {
 				const results = await runCheck( data.plugin, data.checks[ i ] );
+				const errorsLength = Object.values( results.errors ).length;
+				const warningsLength = Object.values( results.errors ).length;
```
spacedmonkey:
```suggestion
				const warningsLength = Object.values( results.warnings ).length;
```

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
 	 * @param {Object} data The response data.
 	 */
 	async function runChecks( data ) {
+		let isSuccessMessage = true;
+		let messageClass = '';
+		let messageText = '';
```
jjgrainger: @mukeshpanchal27 if we're defaulting to a success message would it make sense to include the success values for the class and text by default too? That way we don't need the else statement later.

```suggestion
		let isSuccessMessage = true;
		let messageClass = 'notice-success';
		let messageText = pluginCheck.successMessage;
```

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
...
 			}
 		}
+
+		if ( isSuccessMessage ) {
+			messageClass = 'notice-success';
+			messageText = pluginCheck.successMessage;
+		} else {
+			messageClass = 'notice-error';
+			messageText = pluginCheck.errorMessage;
+		}
```
jjgrainger: @mukeshpanchal27 see my point above, if we default to a success message we can simply check for if success is false and update accordingly.

```suggestion
		if ( ! isSuccessMessage ) {
			messageClass = 'notice-error';
			messageText = pluginCheck.errorMessage;
		}
```

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
...
 	 */
 	async function runChecks( data ) {
+		let isSuccessMessage = true;
+		let messageClass = '';
+		let messageText = '';
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			try {
 				const results = await runCheck( data.plugin, data.checks[ i ] );
+				const errorsLength = Object.values( results.errors ).length;
+				const warningsLength = Object.values( results.warnings ).length;
```
jjgrainger, line 180: @mukeshpanchal27 I wonder if this is an opportunity here to include the warning and errors count as part of the AJAX response instead. That way we can just reference the values in the response than calculating it here.

jjgrainger (commented):
Thanks @mukeshpanchal27 overall looks good, left some minor comments to review but not necessarily needed to be addressed or blocking approval

assets/js/plugin-check-admin.js
spacedmonkey: Count the number of elements here is extremely quick in JS. Adding an extra field to the ajax, adding more php overhead and my bytes down the wire. Unneeded IMO.

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
...
+			messageText = pluginCheck.errorMessage;
+		}
+
+		resultsContainer.innerHTML += renderTemplate(
+			'plugin-check-results-complete',
+			{
+				class: messageClass,
+				message: messageText,
+			}
+		);
```
felixarntz, line 215: Can you please move all the new logic back into the `renderResults()` function below? That's where this template was originally rendered, so I'm not sure why it would need to be moved here now, which intertwines it more with other higher-level logic.

All new logic is based on the `results` variable, which is passed to `renderResults()` anyway, so we can have the logic in there without any notable adjustments.

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,41 @@
...
+			messageText = pluginCheck.successMessage;
+		} else {
+			messageClass = 'notice-error';
+			messageText = pluginCheck.errorMessage;
+		}
+
+		resultsContainer.innerHTML += renderTemplate(
+			'plugin-check-results-complete',
+			{
+				class: messageClass,
```
felixarntz: The `class` name is somewhat problematic to use since it is also a language keyword. I'd prefer that we use something else like `className`.

Or alternatively, we change this to a more general `type` value, to which we just pass "success" or "error". If we do that, then the template itself would need to use `notice-{{ data.type }}`.

felixarntz (commented):
@mukeshpanchal27 Two follow up comments here, but this looks very close now.

assets/js/plugin-check-admin.js
jjgrainger: @felixarntz @mukeshpanchal27 this came from my previous review https://github.com/10up/plugin-check/pull/160#discussion_r1190137462

We call `renderResults` for each check, so if we keep this logic here the final 'Checks Complete' message will display multiple times. Moving this here means the message would only display once after all checks have been run.

mukeshpanchal27: @felixarntz Please check @jjgrainger's [comment here](https://github.com/10up/plugin-check/pull/160#discussion_r1190137462)

> On second thought, I think we'd also want to move this logic out to the `runChecks`. This `renderResults` function is run for each individual check, so when more than one check is run the complete message will display multiple times.

assets/js/plugin-check-admin.js
mukeshpanchal27: I agree with @spacedmonkey. What is your thought @jjgrainger?

assets/js/plugin-check-admin.js
mukeshpanchal27: Make sense. Thanks for pointing it out. Address in https://github.com/10up/plugin-check/pull/160/commits/8d6c539c8824f17777e44c14c98a01d55cb25362

assets/js/plugin-check-admin.js
felixarntz: @mukeshpanchal27 Ahh, I see, thanks for clarifying. In that case, I'd suggest we create a new function, e.g. `renderResultsMessage()`, with the new logic instead of adding that to the other higher-level function, which makes it more complex.

mukeshpanchal27: Thanks @felixarntz for the review. Addressed in https://github.com/10up/plugin-check/pull/160/commits/61dc1cc450c5619cf0a88cb42abd7199bc226265

assets/js/plugin-check-admin.js
mukeshpanchal27: Addressed in https://github.com/10up/plugin-check/pull/160/commits/61dc1cc450c5619cf0a88cb42abd7199bc226265

jjgrainger (approved):
Thanks @mukeshpanchal27 looks good, approved

assets/js/plugin-check-admin.js
spacedmonkey: @mukeshpanchal27 This was dismissed but the change was not made. What happened here?

assets/js/plugin-check-admin.js
```diff
@@ -172,14 +172,50 @@
...
+	 * @param {boolean} isSuccessMessage Whether the message is a success message.
+	 */
+	function renderResultsMessage( isSuccessMessage ) {
+		let messageType = 'success';
+		let messageText = pluginCheck.successMessage;
+
+		if ( ! isSuccessMessage ) {
+			messageType = 'error';
+			messageText = pluginCheck.errorMessage;
+		}
```
spacedmonkey:
```suggestion
		const messageType = isSuccessMessage ? 'success' : 'error';
		const messageText = isSuccessMessage ? pluginCheck.successMessage : pluginCheck.errorMessage;
```

Const are a little better for performance of js. Maybe this is less readable.

spacedmonkey (approved):

felixarntz (approved):

PR #161, opened 2023-05-10
Util function to check if running plugin check command

### LINKED ISSUE #111: Create util function to check if running plugin check command

mukeshpanchal27:
Closes #111 

### Checklist:

### DIFF
--- a/cli.php
+++ b/cli.php
@@ -8,9 +8,9 @@
  * @since n.e.x.t
  */
 
+use WordPress\Plugin_Check\Checker\CLI_Runner;
 use WordPress\Plugin_Check\Plugin_Context;
 use WordPress\Plugin_Check\CLI\Plugin_Check_Command;
-use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
 if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 	return;
@@ -45,14 +45,12 @@ WP_CLI::add_command( 'plugin', $plugin_command );
 WP_CLI::add_hook(
 	'before_wp_load',
 	function() {
-		if ( Plugin_Request_Utility::is_plugin_check() ) {
+		if ( CLI_Runner::is_plugin_check() ) {
 			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
 				if ( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
 					WP_CLI::error( 'Unable to copy object-cache.php file.' );
 				}
 			}
-		} else {
-			return;
 		}
 	}
 );
--- a/cli.php
+++ b/cli.php
@@ -10,6 +10,7 @@
 
 use WordPress\Plugin_Check\Plugin_Context;
 use WordPress\Plugin_Check\CLI\Plugin_Check_Command;
+use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
 if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 	return;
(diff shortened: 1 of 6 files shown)


cli.php
```diff
@@ -44,20 +45,14 @@
...
-		) {
+		if ( Plugin_Request_Utility::is_plugin_check() ) {
 			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
 				if ( ! copy(  __DIR__ . '/object-cache.copy.php', ABSPATH . 'wp-content/object-cache.php' ) ) {
 					WP_CLI::error( 'Unable to copy object-cache.php file.' );
 				}
 			}
+		} else {
+			return;
 		}
```
spacedmonkey, line 56:
```suggestion
		}
		return;
```

spacedmonkey: There is no other code after this return. I am 99% sure that, this is not needed.

spacedmonkey:
@mukeshpanchal27 I played around with this ticket in https://github.com/10up/plugin-check/pull/152/files. I don't think the idea works, but it is still interesting to compare.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -134,4 +134,28 @@ public static function destroy_runner() {
...
 	}
+
+	/**
+	 * Checks if the current request is a CLI request for the Plugin Checker.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns true if is a CLI request for the plugin check else false.
+	 */
+	public static function is_plugin_check() {
```
felixarntz: Instead of a new method in this general class (which is not specific to CLI at all), I suggest we use the `CLI_Runner::is_plugin_check()` method instead and make that static.

includes/Checker/CLI_Runner.php
```diff
@@ -32,15 +33,7 @@ class CLI_Runner extends Abstract_Check_Runner {
 	 * @return bool Returns true if is an CLI request for the plugin check else false.
 	 */
 	public function is_plugin_check() {
```
felixarntz: See my other comment, I think making this method static makes more sense than introducing a similar method somewhere else that is not specific to CLI.

It's perfectly reasonable for this method to be static since it only checks something in the global scope and has nothing to do with an instance of this class.

felixarntz (changes requested):
@mukeshpanchal27 Looks good so far, though I think a cleaner solution would be to keep using the `CLI_Runner` method we already have for that purpose.

includes/Checker/CLI_Runner.php
jjgrainger: @mukeshpanchal27 @felixarntz should we also update the `AJAX_Runner` method to also be static for consistency? We would also need to update the `Plugin_Request_Utility::initialize_runner` method to use that static methods also.

jjgrainger (changes requested):

includes/Utilities/Plugin_Request_Utility.php
spacedmonkey: I did this here in my testing. https://github.com/10up/plugin-check/blob/e5f89c4f8b76bfeeebd408671cbdb9fd82b1950c/cli.php

spacedmonkey:
Keep the static add back the public on the method.

spacedmonkey:
Another lint error.

spacedmonkey (approved):

jjgrainger (approved):
Thanks @mukeshpanchal27 updates look good

felixarntz (approved):

PR #162, opened 2023-05-11
Make methods in abstract classes that should not be overridden `final`

### LINKED ISSUE #129: Make methods in abstract classes that should not be overridden `final`

The plugin checker contains a number of abstract classes that contributors should override. Several of these classes contain methods with an implementation that should not be overwritten.

We should lock down those parts of the codebase so that no intentional overrides are possible. If in doubt, make the method `final`. We can always remove it later if we decide it makes sense, but not vice-versa. Similarly, we should use the strictest feasible method visibility (e.g. `private` instead of `protected` where possible).

Here is a non-comprehensive list of what I propose:
* All `Abstract_Check_Runner` methods that aren't `abstract` and that are either `public` or `protected` should be made `final`. (Note that two of the `protected` methods should be made `private` as mentioned below, and those therefore don't need `final` on them.)
* `Abstract_PHP_CodeSniffer_Check::run()` should be made `final`.
* The following classes should all be made `final`:
    * `Checks`
    * `Check_Context`
    * `Check_Result`
    * `Runtime_Environment_Setup`
    * `Admin_Page`
    * `Admin_AJAX`
    * `Plugin_Check_Command`
* The following methods should all be made `private` instead of `protected`:
    * `Checks::run_check_with_result()`
    * `Abstract_Check_Runner::get_check_slugs()`
    * `Abstract_Check_Runner::has_runtime_check()`
    * All `protected` methods in `Admin_AJAX`
    * All `protected` methods in `Plugin_Check_Command`

### Designs

_No response_

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #129 

### Checklist:

### DIFF
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -19,7 +19,7 @@ use WP_Error;
  *
  * @since n.e.x.t
  */
-class Admin_AJAX {
+final class Admin_AJAX {
 
 	/**
 	 * Nonce key.
@@ -66,7 +66,7 @@ class Admin_AJAX {
 	 *
 	 * @since n.e.x.t
 	 */
-	final public function add_hooks() {
+	public function add_hooks() {
 		add_action( 'wp_ajax_' . self::ACTION_CLEAN_UP_ENVIRONMENT, array( $this, 'clean_up_environment' ) );
 		add_action( 'wp_ajax_' . self::ACTION_SET_UP_ENVIRONMENT, array( $this, 'set_up_environment' ) );
 		add_action( 'wp_ajax_' . self::ACTION_GET_CHECKS_TO_RUN, array( $this, 'get_checks_to_run' ) );
@@ -78,7 +78,7 @@ class Admin_AJAX {
 	 *
 	 * @since n.e.x.t
 	 */
-	final public function get_nonce() {
+	public function get_nonce() {
 		return wp_create_nonce( self::NONCE_KEY );
 	}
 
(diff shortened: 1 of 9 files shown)


spacedmonkey (approved):

includes/Admin/Admin_AJAX.php
```diff
@@ -66,7 +66,7 @@ class Admin_AJAX {
 	 *
 	 * @since n.e.x.t
 	 */
-	public function add_hooks() {
+	final public function add_hooks() {
```
felixarntz: Instead of making the methods in this class final, the entire class should be made final, see the issue description.

includes/Admin/Admin_Page.php
```diff
@@ -37,7 +37,7 @@ class Admin_Page {
 	 *
 	 * @param Admin_AJAX $admin_ajax Instance of Admin_AJAX.
 	 */
-	public function __construct( Admin_AJAX $admin_ajax ) {
+	final public function __construct( Admin_AJAX $admin_ajax ) {
```
felixarntz: Same as above applies to this class and a few others in this PR. Please refer to the issue description.

felixarntz (changes requested):
@mukeshpanchal27 Almost good to go, however note that for the classes that should be made final, the actual _classes_ should be made final, not all the methods in that class.

The result will be more or less the same, but it's a lot of unnecessary code to make all methods final when you can instead make the entire class final. And we don't want people to extend those classes for the time being.

jjgrainger (changes requested):
Thanks @mukeshpanchal27 however it does not follow whats outlines in [the issue](https://github.com/10up/plugin-check/issues/129). 

- Most of the classes outlined in the issue can be made `final` instead of setting it to each method.
- I looks like the `Check_Context` class update is missing

includes/Admin/Admin_AJAX.php
mukeshpanchal27: Thanks @felixarntz, In https://github.com/10up/plugin-check/pull/162/commits/89830ecf2c521b02d3f7e2a140aa3bc78937fcfc i address this feedback. Can you please take a look.

felixarntz (approved):

jjgrainger (approved):
Thanks @mukeshpanchal27 changes look good 👍

PR #163, opened 2023-05-12
Update WPCS to `dev-develop` branch at a specific commit

### LINKED ISSUE #136: Exception thrown by I18nSniff in PHP 8.1
While testing in PHP 8.1 all warnings from the `I18n_Usage_Check` are replaced by the following Exception thrown. It looks as though the issue is to do with WPCS dependency rather than the plugin check itself but we may want to consider how this is handled? 

image

```
An error occurred during processing; checking has been aborted. The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/wp-content/plugins/plugin-check/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WP/I18nSniff.php on line 194
```

mukeshpanchal27:
Closes #136 

### How to test the Change

1. Take the pull from the branch.
2. Set `phpVersion: 8.1` in your `.wp-env.json` file and restart environment.
4. Update composer. Use `composer update` command.
5. Go to Admin Panel > Tools > Plugin Check > Check plugin

### Checklist:

### DIFF
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -51,7 +51,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
       - name: Validate Composer configuration
-        run: composer validate --strict
+        run: composer validate
       - name: Composer Install
         run: composer install --no-interaction --no-progress
       - name: PHP Lint
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
   "require-dev": {
     "phpunit/phpunit": "^9.5",
     "squizlabs/php_codesniffer": "^3.7",
-    "wp-coding-standards/wpcs": "^2.3",
+    "wp-coding-standards/wpcs": "dev-develop#fca9d9e",
     "wp-phpunit/wp-phpunit": "^6.1",
     "yoast/phpunit-polyfills": "^1.0",
     "phpcompatibility/php-compatibility": "^9.3",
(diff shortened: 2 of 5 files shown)


mukeshpanchal27:
PHPCS shows `- The package "wp-coding-standards/wpcs" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.` error because we use unstable version.

swissspidy:
The warning is coming from `composer validate --strict` specifically, not PHPCS. It's usually just a warning, but `--strict` makes it error.

The only way to resolve this is by not using `--strict` for the time being.

Aside, [a note from the Composer docs](https://getcomposer.org/doc/04-schema.md#package-links) about using explicit commit references:

> This feature has severe technical limitations, as the `composer.json` metadata will still be read from the branch name you specify before the hash. You should therefore only use this as a temporary solution during development to remediate transient issues, until you can switch to tagged releases. The Composer team does not actively support this feature and will not accept bug reports related to it.

It's definitely a temporary solution in this case, so IMHO it should be fine to temporarily remove `--strict` as well.

spacedmonkey:
@mukeshpanchal27 The composer install is now failing in CI.

mukeshpanchal27:
@swissspidy, the CI still shows that warning and throws the error.

swissspidy:
The warning is expected to be there still. The step still fails now because of the additional `The lock file is not up to date with the latest changes in composer.json` error. That is new.

mukeshpanchal27:
Thanks, @spacedmonkey, for helping me correct the composer version issue.

spacedmonkey (approved):

swissspidy (approved):

felixarntz (approved):

PR #165, opened 2023-05-12
Make in min requirement PHP 7.2

spacedmonkey:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Changed - Existing functionality
> Deprecated - Soon-to-be removed feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @username, @username2, ...

### Checklist:

- [ ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests pass.

PR #166, opened 2023-05-12
Improve PHP cross-version support and testing

### LINKED ISSUE #164: Run PHPUnit unit tests again different versions of PHP

Run PHPUnit unit tests again different versions of PHP. We could also consider running at against different WordPress versions as well. 

Screenshot 2023-05-12 at 10 28 30

### Designs

_No response_

_No response_

### Code of Conduct

swissspidy:
### Description of the Change

- Fixes `concurrency` setting for GitHub Actions workflows
	- See https://docs.github.com/en/actions/using-jobs/using-concurrency
- Uses test matrix to test from PHP 5.6 upwards
- Moves dev dependencies requiring a higher PHP version to separate folder
	- Ensures that production dependencies can be installed and used on PHP 5.6+ without issues
- Move phpunit tests to subfolder
	- Otherwise phpunit loads phpstan files too
- Fixes errors reported by PHPCS about missing `use` statements

**To-do**

* [ ] Investigate failing test on WP 6.0

Closes #164

### Checklist:


spacedmonkey:
See https://github.com/10up/plugin-check/pull/165 for limiting php version 7.2+

tests/phpunit/Checker/Preparations/Force_Single_Plugin_Preparation_Tests.php
```diff
@@ -21,12 +21,12 @@ public function set_up() {
...
-	public function test_prepare_plugin_exists() {
+	public function test_prepare_plugin_does_not_exist() {
 
-		$preparation = new Force_Single_Plugin_Preparation( 'akismet/akismet.php' );
+		$preparation = new Force_Single_Plugin_Preparation( 'foo-plugin/foo-plugin.php' );
 
 		$this->expectException( 'Exception' );
-		$this->expectExceptionMessage( 'Invalid plugin akismet/akismet.php: Plugin file does not exist.' );
+		$this->expectExceptionMessage( 'Invalid plugin foo-plugin/foo-plugin.php: Plugin file does not exist.' );
 		$preparation->prepare();
```
swissspidy, line 30: This test was inaccurate because in some environments like wp-env Akismet _does_ exist, so there wouldn't be an exception.

It's more robust to use a slug that definitely does not exist.

jjgrainger (approved):
Thanks @swissspidy looks great, approved

phpstan.neon.dist
```diff
@@ -10,8 +10,6 @@ parameters:
     - tests/phpstan/stubs/wp-cli.php
     - tests/phpstan/stubs/exitexception.php
     - tests/phpstan/stubs/formatter.php
-    - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
```
spacedmonkey, line 13: The stub are not needed?

build-cs/.gitignore
```diff
@@ -0,0 +1 @@
+vendor/
```
spacedmonkey: Should this not be in the root gitgnore?

.github/workflows/php-test.yml
```diff
@@ -35,15 +35,45 @@ on:
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 jobs:
   php-test:
-    name: PHP
+    name: PHP ${{ matrix.php }} - WP ${{ matrix.wordpress }}${{ matrix.experimental && ' (experimental)' || '' }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+#      fail-fast: true
```
mukeshpanchal27:
```suggestion
```
If we not use `fail-fast` then remove it.

composer.json
```diff
@@ -5,32 +5,38 @@
...
-    "szepeviktor/phpstan-wordpress": "^1.1",
-    "phpstan/extension-installer": "^1.2",
-    "slevomat/coding-standard": "^8.9"
+    "yoast/phpunit-polyfills": "^1.0"
   },
   "scripts": {
-    "lint": "phpcs --standard=phpcs.xml.dist",
-    "format": "phpcbf --standard=phpcs.xml.dist",
+    "lint": [
+      "composer --working-dir=build-cs install",
```
spacedmonkey, line 18: Could we make this another script, maybe `pre-cs`

.github/workflows/php-test.yml
```diff
@@ -35,15 +35,45 @@ on:
...
+          - php: '8.2'
+            wordpress: 'trunk'
+            experimental: true
+    env:
+      WP_ENV_PHP_VERSION: ${{ matrix.php }}
+      WP_ENV_CORE: ${{ matrix.wordpress == 'trunk' && 'WordPress/WordPress' || format( 'https://wordpress.org/wordpress-{0}.zip', matrix.wordpress ) }}
     steps:
-      - uses: styfle/cancel-workflow-action@0.9.1
       - uses: actions/checkout@v3
       - uses: shivammathur/setup-php@v2
```
mukeshpanchal27:
```suggestion
      - uses: actions/checkout@v3

      - uses: shivammathur/setup-php@v2
```

Add line break between each steps as you did for `php-lint.yml`

spacedmonkey (commented):
This PR is nearly there, just a couple of questions and nitpics. After that, i would approve.

phpstan.neon.dist
swissspidy: The stubs are loaded automatically by the phpstan-wordpress extension. Adding this line here is superfluous.

build-cs/.gitignore
swissspidy: Both works, just a matter of preference I guess.

mukeshpanchal27 (commented):
Thanks, @swissspidy. It looks good to me. I have left some minor suggestions (nitpicks). I noticed there is one pending task mentioned in the PR description. Will it be addressed as part of this PR or in a follow-up PR?

build-cs/.gitignore
spacedmonkey: I would not think to look for this file. Let just have one file.

composer.json
swissspidy: That seems overkill do me, I don't see harm in this tiny bit of repetition. This way it's more connected with the next script and more readable, plus I don't see a reason why someone would want to run this command on its own.

.github/workflows/php-test.yml
swissspidy: This was just leftover from debugging. I need to uncomment it, not remove it. Good catch!

composer.json
```diff
@@ -5,32 +5,38 @@
   "license": "GPL-2.0-or-later",
   "require": {
     "php": ">=5.6",
-    "composer/installers": "^2.2"
+    "composer/installers": "^v1.12.0 || ^2.2",
+    "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
+    "squizlabs/php_codesniffer": "^3.7",
```
spacedmonkey: Could we remove this as it would be loaded with `wp-coding-standards/wpcs`?

spacedmonkey (approved):
Approved with some nicpics.

composer.json
swissspidy: Sure yeah that works.

I figured it's best to keep the explicit reference here like we do today, but right now the result is the same.

swissspidy:
> It looks good to me. I have left some minor suggestions (nitpicks). I noticed there is one pending task mentioned in the PR description. Will it be addressed as part of this PR or in a follow-up PR?

Great question! IMHO that's better addressed in a follow-up issue / PR because it requires some investigation. It could be a bad test or an actual bug in the plugin.

mukeshpanchal27 (approved):
Thanks, @swissspidy, for the update. The changes look good to me.

felixarntz (approved):
Awesome work @swissspidy!

build-cs/composer.json
```diff
@@ -0,0 +1,17 @@
+{
```
westonruter, line 1: Will Dependabot be aware of this package manifest?

swissspidy: No, it will need to be added to the Dependabot config. But this project does not appear to be using Dependabot.

westonruter: Not explicitly, but Dependabot still [sends security updates](https://github.com/10up/plugin-check/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies).

mukeshpanchal27:
> > It looks good to me. I have left some minor suggestions (nitpicks). I noticed there is one pending task mentioned in the PR description. Will it be addressed as part of this PR or in a follow-up PR?
> 
> Great question! IMHO that's better addressed in a follow-up issue / PR because it requires some investigation. It could be a bad test or an actual bug in the plugin.

@swissspidy, can you please open a follow-up issue to ensure it doesn't get missed?

build-cs/composer.json
swissspidy: Opened #168

swissspidy:
Opened #169

PR #167, opened 2023-05-12
Remove constructor parameter from `Checks` class

### LINKED ISSUE #156: Remove constructor parameter from `Checks` class

Based on https://github.com/10up/plugin-check/issues/127#issuecomment-1499122587:
* The `$plugin_main_file` parameter should be removed from the `Checks` class constructor, since it is only used in the `run_checks()` method, which is conditionally called much later.
* The `Checks::context()` method should therefore also be removed.
* The `Checks::run_checks()` method should have its signature changed to `run_checks( Check_Context $context, array $checks )`, so that the `Check_Context` for the plugin to check only needs to be passed there.
* With that, we can also remove the `Abstract_Check_Runner::get_checks_instance()` method and underlying `Abstract_Check_Runner::$checks` property which has been confusing in terms of naming and now doesn't really serve a purpose anymore, since you can just write `new Checks()` and be done with it.
* Instead, we can add a `Abstract_Check_Runner::get_check_context()` method which returns `new Check_Context( WP_PLUGIN_DIR . '/' . $this->get_plugin_basename() )`.
* The `Abstract_Check_Runner::prepare()` method that uses the (now removed) `Checks::context()` method can then instead simply use the new `get_check_context()` method.

### Designs

_No response_

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #156 

In this PR, I have not removed the get_checks_instance method because the new Checks class creates a new instance for each request, and removing the method would cause the check to not work.

### Checklist:


includes/Checker/Checks.php
```diff
@@ -25,47 +25,25 @@ class Checks {
...
-	 * Returns the Check Context.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return Check_Context The plugin context that is being checked.
-	 */
-	public function context() {
-		return $this->check_context;
-	}
+	public function __construct() { }
```
spacedmonkey: This line can be removed.

mukeshpanchal27: Thanks @spacedmonkey, Address in https://github.com/10up/plugin-check/pull/167/commits/efa0818f9985e2f8d1fde088d1ec576f63fc24be

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -346,4 +346,14 @@ public function get_plugin_basename() {
...
 		return $this->plugin_basename;
 	}
+
+	/** Gets the Check_Context for the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return Check_Context The check context for the plugin file.
+	 */
+	protected function get_check_context() {
```
felixarntz: Better to make this private since it's not needed in child classes (related to #162).

felixarntz (commented):
@mukeshpanchal27 Almost LGTM, just one small thing to enhance. Normally not a blocker, but since we're fixing this via #162, we should also stick to that pattern here, to use the lowest visibility possible for each method.

tests/Checker/Checks_Tests.php
```diff
@@ -15,7 +16,8 @@ class Checks_Tests extends WP_UnitTestCase {
 	public function set_up() {
 		parent::set_up();
 
-		$this->checks = new Checks( 'test-plugin/test-plugin.php' );
+		$this->checks  = new Checks( 'test-plugin/test-plugin.php' );
```
jjgrainger: @mukeshpanchal27 we can remove the constructor parameter here

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -311,7 +311,7 @@ protected function get_checks_instance() {
 		}
 
 		$plugin_basename = $this->get_plugin_basename();
```
jjgrainger: @mukeshpanchal27 we can remove this line also as we no longer need the plugin basename here

jjgrainger (changes requested):
Thanks @mukeshpanchal27 looking good, just a few minor changes to be made

felixarntz (approved):
@mukeshpanchal27 Approved from my end, pending the changes from @jjgrainger.

tests/Checker/Checks_Tests.php
mukeshpanchal27: Good catch. Addressed in https://github.com/10up/plugin-check/pull/167/commits/ab432f0c9e691f45e9800dc7014da8f948b09942

jjgrainger (approved):

PR #170, opened 2023-05-17
Decouple available checks

### LINKED ISSUE #127: Decouple available checks from rest of the check running infrastructure

A few problems that can be addressed by decoupling the functionality around available checks:
* It is unclear whether checks should be referenced through their object or their slug.
* Matching checks from given slugs is "randomly" handled in `Abstract_Check_Runner` instead of a dedicated place for this.
* All checks are put together, even though we fundamentally differentiate between static and runtime checks.

Proposed solution:
* Introduce an interface `Check_Repository` with methods:
    * bitwise flags as `const` values, for now only for static and runtime check types (later this can get additional flags, e.g. for stability values to filter by):
        * `const TYPE_STATIC = 1`
        * `const TYPE_RUNTIME = 2`
        * `const TYPE_ALL = 3` (same as `TYPE_STATIC | TYPE_RUNTIME`)
    * `public function register_check( string $slug, Check $check )`: Registers a check instance under a given slug.
    * `public function get_checks( int $flags = self::TYPE_ALL, array $check_slugs = array() ) : array`: Returns an array of just check objects based on the given slugs and flags (indexed array, not a map!).
* Introduce a class `Default_Check_Repository` that implements `Check_Repository` that implements the above methods:
    * The registration method should throw an exception if a check slug is already used.
    * The registration method should throw an exception if the check class implements neither `Static_Check` or `Runtime_Check`.
    * The registration method should store `Static_Check` and `Runtime_Check` instances in separate arrays internally.
    * The get checks method should return checks based on the flags. If only static checks or only runtime checks are requested, it should take the values from the respective array, otherwise it should combine both arrays.
    * The get checks method should, if a non-empty `$check_slugs` parameter is passed, filter down the list of checks based on the flags by the check slugs passed. If any of the check slugs is not present, the method should throw an exception.
    * Examples: The method should then be callable e.g. like `get_checks( Check_Repository::TYPE_STATIC | Check_Repository::TYPE_RUNTIME )` (same as `get_checks( Check_Repository::TYPE_ALL )`), or `get_checks( Check_Repository::TYPE_STATIC )` to only get static checks.
* Create a private method `Abstract_Check_Runner::register_checks()`, which should register all checks on an instance of `Default_Check_Repository`.
* In `Abstract_Check_Runner::__construct()`, call the private `register_checks()` method so that checks are registered as soon as the runner is instantiated.
* The `Abstract_Check_Runner::get_checks_to_run()` method should be updated to use the new repositories:
    * Get the passed check slugs.
    * Call the `get_checks()` method of the `Default_Check_Repository` instance, passing the check slugs and:
        * If the requirements for running runtime checks are met, with flag `TYPE_ALL`
        * Otherwise, with flag `TYPE_STATIC`
        * This method will now throw the exception as needed, so it no longer needs to be implemented here
* Remove the `Abstract_Check_Runner::has_runtime_check()` method and inline that logic instead into `Abstract_Check_Runner::prepare()` which should now be the only place this code is needed in that class. We shouldn't have the dedicated method anymore to not encourage its use. Alternatively, we could make it `private` so that it cannot be used anywhere outside the class.
* Remove the `Checks::get_checks()` method and no longer call it in the `Checks::run_checks` method. Since validation now happens elsewhere, that method should simply trust any check instances passed. Registration of the checks now happens in the new `Abstract_Check_Runner::register_checks()` method instead of here via a filter.

### Designs

_No response_

_No response_

### Code of Conduct

jjgrainger:
Addresses the acceptance criteria in #127

Fixes #127


includes/Checker/Default_Check_Repository.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param Check  $check The Check instance.
+	 *
+	 * @throws Exception Thrown if Check does not use correct interface, or slug already exists.
+	 */
+	public function register_check( $slug, Check $check ) {
+		if ( ! $check instanceof Runtime_Check && ! $check instanceof Static_Check ) {
+			throw new Exception( __( 'Check must be an instance of Runtime_Check or Static_Check.', 'plugin-check' ) );
+		}
+
+		if ( array_key_exists( $slug, $this->runtime_checks ) || array_key_exists( $slug, $this->static_checks ) ) {
```
felixarntz: Nit-pick: `isset()` is faster than `array_key_exists()`, and since the entries would never be `null`, there's no functional value in using `array_key_exists()` either.

includes/Checker/Default_Check_Repository.php
```diff
@@ -0,0 +1,111 @@
...
+			self::TYPE_ALL     => array_merge( $this->static_checks, $this->runtime_checks ),
+			self::TYPE_STATIC  => $this->static_checks,
+			self::TYPE_RUNTIME => $this->runtime_checks,
+		);
+
+		if ( ! isset( $check_arrays[ $flags ] ) ) {
+			throw new Exception( __( 'Invalid check flags passed.', 'plugin-check' ) );
+		}
+
+		$checks = $check_arrays[ $flags ];
```
felixarntz: We can solve this in a simpler way by relying on the bitwise operators, which is a benefit of using such flags. We can also avoid `array_merge()` in favor of the `+` operator, which is faster, and it is ensured there are no duplicate slugs between the two arrays anyway.

```suggestion
		$checks = array();
		if ( $flags & self::TYPE_STATIC ) {
			$checks += $this->static_checks;
		}
		if ( $flags & self::TYPE_RUNTIME ) {
			$checks += $this->runtime_checks;
		}
```

I think throwing the exception here isn't strictly necessary, since the logic works regardless. Passing flag values that aren't supported (i.e. don't mean anything) doesn't result in exceptions in PHP itself either.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -255,45 +264,18 @@ private function get_shared_preparations( array $checks ) {
...
-			if ( ! $plugin_active || ! $this->initialized_early ) {
-				$checks_to_run = array_filter(
-					$checks_to_run,
-					function ( $check ) {
-						return ! $check instanceof Runtime_Check;
-					}
-				);
-			}
+		// Check if conditions are met in order to perform Runtime Checks.
+		if ( $this->initialized_early ) {
```
felixarntz: I think you missed the check for whether the plugin is active here, which is now missing.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -58,6 +58,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	protected $plugin_basename;
 
+	/**
+	 * An instance of the Check_Repository.
+	 *
+	 * @since n.e.x.t
+	 * @var Check_Repository
+	 */
+	protected $check_repository;
```
felixarntz: Please keep in mind to use the lowest visibility feasible, i.e. this should be `private`.

includes/Checker/AJAX_Runner.php
```diff
@@ -70,6 +70,9 @@ protected function get_plugin_param() {
 	 * @return array An array of Check slugs to run.
 	 */
 	protected function get_check_slugs_param() {
-		return filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
+		$checks = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
+		$checks = is_null( $checks ) ? array() : $checks;
```
felixarntz, line 74: Good catch!

tests/phpunit/Checker/Default_Check_Repository_Tests.php
```diff
@@ -0,0 +1,104 @@
...
+
+		$this->repository->register_check( 'empty_check', new Invalid_Check() );
+	}
+
+	public function test_register_exception_thrown_for_existing_check_slug() {
+		$this->expectException( 'Exception' );
+		$this->expectExceptionMessage( 'Check slug "static_check" is already in use.' );
+
+		$this->repository->register_check( 'static_check', new Static_Check() );
+		$this->repository->register_check( 'static_check', new Static_Check() );
```
felixarntz, line 48: Nit-pick: Maybe use a `Runtime_Check` in one of the two, to ensure the conflicts are also checked against the list of the respective other type? Or add a separate unit test for it, either works.

tests/phpunit/Checker/Default_Check_Repository_Tests.php
```diff
@@ -0,0 +1,104 @@
...
+
+		$this->assertSame( array( $static_check ), $this->repository->get_checks( Check_Repository::TYPE_ALL, array( 'static_check' ) ) );
+	}
+
+	public function test_get_checks_throws_exception_for_invalid_flag() {
+		$this->expectException( 'Exception' );
+		$this->expectExceptionMessage( 'Invalid check flags passed.' );
+
+		$this->repository->get_checks( 5 );
+	}
```
felixarntz: See my comment above, this is probably not necessary.

felixarntz (commented):
@jjgrainger This looks great so far, I left a few notes of feedback below.

includes/Checker/Check_Repository.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Bitwise flag for runtime type checks.
+	 *
+	 * @since n.e.x.t
+	 * @var int
+	 */
+	const TYPE_RUNTIME = 2;
+
+	/**
+	 * Bitwise flag for all check types.
```
felixarntz, line 34: Maybe add a comment to clarify why this value is `3`.
```suggestion
	 * Bitwise flag for all check types.
	 *
	 * This is the same as `TYPE_STATIC | TYPE_RUNTIME`.
```

includes/Checker/Default_Check_Repository.php
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * Array map holding all runtime checks.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	protected $runtime_checks = array();
+
+	/**
+	 * Array map holding all runtime checks.
```
mukeshpanchal27:
```suggestion
	 * Array map holding all static checks.
```

includes/Checker/Check_Repository.php
```diff
@@ -0,0 +1,62 @@
...
+
+	/**
+	 * Returns an array of checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int   $flags       The check type flag.
+	 * @param array $check_slugs An array of check slugs to return.
+	 *
+	 * @return array An indexed array of check instances.
```
mukeshpanchal27:
```suggestion
	 * @param array $check_slugs An array of check slugs to return.
	 * @return array An indexed array of check instances.
```

includes/Checker/Default_Check_Repository.php
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Returns an array of checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int   $flags       The check type flag.
+	 * @param array $check_slugs An array of check slugs to return.
+	 *
+	 * @return array An indexed array of check instances.
```
mukeshpanchal27:
```suggestion
	 * @param array $check_slugs An array of check slugs to return.
	 * @return array An indexed array of check instances.
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -355,4 +337,25 @@ final public function get_plugin_basename() {
...
+
+	/**
+	 * Registers Checks to the Check_Repository.
+	 *
+	 * @since n.e.x.t
+	 */
+	private function register_checks() {
+		$this->check_repository = new Default_Check_Repository();
+
+		$checks = apply_filters(
```
mukeshpanchal27, line 359: Document is missing for filter.

mukeshpanchal27 (commented):
Thanks @jjgrainger, Some quick feedback.

includes/Checker/Check_Repository.php
jjgrainger: Updated

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks, updated

tests/phpunit/Checker/Default_Check_Repository_Tests.php
jjgrainger: I've included a new test for runtime checks

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -47,8 +47,6 @@ class Enqueued_Scripts_Size_Check extends Abstract_Runtime_Check implements With
 	 */
 	public function __construct( $threshold_size = 300000 ) {
 		$this->threshold_size = $threshold_size;
-
-		$this->viewable_post_types = array_filter( get_post_types(), 'is_post_type_viewable' );
```
felixarntz, line 51: Good catch, looks like setting this here may indeed be too early.

tests/phpunit/Checker/Default_Check_Repository_Tests.php
felixarntz: @jjgrainger My point here was not to have another check for `Runtime_Check` checks specifically, but rather to ensure the slug check works also _between_ the two types of checks. E.g. if you first register a static check with a slug `my_slug`, and then register a runtime check with slug `my_slug`, it should still throw the exception, even though the checks are of a different type.

felixarntz (approved):
@jjgrainger Thanks, LGTM!

One small follow up for something to potentially cover in the tests, but not all that critical.

includes/Checker/Check_Repository.php
```diff
@@ -0,0 +1,63 @@
...
+	public function register_check( $slug, Check $check );
+
+	/**
+	 * Returns an array of checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int   $flags       The check type flag.
+	 * @param array $check_slugs An array of check slugs to return.
+	 * @return array An indexed array of check instances.
```
mukeshpanchal27, line 60:
```suggestion
```
Do we need the return here?

jjgrainger: @mukeshpanchal27 we have returns on other interface methods, for example in the `Preparation` interface.

tests/phpunit/Checker/Default_Check_Repository_Tests.php
jjgrainger: @felixarntz Ah, I understand now. I've added a new test to capture this too.

includes/Checker/Default_Check_Repository.php
```diff
@@ -0,0 +1,108 @@
...
+						);
+					}
+
+					return $checks[ $slug ];
+				},
+				$check_slugs
+			);
+		}
+
+		return array_values( $checks );
```
mukeshpanchal27:
```suggestion
		return $checks;
```
This change will solve the admin checks issue

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -255,45 +264,18 @@ private function get_shared_preparations( array $checks ) {
...
-			if ( ! $plugin_active || ! $this->initialized_early ) {
-				$checks_to_run = array_filter(
-					$checks_to_run,
-					function ( $check ) {
-						return ! $check instanceof Runtime_Check;
-					}
-				);
-			}
+		// Check if conditions are met in order to perform Runtime Checks.
+		if ( $this->initialized_early && is_plugin_active( $this->get_plugin_basename() ) ) {
```
mukeshpanchal27, line 277: The CLI check shows error for `is_plugin_active`

felixarntz (approved):
Thanks for the updates @jjgrainger, LGTM!

mukeshpanchal27 (approved):
Thanks @jjgrainger, The changes resolve the issues i have raised.

PR #173, opened 2023-05-23
Fix Runtime Checks in AJAX

### LINKED ISSUE #132: Runtime Checks do not run with Admin AJAX
During QA I was unable to see/run the Runtime check within the Admin page.

Looking into this, the [`get_checks_to_run`](https://github.com/10up/plugin-check/blob/trunk/includes/Admin/Admin_AJAX.php#L153) endpoint uses the `AJAX_Runner` to determine the checks to run. However, the checks returned take into consideration if the runner was initialized early.

As the runtime environment is not setup until the following `set_up_runtime_environment` AJAX call, the checks returned by [$runner->get_checks_to_run()](https://github.com/10up/plugin-check/blob/trunk/includes/Admin/Admin_AJAX.php#L183) will only ever be static checks.

In this specific instance, we need return checks based on the request ignoring the rest of the check running logic that will happen in future requests.

We should look into decoupling returning available checks from the rest of the infrastructure. This is related to #127

jjgrainger:
Closes #132


mukeshpanchal27:
@jjgrainger It seems it was issue of permission in GH CI. If we add below code in `bootstrap.php` then it resolve the error.

```
if ( ! defined( 'FS_METHOD' ) ) {
	define( 'FS_METHOD', 'direct' );
}
```

jjgrainger:
Thanks @mukeshpanchal27 I believe I was able to resolve this by using the mock file system on the failing tests. This is ready for review.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
...
+	 * @return bool Returns true if the runtime environment can be created, false if not.
+	 */
+	public function can_setup_runtime_environment() {
+		global $wp_filesystem;
+
+		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+
+		if ( ! $wp_filesystem && ! WP_Filesystem() ) {
+			return false;
+		}
```
mukeshpanchal27: Instead of checking this can we check something below?

```
if ( $wp_filesystem || WP_Filesystem() ) {
        // Check if the object-cache.php file exists.
	if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
	.....
} else {
	return false;
}
```

mukeshpanchal27 (approved):
Great work, @jjgrainger! The changes look good to me. Just had one thought left.

tests/phpunit/utils/Traits/With_Mock_Filesystem.php
```diff
@@ -0,0 +1,37 @@
...
+	 *
+	 * @since n.e.x.t
+	 */
+	protected function set_up_mock_filesystem() {
+		global $wp_filesystem;
+
+		add_filter(
+			'filesystem_method_file',
+			function() {
+				return __DIR__ . '/../testdata/Filesystem/WP_Filesystem_MockFilesystem.php';
```
spacedmonkey:
```suggestion
				return TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/Filesystem/WP_Filesystem_MockFilesystem.php';
```

spacedmonkey: I don't like path like this.

includes/Checker/Runtime_Environment_Setup.php
spacedmonkey: I don't like the above suggestion.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
...
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns true if the runtime environment can be created, false if not.
+	 */
+	public function can_setup_runtime_environment() {
+		global $wp_filesystem;
+
+		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+
+		if ( ! $wp_filesystem && ! WP_Filesystem() ) {
```
spacedmonkey:
```suggestion
		if ( ! is_object( $wp_filesystem ) && ! WP_Filesystem() ) {
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -270,11 +270,12 @@ final public function get_checks_to_run() {
 		// Include file to use is_plugin_active() in CLI context.
 		require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
+		$runtime     = new Runtime_Environment_Setup();
```
spacedmonkey: This breaks the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). In short, you should not hard code a class instantiation. This should be injected into the class. This helps mocking in tests. We do this in other places in the codebase.  Are we following the pattern of dependancy injection model in this plugin @felixarntz ?

includes/Checker/Runtime_Environment_Setup.php
spacedmonkey: Should this throw an Exception?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -270,11 +270,12 @@ final public function get_checks_to_run() {
...
 		$check_slugs = $this->get_check_slugs();
 		$check_flags = Check_Repository::TYPE_STATIC;
 
 		// Check if conditions are met in order to perform Runtime Checks.
-		if ( $this->initialized_early && is_plugin_active( $this->get_plugin_basename() ) ) {
+		if ( $runtime->can_setup_runtime_environment() && is_plugin_active( $this->get_plugin_basename() ) ) {
```
felixarntz: I think this needs to check if _either_ the runner was initialized early _or_ the runtime environment can be set up. If it was initialized early, we know the runtime environment is already set up, so no need to check for it.
```suggestion
		if ( ( $this->initialized_early || $runtime->can_setup_runtime_environment() ) && is_plugin_active( $this->get_plugin_basename() ) ) {
```

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
...
 	}
+
+	/**
+	 * Checks if the Runtime Environment can be created.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns true if the runtime environment can be created, false if not.
+	 */
+	public function can_setup_runtime_environment() {
```
felixarntz: Including "runtime environment" in the method name seems unnecessary? Why not just `can_setup()`?

On another (nit-pick) note, the methods in this class should better be named `set_up()`, `clean_up()`, and `can_set_up()`, given that those are verbs, while "setup" and "cleanup" are nouns.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
 			}
 		}
 	}
+
+	/**
+	 * Checks if the Runtime Environment can be created.
```
felixarntz: This description sounds confusing, it doesn't "create" a runtime environment. I'd suggest you instead reuse the same wording you have on the `setup()` method:
```suggestion
	 * Checks if the WordPress environment can be set up for runtime checks.
```

includes/Checker/Runtime_Environment_Setup.php
felixarntz: I'd say no, given this method is just to return whether it's possible anyway, not doing anything. So I think returning `false` here makes sense.

includes/Checker/Runtime_Environment_Setup.php
felixarntz: Agreed, early returns are preferable over nesting which leads to heavily indented / harder-to-read code.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
...
+				$object_cache_copy = WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'object-cache.copy.php';
+			}
+
+			// Check the drop-in file matches the plugin check version.
+			$original_content = $wp_filesystem->get_contents( WP_CONTENT_DIR . '/object-cache.php' );
+			$copy_content     = $wp_filesystem->get_contents( $object_cache_copy );
+
+			if ( $original_content && $original_content !== $copy_content ) {
+				return false;
+			}
```
felixarntz: Instead of checking the exact contents of the file, I think we should simplify this to just check whether the constant that _our_ `object-cache.php` drop-in sets is defined. If not, the existing file is not ours, so we should return `false`. Otherwise, we can return `true` right here as our file would be already placed.

I acknowledge that's not as precise and strict as comparing the exact file contents, but the latter can also be overly limiting, for example if someone wants to use our plugin checker but also use their actual object cache, they could implement a way so that both works in some custom approach, e.g. a custom `object-cache.php` drop-in that requires two other files that could be the drop-in.

TLDR: If our file already exists _at this point_ here, the constant should be defined, so we can use that to check if it's our or another drop-in.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,42 @@ public function cleanup() {
...
+		}
+
+		return true;
```
felixarntz: Something that is missing from this method is a check whether we can actually place our `object-cache.php` drop-in (only relevant if this drop-in does not exist yet, which you covered above).

To reliably detect that, we should do the following:
1. Run the code to place our file, basically line 58 (the `copy()` call) in the `setup()` method.
2. Check afterwards that the file exists. If not, return `false`.
    * In theory we should be able to rely on the return value from the `copy()` call, but that can be unreliable depending on the environment. So a separate exists check is more accurate for whether the file is placed successfully or not.
3. If it does exist, we can place it. Delete it again immediately, and then return `true`.

tests/phpunit/Checker/AJAX_Runner_Tests.php
```diff
@@ -10,9 +10,16 @@
...
 use WordPress\Plugin_Check\Test_Data\Runtime_Check;
+use WordPress\Plugin_Check\Test_Utils\Traits\With_Mock_Filesystem;
 
 class AJAX_Runner_Tests extends WP_UnitTestCase {
 
+	use With_Mock_Filesystem;
+
+	public function set_up() {
+		$this->set_up_mock_filesystem();
+	}
```
felixarntz, line 22: What is the rationale for setting that up here unrelated to any actual test changes in this class? Please clarify, but also add a code comment so that it remains clear for people that look at this code in the future.

tests/phpunit/Checker/CLI_Runner_Tests.php
```diff
@@ -10,9 +10,16 @@
...
 use WordPress\Plugin_Check\Test_Data\Runtime_Check;
+use WordPress\Plugin_Check\Test_Utils\Traits\With_Mock_Filesystem;
 
 class CLI_Runner_Tests extends WP_UnitTestCase {
 
+	use With_Mock_Filesystem;
+
+	public function set_up() {
+		$this->set_up_mock_filesystem();
+	}
```
felixarntz, line 22: Same as above.

tests/phpunit/Checker/Runtime_Environment_Setup_Tests.php
```diff
@@ -74,25 +77,25 @@ public function test_cleanup_with_existing_object_cache() {
...
+		$this->set_up_mock_filesystem();
+
+		// Simulate a different object-cache.php.
+		$dummy_file_content = '<?php /* Empty object-cache.php drop-in file. */';
+		$wp_filesystem->put_contents( WP_CONTENT_DIR . '/object-cache.php', $dummy_file_content );
+
+		$runtime_setup = new Runtime_Environment_Setup();
+
+		$this->assertFalse( $runtime_setup->can_setup_runtime_environment() );
 	}
```
felixarntz, line 110: See my comment above on checking whether we can actually place our file. It would be good to add another test case to cover that part which needs to be added to `can_setup()` / `can_set_up()`.

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
```diff
@@ -75,19 +78,21 @@ public function test_initialize_runner_with_ajax() {
 	public function test_destroy_runner_with_cli() {
 		global $wpdb, $table_prefix, $wp_actions;
 
+		$this->set_up_mock_filesystem();
```
felixarntz, line 81: Similar to above, why does this need to be called here?

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
```diff
@@ -120,6 +125,8 @@ function( $checks ) {
 	public function test_destroy_runner_with_ajax() {
 		global $wpdb, $table_prefix, $wp_actions;
 
+		$this->set_up_mock_filesystem();
```
felixarntz, line 128: Same as above.

felixarntz (changes requested):
@jjgrainger This mostly looks solid, I have a few mostly small notes to address.

The most significant one is that we also need to check whether we can actually place the drop-in file - WordPress servers may often be locked down for security reasons and not allow plugins to write to `wp-content`, and if so, we should catch that.

tests/phpunit/Checker/AJAX_Runner_Tests.php
jjgrainger: @felixarntz the changes in the PR now mean we use the file system to determine if we can return runtime checks. After the update we were experiencing errors in some of the tests in the [GitHub actions](https://github.com/10up/plugin-check/actions/runs/5058336367/jobs/9078334371).

This uses the mock file system in those tests when the runner is prepared and fixes the issues in the GitHub actions.

tests/phpunit/Checker/CLI_Runner_Tests.php
jjgrainger: See my comment on the above - https://github.com/10up/plugin-check/pull/173#discussion_r1206553708

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
jjgrainger: @felixarntz we prepare the runner in these tests, so we use the mock filesystem to prevent errors in the GitHub action (https://github.com/10up/plugin-check/pull/173#discussion_r1206553708)

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
jjgrainger: See my comment above - https://github.com/10up/plugin-check/pull/173#discussion_r1206557599

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Marking this as resolved

tests/phpunit/utils/Traits/With_Mock_Filesystem.php
jjgrainger: Thanks @spacedmonkey this has been updated

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: I've updated the PR with the suggestion but agree we shouldn't throw an exception here

includes/Checker/Abstract_Check_Runner.php
jjgrainger: We do instantiate other classes in this class, but agree this does not align with how we approach this in other classes. I'm interested to hear @felixarntz thoughts on this for this case.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz updated

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Thanks @felixarntz agreed. I've updated this and other methods as per your suggestion.

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Thanks @felixarntz updated.

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Thanks @felixarntz I've made this update.

includes/Checker/Runtime_Environment_Setup.php
jjgrainger: Thanks @felixarntz I've made this update and added a new test.

tests/phpunit/Checker/Runtime_Environment_Setup_Tests.php
jjgrainger: Thanks @felixarntz I made the updates alongside the additional test

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,49 @@ public function cleanup() {
...
+		if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+			// Check If the object-cache.php file is the Plugin Check version.
+			if ( defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) && WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION ) {
+				return true;
+			}
+		} else {
+			// Get the correct Plugin Check directory when run too early.
+			if ( ! defined( 'WP_PLUGIN_CHECK_PLUGIN_DIR_PATH' ) ) {
+				$plugins_dir       = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : WP_CONTENT_DIR . '/plugins';
+				$object_cache_copy = $plugins_dir . '/plugin-check/object-cache.copy.php';
```
felixarntz: Like in a few other places before, keep in mind that `plugin-check` is not guaranteed to be the folder name. Since we're only dealing with paths here, I'd say a simpler and more reliable solution would be to use `dirname( dirname( __DIR__ ) ) . '/plugin-check/object-cache.copy.php'`.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -100,4 +100,49 @@ public function cleanup() {
...
+			} else {
+				$object_cache_copy = WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'object-cache.copy.php';
+			}
+
+			// If the file does not exist, check if we can place it.
+			$wp_filesystem->copy( $object_cache_copy, WP_CONTENT_DIR . '/object-cache.php' );
+
+			if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) { // @phpstan-ignore-line
```
felixarntz: Curious, what did PHPStan complain about with this line? Can we exclude the line only for that specific complaint rather than generally? Not familiar, but at least with PHPCS such specific exclusions are possible.

includes/Checker/Abstract_Check_Runner.php
felixarntz: Makes sense to allow setting this instance instead of hard-coding it, though we also should keep that optional, so that we don't have to set it everywhere where a runner is instantiated. I also wouldn't want to add it as a constructor parameter, since that would mean we have to maintain this forever or deprecate a parameter, which you can never truly replace later.

I would suggest we go with:
* A `private` class property for the `Runtime_Environment_Setup` instance
* Instantiate that property in the constructor with our regular instance
* Add a `final public set_runtime_environment_setup( $runtime_environment_setup )` method that e.g. a test can optionally call to overwrite it.

felixarntz (commented):
@jjgrainger Thank you for the updates! This is almost good to go, just a few tiny last points.

includes/Checker/Runtime_Environment_Setup.php
mukeshpanchal27: @felixarntz PHPStan shows an `If condition is always false` error, and they don't have a specific rule to exclude this complaint. Someone has already opened an issue upstream at https://github.com/phpstan/phpstan/issues/7491.

Until the next release, ignoring the line seems like a better solution in my opinion. cc. @jjgrainger

includes/Checker/Runtime_Environment_Setup.php
mukeshpanchal27: Thanks! Updated in https://github.com/10up/plugin-check/pull/173/commits/8717db9cd7e7c87099df67324a70187716cd52ea

includes/Checker/Abstract_Check_Runner.php
mukeshpanchal27: The suggestion is implemented in https://github.com/10up/plugin-check/pull/173/commits/8717db9cd7e7c87099df67324a70187716cd52ea

includes/Checker/Runtime_Environment_Setup.php
mukeshpanchal27: Another option we can remove in `phpstan.neon` like https://gitlab.com/foodsharing-dev/foodsharing/-/blob/master/phpstan.neon#L19

```
-
  message: '#If condition is always true#'
  paths:
     - includes/Checker/Runtime_Environment_Setup.php

```

jjgrainger: @felixarntz @mukeshpanchal27 is right, PHPStan raised an issue because we are checking the file exists twice. We do this because we place the file to test we can write it. I believe there is no way to ignore a specific complaint with PHPStan except for what Mukesh has highlighted above.

mukeshpanchal27 (approved):

includes/Checker/Runtime_Environment_Setup.php
felixarntz: @jjgrainger Okay, then I'd say let's keep the ignore statement as is, but please add a comment (for human readers of the code explaining what is being ignored there and why.

felixarntz: Could you please add a comment here to clarify the reason for the ignore statement for people reviewing this code in the future? See https://github.com/10up/plugin-check/pull/173/files#r1210423380.

felixarntz (approved):
@jjgrainger @mukeshpanchal27 LGTM, though it would be great if you could address the point below before merging.

includes/Checker/Runtime_Environment_Setup.php
joemcgill: FWIW, `WP_Filesystem_Base::copy` returns `false` if it fails, so we could use that value and not do an additional check to see if the placed file exists. See: https://developer.wordpress.org/reference/classes/wp_filesystem_base/copy/

felixarntz: @joemcgill I'm aware of that, though there are some odd quirks about how certain hosts may have their filesystem access configured, as well as how filesystem implementation classes work. For example, some may just call a low-level function without return value and return `true`. Relying on a separate check whether the file actually exists after placing it is therefore safer than to check a return value.

mukeshpanchal27: PHPStan ignore reason added in https://github.com/10up/plugin-check/pull/173/commits/cade428e4030667dcb8a4fd72265e304d07429c3

jjgrainger:
Thanks @mukeshpanchal27 updates look good, merging

PR #175, opened 2023-05-23
Implement `Abstract_File_Check` base class as well as two plugin requirement checks using it

### LINKED ISSUE #174: Implement foundation for static checks that check files with custom code rather than PHPCodeSniffer

The [WordPress theme checker](https://github.com/WordPress/theme-check) has existed for several years and contains several checks that would also be useful for plugins. Those checks are implemented using custom logic that iterates through files and their contents and searches for certain substrings or regexes to identify whether they violate a certain requirement. The recent [plugin check tool](https://github.com/WordPress/plugin-check) started by the plugin review team also almost exclusively contains checks using that approach (partially as its original code was inspired by the theme checker).

While using PHPCodeSniffer is a great solution for static checks, it shouldn't be the only way to implement static checks. While PHPCS comes with a large ecosystem where lots of useful sniffs are already available, writing custom sniffs requires specialized knowledge on the internals of PHPCS, and for certain simpler checks it can therefore be more straightforward to write custom static checks like those in the aforementioned tools.

The plugin checker should provide an easy-to-use foundation to write such static checks.

### Designs

An abstract class for the manual kind of static checks should be implemented, inspired by the approach taken by the aforementioned tools. The class should:
* provide infrastructure to run a check over the list of files from the given plugin
* provide methods to filter the list of files, e.g. by certain extensions or a regular expression
* provide methods to scan the file contents for certain substrings or regular expressions

Much of this infrastructure should be implemented as static methods, given that it is reusable logic that doesn't need to be tied to a specific check's instance. This is particularly important when it comes to the methods to get the list of files and file contents, to ensure those can be cached across all checks using this new abstract class, rather than being cached only per individual instance, which would make it almost useless, as then it wouldn't be cached between multiple checks within the same request.

In order to establish the usefulness of the new class, the PR should come with 1-3 simple checks that are applicable to use the new class and are essential for any plugins to pass, e.g. https://github.com/WordPress/plugin-check/blob/trunk/checks/code-obfuscation.php, https://github.com/WordPress/plugin-check/blob/trunk/checks/file-checks.php or https://github.com/WordPress/theme-check/blob/master/checks/class-php-short-tags-check.php.

_No response_

### Code of Conduct

felixarntz:
### Description of the Change

* Implements an `Abstract_File_Check` base class to make it easy to implement checks like those used by the WordPress theme check plugin or by the early plugin review tool, as described in #174
* Implement two essential checks for minimum plugin requirements that use the new base class:
    * `Code_Obfuscation_Check`, based on https://github.com/WordPress/plugin-check/blob/trunk/checks/code-obfuscation.php
    * `File_Type_Check`, based on https://github.com/WordPress/plugin-check/blob/trunk/checks/file-checks.php
* Test coverage for the new checks is added, also indirectly covering the base class.

Closes #174

### How to test the Change
* Install a plugin that contains any of the following:
    * compressed files, e.g. ZIP, TAR, GZ, etc.
    * `.phar` files
    * version control checkouts, e.g. a `.git` or `.svn` folder
    * hidden files starting with `.`
    * obfuscated PHP files obfuscated with either Zend Guard, Source Guardian, or ionCube
* Ensure those plugin rule violations are flagged when running the plugin checker for that plugin.
* Alternatively to real plugins, you can also install/activate the relevant plugins in the `tests/phpunit/testdata/plugins` folder.

### Checklist:


includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_vcs_directories( Check_Result $result, array $files ) {
+		$directories = array_flip( array( '.git', '.svn', '.hg', '.bzr' ) );
+
+		$vcs_directories = array_filter(
+			array_unique(
```
mukeshpanchal27, line 117: As we have static array above we don't get any duplicate item.

tests/phpunit/testdata/plugins/test-plugin-code-obfuscation-ioncube-errors/load.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Plugin Name: Test Plugin Code Obfuscation ionCube Errors
+ * Plugin URI: https://github.com/wordpress/plugin-check
+ * Description: Plugin Check plugin from the WordPress Performance Team, a collection of tests to help improve plugin performance.
+ * Author: WordPress Performance Team
+ * Author URI: https://make.wordpress.org/performance/
+ * License: GPLv2 or later
+ * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ * Text Domain: test-plugin-code-obfuscation-ioncube-errors
```
mukeshpanchal27, line 10: Do we need to include `Requires at least`, `Requires PHP`, and `Version` for all new plugins?

tests/phpunit/testdata/plugins/test-plugin-code-obfuscation-sourceguardian-errors/obfuscated.php
```diff
@@ -0,0 +1,4 @@
+<?php
+if(function_exists('sg_load')) {
```
mukeshpanchal27, line 2:
```suggestion
if ( function_exists( 'sg_load' ) ) {
```
Correct format

includes/Checker/Checks/Code_Obfuscation_Check.php
```diff
@@ -0,0 +1,143 @@
...
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_zendguard( Check_Result $result, array $php_files ) {
+		$obfuscated_file = self::file_preg_match( '/(<\?php \@Zend;)|(This file was encoded by)/', $php_files );
+		if ( $obfuscated_file ) {
+			$result->add_message(
+				true,
+				sprintf(
+					'Code Obfuscation tools are not permitted. Detected: %s',
```
mukeshpanchal27: Should use translation function for all this strings. WDYT?

mukeshpanchal27 (commented):
Great work, @felixarntz! 👏  I have left some feedback and have a few questions. Overall, it looks solid to me.

jjgrainger (commented):
Thanks @felixarntz all looks good, though I think some of @mukeshpanchal27 feedback may be worth addressing.

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @param string $needle The substring to search for.
+	 * @return string|bool File path if needle was found, false otherwise.
+	 */
+	final protected static function file_str_contains( array $files, $needle ) {
+		// Backward compatibility prior to PHP 8.
+		$find_needle = 'str_contains';
+		if ( ! function_exists( 'str_contains' ) ) {
+			$find_needle = function( $haystack, $needle ) {
+				return false !== strpos( $haystack, $needle );
+			};
```
spacedmonkey: Could we just require `wp-includes/compat.php` here?

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_compressed_files( Check_Result $result, array $files ) {
+		$compressed_files = self::filter_files_by_extensions( $files, array( 'zip', 'gz', 'tgz', 'rar', 'tar', '7z' ) );
+		if ( $compressed_files ) {
+			foreach ( $compressed_files as $file ) {
+				$result->add_message(
+					true,
+					'Compressed files are not permitted.',
```
spacedmonkey: Should this not be translated?

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_phar_files( Check_Result $result, array $files ) {
+		$phar_files = self::filter_files_by_extension( $files, 'phar' );
+		if ( $phar_files ) {
+			foreach ( $phar_files as $file ) {
+				$result->add_message(
+					true,
+					'Phar files are not permitted.',
```
spacedmonkey: Should this not be translated?

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Looks for VCS directories and amends the given result with an error if found.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_vcs_directories( Check_Result $result, array $files ) {
+		$directories = array_flip( array( '.git', '.svn', '.hg', '.bzr' ) );
```
spacedmonkey, line 114: Wonder if we should also add .idea, as phpstorm adds these as well.

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+			}
+		);
+
+		if ( $vcs_directories ) {
+			// Only use an error in production, otherwise a warning.
+			$is_error = ( ! defined( 'WP_DEBUG' ) || ! WP_DEBUG ) && 'production' === wp_get_environment_type();
+			foreach ( $vcs_directories as $dir ) {
+				$result->add_message(
+					$is_error,
+					'Version control checkouts should not be present.',
```
spacedmonkey: https://github.com/10up/plugin-check/pull/175/files#r1205362655

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Looks for VCS directories and amends the given result with an error if found.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_vcs_directories( Check_Result $result, array $files ) {
```
spacedmonkey, line 113: Not, I have used composer to deploy WP + plugins to production. If composer does a git checkout, this would result in .git files in production.

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	protected function check_files( Check_Result $result, array $files ) {
+		if ( $this->flags & self::TYPE_COMPRESSED ) {
+			$this->look_for_compressed_files( $result, $files );
```
spacedmonkey, line 58: These 4 methods seem like they do the same thing. Could we make this more generic like this. 
```
$this->look_for_files( $result, $files, array( 'zip', 'gz', 'tgz', 'rar', 'tar', '7z' ),  'Compressed files are not permitted.', 'compressed_files' );
```

includes/Checker/Checks/Code_Obfuscation_Check.php
```diff
@@ -0,0 +1,143 @@
...
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_ioncube( Check_Result $result, array $php_files ) {
+		$obfuscated_file = self::file_str_contains( $php_files, 'ionCube' );
+		if ( $obfuscated_file ) {
+			$result->add_message(
+				true,
+				sprintf(
+					'Code Obfuscation tools are not permitted. Detected: %s',
```
spacedmonkey: Should these be translated?

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,227 @@
...
+	 * This is a wrapper around the native `str_contains()` function that will find the needle within the list of
+	 * files.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array  $files  List of absolute file paths.
+	 * @param string $needle The substring to search for.
+	 * @return string|bool File path if needle was found, false otherwise.
+	 */
+	final protected static function file_str_contains( array $files, $needle ) {
```
spacedmonkey, line 155: Why does this need to static?

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,227 @@
...
+	/**
+	 * Filters a given list of files to only contain those with specific extension.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array  $files     List of absolute file paths.
+	 * @param string $extension File extension to match.
+	 * @return array Filtered $files list.
+	 */
+	final protected static function filter_files_by_extension( array $files, $extension ) {
```
spacedmonkey, line 77: Why does this need to be static?

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,227 @@
...
+	/**
+	 * Filters a given list of files to only contain those with specific extensions.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $files      List of absolute file paths.
+	 * @param array $extensions List of file extensions to match.
+	 * @return array Filtered $files list.
+	 */
+	final protected static function filter_files_by_extensions( array $files, array $extensions ) {
```
spacedmonkey, line 90: Why does this need to be static?

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -0,0 +1,184 @@
...
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_hidden_files( Check_Result $result, array $files ) {
+		// Any files outside of 'vendor' or 'node_modules' directories that start with a period.
+		$hidden_files = self::filter_files_by_regex( $files, '/^((?!\/vendor\/|\/node_modules\/).)*\/\.\w+(\.\w+)*$/' );
+		if ( $hidden_files ) {
+			foreach ( $hidden_files as $file ) {
+				$result->add_message(
+					true,
+					'Hidden files are not permitted.',
```
spacedmonkey: Translation?

tests/phpunit/testdata/plugins/test-plugin-code-obfuscation-ioncube-errors/load.php
spacedmonkey: These are test plugins, I don't think they are needed?

spacedmonkey:
Screenshot 2023-05-25 at 12 52 26

Styling of this error looks a little strange.

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param array  $files  List of absolute file paths.
+	 * @param string $needle The substring to search for.
+	 * @return string|bool File path if needle was found, false otherwise.
+	 */
+	final protected static function file_str_contains( array $files, $needle ) {
+		// Backward compatibility prior to PHP 8.
+		$find_needle = 'str_contains';
+		if ( ! function_exists( 'str_contains' ) ) {
```
spacedmonkey: Why would this need exist?

includes/Checker/Checks/File_Type_Check.php
felixarntz: This doesn't filter the above array though. It filters the list of files, and as soon as more than 1 file is in the same directory, there will be duplicates.

tests/phpunit/testdata/plugins/test-plugin-code-obfuscation-ioncube-errors/load.php
felixarntz: Agreed, no need to include those things as they are test plugins.

tests/phpunit/testdata/plugins/test-plugin-code-obfuscation-sourceguardian-errors/obfuscated.php
felixarntz: Even though this violates WPCS, this must not be changed since that is the string to look for to detect Source Guardian. See the original check linked in the PR description.

It's a test plugin anyway, which is excluded from linting for reasons like this.

includes/Checker/Checks/Code_Obfuscation_Check.php
felixarntz: Good point, I copied that over from the other code. Will update to use translation functions.

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: I forgot that we require WP 6.0+ for the plugin checker, therefore I added this for older WP version support. But not needed, I remove this entirely, since `str_contains()` should always exist in WP 6.0+.

includes/Checker/Checks/File_Type_Check.php
felixarntz: See https://github.com/10up/plugin-check/pull/175#discussion_r1205828773, will fix.

includes/Checker/Checks/File_Type_Check.php
felixarntz: See above.

includes/Checker/Checks/File_Type_Check.php
felixarntz: This test is only for version control directories, `.idea` is an unrelated hidden file, which will anyways be covered by the `look_for_hidden_files()` method.

includes/Checker/Checks/File_Type_Check.php
felixarntz: This is taken over from the plugin review team's tool, so I think we should stick with that for now for parity. If this becomes a concern, we can further down in the future report an issue, but let's not do that just yet.

includes/Checker/Checks/File_Type_Check.php
felixarntz: Will update.

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: Because these methods have nothing to do with the class instance. Static methods are bad when they manage state, but all of the static methods in this class are stateless.

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: See my reply in https://github.com/10up/plugin-check/pull/175#discussion_r1205834776.

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: Same as https://github.com/10up/plugin-check/pull/175#discussion_r1205834776

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: See https://github.com/10up/plugin-check/pull/175#discussion_r1205830647, will fix.

felixarntz:
@mukeshpanchal27 @jjgrainger @spacedmonkey Please take another look, I have addressed all your feedback.

There is one PHPStan failure, however that is invalid since `str_contains` is available in all WordPress versions that we support. How can we instruct PHPStan to ignore that? cc @swissspidy 
```
Error: Function str_contains not found.
 ------ --------------------------------------------------------------------- 
  Line   includes/Checker/Checks/Abstract_File_Check.php                      
 ------ --------------------------------------------------------------------- 
  158    Function str_contains not found.                                     
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols   
 ------ ---------------------------------------------------------------------
```

swissspidy:
> There is one PHPStan failure, however that is invalid since `str_contains` is available in all WordPress versions that we support. How can we instruct PHPStan to ignore that? cc @swissspidy

Great question!

Looks like the stubs don't include that polyfill, maybe that's a bug.

For now I recommend just ignoring this error, see 502400ee15d4ed70524980c2cb8a9a88729bfea6

felixarntz:
Thanks @swissspidy! 🙌

includes/Checker/Checks/Abstract_File_Check.php
spacedmonkey: Static methods are pointless. Why not just have a namespaced function? Is there is no benefit for these being static, then make them not static.

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -0,0 +1,219 @@
...
+	 */
+	private static function file_get_contents( $file ) {
+		if ( isset( self::$file_contents_cache[ $file ] ) ) {
+			return self::$file_contents_cache[ $file ];
+		}
+
+		self::$file_contents_cache[ $file ] = file_get_contents( $file );
+
+		return self::$file_contents_cache[ $file ];
+	}
```
spacedmonkey, line 183: So we are using this as a cache?

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: I don't think that's correct. If static methods are pointless, why do they exist in lots of programming languages? One benefit of using a static method here instead of a function is that it allows the method to be scoped to a specific class (`protected`) rather than a function any code can call. The more code is public, the harder it is to make changes without breaking backward compatibility.

includes/Checker/Checks/Abstract_File_Check.php
felixarntz: Yes, this is particularly relevant as the same file contents may be needed again across multiple checks in a single request. Note that this was inspired by https://github.com/WordPress/plugin-check/blob/trunk/checks/check-base.php#L159 which does the same thing, which makes sense to me.

mukeshpanchal27 (approved):
Thanks @felixarntz for the update. The changes look good to me.

jjgrainger (approved):
Thanks @felixarntz changes look good, approved

PR #181, opened 2023-06-01
Fix running tests not working unless the repository is cloned to a `plugin-check` directory

mukeshpanchal27:
When having the repository checked out to another folder name than `plugin-check`, the `npm run test-php` and `npm run test-php-multisite` scripts will fail since they have the `plugin-check` path hard-coded.

We should fix that, for example so that the script detects the current folder name ("base name").

### Checklist:


jjgrainger (approved):

felixarntz (approved):

PR #186, opened 2023-06-06
Add `Enqueued_Scripts_In_Footer_Check` static check

### LINKED ISSUE #22: Create Enqueued_Scripts_In_Footer_Check
Create Enqueued_Scripts_In_Footer_Check class, and this class will basically warn for following things:

- Checks registered and enqueued scripts in the plugin and warn if the $in_footer parameter is false, It is encouraged that any non-essential scripts are loaded in the footer to prevent render-blocking on the page.
- It will also check that the version parameter is set to a non-falsy value. Using versions in the scripts can be beneficial for performance and cache busting.

This will use the existing WordPress.WP.EnqueuedResourceParameters sniff in the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards/).

## Acceptance Criteria

- Class `Enqueued_Scripts_In_Footer_Check` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Scripts_In_Footer_Check` should extend the `Abstract_PHP_CodeSniffer_Check` abstract class
    - `'sniffs' => 'WordPress.WP.EnqueuedResourceParameters'`
    - `'standard' => 'WordPress'`
- A sample code snippet should be added to the `tests/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-with-errors/load.php` file, which triggers the `WordPress.WP.EnqueuedResourceParameters` error. The following code can be used as an example:

```
add_action(
	'wp_enqueue_scripts',
	function() {
		wp_enqueue_script(
			'plugin_check_header_script',
			plugin_dir_url( __FILE__ ) . 'header.js'
		);
	}
);
```
- Another sample code snippet should be added to the `tests/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-without-errors/load.php` file, which does not trigger any errors. The code snippet can be as follows::

```
add_action(
	'wp_enqueue_scripts',
	function() {
		wp_enqueue_script(
			'plugin_check_footer_script',
			plugin_dir_url( __FILE__ ) . 'footer.js',
			array(),
			'1.0.0',
			true
		);
	}
);
```

## Tests Coverage
- The `test_run_with_errors()` method should verify that errors are returned when running the check.
- The `test_run_without_errors()` method should confirm that no errors are returned when running the check.

mukeshpanchal27:
Closes #22

### Checklist:


mukeshpanchal27:
@jjgrainger, could you please take a look at why the unit tests are failing? It seems unrelated to the changes I made. cc. @joemcgill @felixarntz

costdev:
After a ping from @mukeshpanchal27, I took a look into the test failure.

The failure seems to be caused by `\PHP_CodeSniffer\Config::$overriddenDefaults` being a `static` property. As a test, removing `static` and updating references from `self::$overriddenDefaults` to `$this->overriddenDefaults` allows the tests to pass.

This is actually a pre-existing issue in the plugin and affects four tests:
- `I18n_Usage_Check_Tests::test_run_with_errors()`
- `I18n_Usage_Check_Tests::test_run_without_errors()`
- `Enqueued_Scripts_In_Footer_Check_Tests::test_run_with_errors()` (added in this PR)
- `Enqueued_Scripts_In_Footer_Check_Tests::test_run_without_errors()` (added in this PR)

There are two ways I can think to resolve this:

1. Add the following docblock to each test class (or just the specific methods) that run PHPCS checks:
```php
/**
 * @runTestsInSeparateProcesses (or @runInSeparateProcess for a specific test method)
 * @preserveGlobalState disabled
 */
```
However, this will very noticeably slow down the respective tests and will make the test suite take even longer if more PHPCS checks are added in future.

2. Add the following to `Abstract_PHP_CodeSniffer_Check::run()` between setting `$_SERVER['argv']` and the `try/catch` block (or in a `set_up()` method for tests only):
```php
/*
 * Reset \PHP_CodeSniffer\Config::$overriddenDefaults to prevent
 * incorrect results when running multiple checks.
 */
$reflected_phpcs_config = new \ReflectionClass( '\PHP_CodeSniffer\Config' );
$reflected_phpcs_config->setStaticPropertyValue( 'overriddenDefaults', array() );
```
This will not slow down the tests and should work in any context in which the method is run, but is a little 'hacky' as it uses the Reflection API.

Edit: The above way of using the Reflection API will fail on earlier PHP versions. This should work:

```php
 $reflected_phpcs_config = new \ReflectionClass( '\PHP_CodeSniffer\Config' );
 $overridden_defaults    = $reflected_phpcs_config->getProperty( 'overriddenDefaults' );
 $overridden_defaults->setAccessible( true );
 $overridden_defaults->setValue( array() );
 $overridden_defaults->setAccessible( false );
```

mukeshpanchal27:
@costdev, thank you for your valuable feedback. I have thoroughly investigated the issue, and it appears that the root cause lies in the behavior of `PHP_CodeSniffer`. It does not update the configuration for each run, and while I did find a workaround, it is quite hacky and may not be compatible across different PHP versions.

Considering this, I believe it would be beneficial to raise the issue with `PHP_CodeSniffer` upstream. By doing so, we can involve the maintainers and discuss the best approach to address this limitation. In the meantime, I will move this pull request into the code review phase so that other team members can examine it and provide guidance on finding a proper solution for this issue.

cc. @jjgrainger @joemcgill @felixarntz @spacedmonkey

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -74,6 +74,18 @@ final public function run( Check_Result $result ) {
...
+		 * incorrect results when running multiple checks.
+		 *
+		 * PHPStan ignore reason: PHPStan raised an issue because we can't
+		 * use class in ReflectionClass.
+		 *
+		 * @phpstan-ignore-next-line
+		 */
+		$reflected_phpcs_config = new \ReflectionClass( '\PHP_CodeSniffer\Config' );
+		$reflected_phpcs_config->setStaticPropertyValue( 'overriddenDefaults', array() );
+
```
jjgrainger: @mukeshpanchal27 I see this is working for some tests but not others.

To resolve that first issue you could look at wrapping this in a `class_exists` conditional.

Secondly, as this is only needed to resolve the issues in tests we shouldn't include this in the production code here. Can we explore create a `Static_Check_UnitTestCase` similar to the [`Runtime_Check_UnitTestCase`](https://github.com/10up/plugin-check/blob/trunk/tests/phpunit/utils/TestCase/Runtime_Check_UnitTestCase.php) and including this in a setup method? I'm not 100% sure this would work but I'm against including this code here just to fix unit tests.

cc @joemcgill

jjgrainger: @mukeshpanchal27 as well as a class exists check we will also want to include a check for the property as well.

jrfnl:
> Considering this, I believe it would be beneficial to raise the issue with `PHP_CodeSniffer` upstream. By doing so, we can involve the maintainers and discuss the best approach to address this limitation.

PHPCS maintainer here: you are doing things wrong and trying to force a round block into a square hole. Not our problem. Raising this upstream is not going to go anywhere.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
mukeshpanchal27: @jjgrainger, I have made slight updates to the code to ensure compatibility with PHP 7.X, and now all the tests are passing. ✅

tests/phpunit/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-without-errors/load.php
```diff
@@ -0,0 +1,18 @@
...
+		wp_enqueue_script(
+			'plugin_check_test_script',
+			plugin_dir_url( __FILE__ ) . 'footer.js',
+			array(),
+			'1.0.0',
+			true
+		);
+	}
+);
+
```
mukeshpanchal27:
```suggestion
```

tests/phpunit/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-without-errors/footer.js
```diff
@@ -0,0 +1 @@
+console.log( 'test' );
```
spacedmonkey, line 1: Do these files need to exist?

tests/phpunit/Checker/Checks/Enqueued_Scripts_In_Footer_Check_Tests.php
```diff
@@ -0,0 +1,57 @@
...
+
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Enqueued_Scripts_In_Footer_Check;
+use WordPress\Plugin_Check\Test_Utils\TestCase\Static_Check_UnitTestCase;
+
+class Enqueued_Scripts_In_Footer_Check_Tests extends Static_Check_UnitTestCase {
+
+	public function test_run_with_errors() {
+
```
spacedmonkey:
```suggestion
	public function test_run_with_errors() {
```

tests/phpunit/Checker/Checks/Enqueued_Scripts_In_Footer_Check_Tests.php
```diff
@@ -0,0 +1,57 @@
...
+		$this->assertArrayHasKey( 'code', $warnings['load.php'][9][9][0] );
+		$this->assertEquals( 'WordPress.WP.EnqueuedResourceParameters.MissingVersion', $warnings['load.php'][9][9][0]['code'] );
+
+		// Check for WordPress.WP.EnqueuedResourceParameters.NotInFooter warning on Line no 9 and column no at 9.
+		$this->assertArrayHasKey( 'code', $warnings['load.php'][9][9][1] );
+		$this->assertEquals( 'WordPress.WP.EnqueuedResourceParameters.NotInFooter', $warnings['load.php'][9][9][1]['code'] );
+	}
+
+	public function test_run_without_errors() {
+
```
spacedmonkey:
```suggestion
	public function test_run_without_errors() {
```

tests/phpunit/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-without-errors/footer.js
mukeshpanchal27: We enqueue that JS in plugin file so yes.

tests/phpunit/Checker/Checks/Enqueued_Scripts_In_Footer_Check_Tests.php
```diff
@@ -0,0 +1,55 @@
...
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Enqueued_Scripts_In_Footer_Check;
+use WordPress\Plugin_Check\Test_Utils\TestCase\Static_Check_UnitTestCase;
+
+class Enqueued_Scripts_In_Footer_Check_Tests extends Static_Check_UnitTestCase {
+
+	public function test_run_with_errors() {
+		$enqueued_scripts_in_footer_check = new Enqueued_Scripts_In_Footer_Check();
+		$check_context                    = new Check_Context( TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-with-errors/load.php' );
```
felixarntz: After merging `trunk` back, please update this to use the new constant introduced in #189.

tests/phpunit/Checker/Checks/Enqueued_Scripts_In_Footer_Check_Tests.php
```diff
@@ -0,0 +1,55 @@
...
+		$this->assertEquals( 'WordPress.WP.EnqueuedResourceParameters.MissingVersion', $warnings['load.php'][9][9][0]['code'] );
+
+		// Check for WordPress.WP.EnqueuedResourceParameters.NotInFooter warning on Line no 9 and column no at 9.
+		$this->assertArrayHasKey( 'code', $warnings['load.php'][9][9][1] );
+		$this->assertEquals( 'WordPress.WP.EnqueuedResourceParameters.NotInFooter', $warnings['load.php'][9][9][1]['code'] );
+	}
+
+	public function test_run_without_errors() {
+		$enqueued_scripts_in_footer_check = new Enqueued_Scripts_In_Footer_Check();
+		$check_context                    = new Check_Context( TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/test-plugin-enqueued-scripts-in-footer-check-without-errors/load.php' );
```
felixarntz: See above.

tests/phpunit/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
```diff
@@ -94,7 +94,7 @@ public function test_run_with_errors() {
 		$this->assertNotEmpty( $warnings );
 
 		$this->assertEquals( 0, $results->get_error_count() );
-		$this->assertEquals( 4, $results->get_warning_count() );
+		$this->assertEquals( 8, $results->get_warning_count() );
```
felixarntz: Why this change? Is there something in our new code that affects this test? If so, we probably have to fix it.

tests/phpunit/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
```diff
@@ -113,6 +113,6 @@ public function test_run_with_errors_for_inline_script() {
...
```
felixarntz: Same question as above.

felixarntz (commented):
@mukeshpanchal27 This looks solid so far, just a few comments.

Also regarding the merge conflicts, I would suggest you first update and merge #187, and then merge `trunk` back into this PR here. In the meantime, please review my feedback below.

tests/phpunit/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
mukeshpanchal27: @felixarntz The unit tests automatically load all the plugin files, causing all the plugins to be loaded by default. In the commit https://github.com/10up/plugin-check/pull/186/commits/919ec2cd4adffa4d58533eb1720cfe94dcda2792, I removed that functionality to address the issue.

mukeshpanchal27:
@felixarntz @spacedmonkey, the PR is ready for another round of review. Thank you!

spacedmonkey (approved):
LGTM.

tests/phpunit/Checker/Checks/Enqueued_Scripts_Size_Check_Tests.php
felixarntz: @mukeshpanchal27 Good catch, great that you were able to fix that.

felixarntz (approved):
@mukeshpanchal27 Great work, LGTM!

PR #188, opened 2023-06-08
Add `Enqueued_Styles_Scope_Check` runtime check

### LINKED ISSUE #24: Create Enqueued_Styles_Scope_Check
## Description

Checks that styles are only enqueued in certain contexts and not just everywhere. This is the same as the [Enqueued_Scripts_Scope_Check](https://github.com/10up/plugin-check/issues/23) only for stylesheets instead of scripts.

## Acceptance Criteria

- Class `Enqueued_Styles_Scope_Check` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Styles_Scope_Check` should extend the `Abstract_Runtime_Check` class.
- Class `Enqueued_Styles_Scope_Check` will use the `URL_Aware` trait.
- Class `Enqueued_Styles_Scope_Check` will implement the`With_Shared_Preparations` interface.
- A `prepare()` method will be implemented in the class.
    - This method will save the `$GLOBALS['wp_styles'];` to a temporary variable.
    - Then set the `$GLOBALS['wp_styles'];` to null if set.
    - A clean up function is returned which uses the temporary variable to reinstate the original `$GLOBALS['wp_styles'];` value.
- A `get_shared_preparations()` method will be implemented.
    - The `get_shared_preparations()` method will be used to return a preparation that creates a demo post for all publicly viewable post types.
    - The method will return an array where the `Demo_Post_Creation_Preparation` class name is an array key and it's value is an array of posts data to create, compatible with [`wp_insert_post`](https://developer.wordpress.org/reference/functions/wp_insert_post/).
- A protected `get_urls()` method will be created.
    - The method will return an array of URLs containing the home URL and URLs for all publicly viewable post types.  
- Method `run( Check_Result $results )` should be implemented from the `Runtime_Check` abstract class.
    - The `run()` method will call the `run_for_urls()` method passing an array of URLs returned by the `get_urls()` method, a callable which will be the classes `check_url()` method, and the `Check_Results` instance.
    -  The `Check_Result $check_result` will be returned by the `run` method.
- A protected `check_url()` method will be created.
    - The method will accept an instance of the `Check_Result` class and a URL string as parameters.  
    - The method will reset the `$GLOBALS['wp_styles'];`
    - Then will call the `wp_enqueue_scripts()` to enqueue all of the checked plugins styles.
    - Using the object returned by `wp_styles()` all enqueued styles for the URL will be looped over:
        - Check the style is part of the plugin being checked by matching the plugin URL with the style URL
        - It will return styles that print in all pages.
        - Return notice with the message.

## Tests Coverage
- `Enqueued_Styles_Scope_Check` class is expected to extend `Runtime_Check` class
- `Enqueued_Styles_Scope_Check` class is expected to be instance of `Runtime_Check` class

mukeshpanchal27:
Closes #24 

### Checklist:


tests/phpunit/Checker/Checks/Enqueued_Styles_Scope_Check_Tests.php
```diff
@@ -0,0 +1,54 @@
...
+		$this->assertEquals( 0, $results->get_error_count() );
+		$this->assertEquals( 1, $results->get_warning_count() );
+	}
+
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 *
+	 * @return void
+	 */
```
mukeshpanchal27: Added the `@runInSeparateProcess` annotation to separate the process for the test because PHPUnit is running in one long process and the tests are leaking in those processes. Without this annotation, the test fails with the following error:

```
1) Enqueued_Styles_Scope_Check_Tests::test_run_without_errors
Failed asserting that an array is empty.

/var/www/html/wp-content/plugins/plugin-check/tests/phpunit/Checker/Checks/Enqueued_Styles_Scope_Check_Tests.php:43
```

Although it slows down the test suite, we had to add this annotation as a temporary workaround. If we find any alternative solution, we can remove it.

cc. @jjgrainger @felixarntz @joemcgill

costdev: [This comment](https://github.com/10up/plugin-check/pull/186#issuecomment-1584433805) may provide an alternative solution.

mukeshpanchal27: Thanks, @costdev, for your feedback on the other issue. This is a different issue than the `PHP_CodeSniffer` issue, so it requires some troubleshooting.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+	 *
+	 * @return array List of URL strings (either full URLs or paths).
+	 *
+	 * @throws Exception Thrown when a post type URL cannot be retrieved.
+	 */
+	protected function get_urls() {
+		$urls = array( home_url( '/' ) );
+
+		foreach ( $this->get_viewable_post_types() as $post_type ) {
+			$posts = get_posts(
```
spacedmonkey: Let's avoid using `get_posts`, it is always use `WP_Query`.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+				array(
+					'posts_per_page' => 1,
+					'post_type'      => $post_type,
+					'post_status'    => array( 'publish', 'inherit' ),
+				)
```
spacedmonkey: `lazy_load_term_meta`, `update_post_meta_cache`, `update_post_term_cache` set to false. 

Also `ignore_sticky_posts` and `no_found_rows`.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	protected function check_url( Check_Result $result, $url ) {
+		// Reset the wp_styles instance.
+		unset( $GLOBALS['wp_styles'] );
+
+		/*
+		 * Run the 'wp_enqueue_script' action, wrapped in an output buffer in case of any callbacks printing styles
+		 * directly. This is discouraged, but some plugins or themes are still doing it.
+		 */
+		ob_start();
```
spacedmonkey, line 219:
```suggestion
        $wp_styles = wp_styles();
		ob_start();
```

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+			wp_enqueue_scripts();
+			wp_styles()->do_footer_items();
```
spacedmonkey:
```suggestion
			$wp_styles->do_footer_items();
```

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+		ob_get_clean();
+
+		foreach ( wp_styles()->done as $handle ) {
+			$style = wp_styles()->registered[ $handle ];
```
spacedmonkey:
```suggestion
		foreach ( $wp_styles->done as $handle ) {
			$style = $wp_styles->registered[ $handle ];
```

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,234 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param string       $url    URL to run the check for.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	protected function check_url( Check_Result $result, $url ) {
```
spacedmonkey: What does the $url parameter do? I can't see it used.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Addressed in https://github.com/10up/plugin-check/pull/188/commits/112f11490c3b550b69d0b0e146979fae20c27c52

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Added in https://github.com/10up/plugin-check/pull/188/commits/112f11490c3b550b69d0b0e146979fae20c27c52

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Nice catch. remove it in https://github.com/10up/plugin-check/pull/188/commits/112f11490c3b550b69d0b0e146979fae20c27c52

tests/phpunit/Checker/Checks/Enqueued_Styles_Scope_Check_Tests.php
mukeshpanchal27: The issue was that all the plugins in the unit tests were autoloaded, which I removed in https://github.com/10up/plugin-check/pull/188/commits/5084dcdf41d389229ad5141304d9ec8f9e1fe8e0. Now it is working fine.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,247 @@
...
+	/**
+	 * Runs the check on the plugin and amends results.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check results to amend and the plugin context.
+	 */
+	public function run( Check_Result $result ) {
+		$this->run_for_urls(
+			$this->get_urls(),
```
spacedmonkey:
```suggestion
        $urls = $this->get_urls();
		$this->run_for_urls(
			$urls,
```

Get URL, does `WP_Query`, let's cache these in a variable.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,247 @@
...
+			function () use ( $result ) {
+				$this->check_url( $result );
+			}
+		);
+
+		if ( ! empty( $this->plugin_styles ) ) {
+			$url_count = count( $this->get_urls() );
```
spacedmonkey:
```suggestion
			$url_count = count( $urls );
```

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,247 @@
...
+					$the_query->the_post();
+
+					$urls[] = get_permalink();
+
+					// Add post category link.
+					$id = get_the_ID();
+					if ( 'post' === get_post_type( $id ) ) {
+						$category = get_the_category( $id );
+						$urls[]   = get_category_link( $category[0]->term_id );
+					}
```
spacedmonkey: I think we should test all taxonomy types as well here. Just testing categories seems a little random.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,247 @@
...
+	 * Gets the list of URLs to run this check for.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array List of URL strings (either full URLs or paths).
+	 *
+	 * @throws Exception Thrown when a post type URL cannot be retrieved.
+	 */
+	protected function get_urls() {
+		$urls = array( home_url( '/' ) );
```
spacedmonkey: We could test author and search pages as well.. Not to mention date based archives.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,247 @@
...
+				continue;
+			}
+
+			$style_path = str_replace( $result->plugin()->url(), $result->plugin()->path(), $style->src );
+
+			if ( isset( $this->plugin_style_count[ $handle ] ) ) {
+				$this->plugin_style_count[ $handle ] += 1;
+			} else {
+				$this->plugin_style_count[ $handle ] = 1;
+			}
```
spacedmonkey:
```suggestion
			if ( ! isset( $this->plugin_style_count[ $handle ] ) ) {
				$this->plugin_style_count[ $handle ] = 0;
			} 
			$this->plugin_style_count[ $handle ] += 1;
```

A little cleaner logic.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Updated. Thanks!

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Added search and author page URL.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: For the initial implementation of the check, I don't think it's a blocker if it's not required. We can always add it later if needed. What do you think? cc. @jjgrainger @felixarntz

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,249 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return array List of URL strings (either full URLs or paths).
+	 *
+	 * @throws Exception Thrown when a post type URL cannot be retrieved.
+	 */
+	protected function get_urls() {
+		$urls   = array( home_url( '/' ) );
+		$urls[] = get_search_link();
+		$urls[] = get_author_posts_url( (int) get_the_author_meta( 'ID' ) );
```
spacedmonkey:
```suggestion
		$urls   = array( home_url( '/' ) , get_search_link(), get_author_posts_url( 1 ) );
```

Just setup on array like this. Also just hardcode user 1, as this always exists.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -0,0 +1,249 @@
...
+					$the_query->the_post();
+
+					$urls[] = get_permalink();
+
+					// Add post category link.
+					$id = get_the_ID();
+					if ( 'post' === get_post_type( $id ) ) {
+						$category = get_the_category( $id );
+						$urls[]   = get_category_link( $category[0]->term_id );
+					}
```
spacedmonkey:
```suggestion
                     $post = get_post();
					$taxonomy_names = get_post_taxonomies( $post );
					$taxonomy_names = array_map( $taxonomy_names, 'is_taxonomy_viewable' );
					foreach( $taxonomy_names as $taxonomy_name ) {
						 $terms = get_the_terms( $post, $taxonomy_name );
						 if( !is_array( $terms ) ){
						    continue;
						 }
						 foreach( $terms as $term ){
						   $term_link = get_term_link( $term );
						   if( !is_wp_error( $term_link ) ){
						        $urls[]   = $term_link;
						   }
						 }
					}
```

Let's make it generic like this.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Thanks for suggestion.

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
mukeshpanchal27: Thanks, address in https://github.com/10up/plugin-check/pull/188/commits/4cb88eb2f13d2a790dac875faf96a3a89732f02c

tests/phpunit/Checker/Checks/Enqueued_Styles_Scope_Check_Tests.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+use WordPress\Plugin_Check\Checker\Checks\Enqueued_Styles_Scope_Check;
+use WordPress\Plugin_Check\Test_Utils\TestCase\Runtime_Check_UnitTestCase;
+
+class Enqueued_Styles_Scope_Check_Tests extends Runtime_Check_UnitTestCase {
+
+	public function test_run_with_errors() {
+		// Load the test plugin.
+		require TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/test-plugin-enqueued-styles-scope-check-with-error/load.php';
```
spacedmonkey:
```suggestion
		require UNIT_TESTS_PLUGIN_DIR . 'test-plugin-enqueued-styles-scope-check-with-error/load.php';
```

tests/phpunit/Checker/Checks/Enqueued_Styles_Scope_Check_Tests.php
```diff
@@ -0,0 +1,48 @@
...
+		$this->assertEmpty( $errors );
+		$this->assertNotEmpty( $warnings );
+
+		$this->assertEquals( 0, $results->get_error_count() );
+		$this->assertEquals( 1, $results->get_warning_count() );
+	}
+
+	public function test_run_without_errors() {
+		// Load the test plugin.
+		require TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/test-plugin-enqueued-styles-scope-check-without-error/load.php';
```
spacedmonkey:
```suggestion
		require UNIT_TESTS_PLUGIN_DIR . 'test-plugin-enqueued-styles-scope-check-without-error/load.php';
```

spacedmonkey (approved):

mukeshpanchal27:
@felixarntz @jjgrainger, Can you please review PR. Thank you!

spacedmonkey (approved):

jjgrainger (approved):

PR #189, opened 2023-06-09
Add `Plugin_Updater_Check`

### LINKED ISSUE #179: Implement plugin-updaters check

https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php (can be implemented using the new abstract base class from https://github.com/10up/plugin-check/pull/175 )

### Acceptance Criteria
- The `Plugin_Updater_Check` class should be created and located at `includes/Checker/Checks`.
- The `Plugin_Updater_Check` class should extend the `Abstract_File_Check` abstract class.
- The `look_for_update_uri_header()` method should be defined within the `Plugin_Updater_Check` class.
- The `look_for_update_uri_header()` method should retrieve the contents of the plugin's main file and extract the plugin headers using [get_plugin_data](https://developer.wordpress.org/reference/functions/get_plugin_data/) function.
- The `look_for_update_uri_header()` method should check if the UpdateURI header is present and has a value.
- If the UpdateURI header is not present, an error should be reported.
- The `look_for_updater_file()`, `look_for_plugin_updaters()` and `look_for_updater_routines()` method should be defined within the `Plugin_Updater_Check` class.
- The `look_for_updater_file()` method (based on [check_updaters()](https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php#L17-L30)) should scan the plugin's files for `plugin-update-checker.php` file. 
- If that file found, an error should be reported.
- The `look_for_plugin_updaters()` method (based on [check_updaters()](https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php#L32-L51)) should scan the plugin's files for any indications of plugin updaters being implemented. 
- If any plugin updater implementations are found, an error should be reported.
- The `look_for_updater_routines()` method (based on [check_updater_warnings()](https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php#L54)) should scan the plugin's files for any indications of modifications to WordPress update routines.
- If any modifications to WordPress update routines are found, a warning should be reported.

### Test Coverage
- The `look_for_update_uri()` method should be tested to ensure it reports an error when the UpdateURI header is not present in the plugin header.
- The `look_for_updater_file()`, `look_for_plugin_updaters()` and `look_for_updater_routines()` methods should be tested to ensure they raise a error/warning when the plugin header does not have an UpdateURI and when there are indications of plugin updaters and WordPress update routines in the plugin's files.
- Test cases should cover scenarios where the UpdateURI header is missing, present but empty, and present with a valid value.

### Code of Conduct

mukeshpanchal27:
Closes #179 

### Checklist:


includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,187 @@
...
+	 *
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_updater_file( Check_Result $result, array $php_files ) {
+
+		$filenames     = array_map( 'strtolower', array_map( 'basename', $php_files ) );
+		$blocked_files = array( 'plugin-update-checker.php' );
+		$matches       = array_intersect( $filenames, $blocked_files );
+
```
felixarntz: I think it would be more efficient to use the `filter_files_by_regex()` method in the base class here instead of the several array functions. More importantly, the code here has the problem that it only shows the basename of the file in the error message instead of the full file name including path. I know it's like that in the original code too, but it isn't right there either.

I think you can do something like:
```php
$found_files = self::filter_files_by_regex( $php_files, '/plugin-update-checker\.php$/' )
```

If the result is not empty, it will include the full file paths.

includes/Plugin_Context.php
```diff
@@ -85,4 +85,15 @@ public function location() {
...
+	/**
+	 * Returns the plugin absolute path of the main file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The absolute path to the plugin main file.
+	 */
+	public function abspath() {
+		return $this->main_file;
+	}
```
felixarntz: This method isn't necessary. You can always use `WP_PLUGIN_DIR` plus the plugin's basename (via the above `basname()` method) to get the path to the plugin main file.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,187 @@
...
+			"'plugin-update-checker'",
+			'WP_GitHub_Updater',
+			'WPGitHubUpdater',
+			'#class [A-Z_]+_Plugin_Updater#i',
+			'#updater\.\w+\.\w{2,5}#i',
+			'site_transient_update_plugins',
+		);
+
+		foreach ( $needles as $needle ) {
+			$match = self::file_str_contains( $php_files, $needle );
```
felixarntz: Some of the strings in `$needles` are regular expressions, so for those `file_str_contains()` doesn't work. Instead you need to use `file_preg_match()` for them.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,187 @@
...
+	protected function look_for_updater_routines( Check_Result $result, array $php_files ) {
+
+		$needles = array(
+			'#pre_set_site_transient_update_\w+#i',
+			'auto_update_plugin',
+			'#_site_transient_update_\w+#i',
+		);
+
+		foreach ( $needles as $needle ) {
+			$match = self::file_str_contains( $php_files, $needle );
```
felixarntz: Same comment as above applies here too.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,187 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result       The check result to amend, including the plugin context to check.
+	 * @param string       $updater_file Absolute path to the updater file found.
+	 * @param string       $message      Error message for updater.
+	 * @param bool         $error        Whether it is an error message. Defaults to true.
+	 * @param string       $code         Violation code according to the message. Default "plugin_updater_detected".
+	 */
+	private function add_result_error_for_file( Check_Result $result, $updater_file, $message, $error = true, $code = 'plugin_updater_detected' ) {
```
felixarntz: Maybe it would be better here to keep the parameters in the same order as they are in the `Check_Result::add_message()` method? Additionally, I would advise against using a default value for `$code`: Either it should always be passed, or it should always be the same value for the method, since otherwise it can become confusing.
```suggestion
	private function add_result_error_for_file( Check_Result $result, $error, $message, $code, $updater_file ) {
```

tests/phpunit/Checker/Checks/Plugin_Updater_Check_Tests.php
```diff
@@ -0,0 +1,73 @@
...
+		$warnings = $check_result->get_warnings();
+
+		$this->assertEmpty( $errors );
+		$this->assertEmpty( $warnings );
+
+		$this->assertEquals( 0, $check_result->get_error_count() );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
+	}
+
+	public function test_look_for_update_uri_header() {
```
felixarntz: Not a blocker, but I would generally advise for more granular tests per method rather than a single big test that covers every part of the check, which is more complex to maintain.

You can introduce bitwise flags in the check class to only test for specific things based on those flags passed in the constructor. Of course by default all methods in the check should be called, but for tests you could pass other flag values so that only a specific method is called. Please take a look at e.g. the `Code_Obfuscation_Check` class and its tests which follows that principle.

felixarntz (changes requested):
@mukeshpanchal27 This looks solid so far, though there are a few small problems to address. It would be good if you could also include test coverage for 1-2 regular expressions here, which is currently missing and is critical as right now those are actually not working.

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Thanks for feedback. Addressed in https://github.com/10up/plugin-check/pull/189/commits/e1572f3ffd07a9aeba3463cc7b981cdc2b386ead

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Nice catch. Missed it 😞

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Addressed in https://github.com/10up/plugin-check/pull/189/commits/e1572f3ffd07a9aeba3463cc7b981cdc2b386ead

tests/phpunit/Checker/Checks/Plugin_Updater_Check_Tests.php
mukeshpanchal27: Thanks for the great feedback, Agree with it and addressed in https://github.com/10up/plugin-check/pull/189/commits/e1572f3ffd07a9aeba3463cc7b981cdc2b386ead

mukeshpanchal27:
@felixarntz, the PR is ready for another round of review. Thank you!

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+		$plugin_header    = get_plugin_data( $plugin_main_file );
+		if ( ! empty( $plugin_header['UpdateURI'] ) ) {
+			$this->add_result_error_for_file(
+				$result,
+				true,
+				__( 'Plugin Updater detected. Use of the Update URI header is not helpful in plugins hosted on WordPress.org.', 'plugin-check' ),
+				'plugin_updater_detected',
+				$plugin_main_file
+			);
+		}
```
spacedmonkey, line 102: How do you know if this is hosted in WordPress.org?

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+
+		foreach ( $look_for_regex as $regex ) {
+			$updater_file = self::file_preg_match( $regex, $php_files, $matches );
+			if ( $updater_file ) {
+				$this->add_result_error_for_file(
+					$result,
+					true,
+					sprintf(
+						/* translators: %s: The match updater string. */
+						__( 'sdfsfsf Plugin Updater detected. These are not permitted in WordPress.org hosted plugins. Detected: %s', 'plugin-check' ),
```
spacedmonkey: sdfsfsf?

tests/phpunit/Checker/Checks/Plugin_Updater_Check_Tests.php
```diff
@@ -0,0 +1,109 @@
...
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Plugin_Updater_Check;
+
+class Plugin_Updater_Check_Tests extends WP_UnitTestCase {
+
+	/**
+	 * @dataProvider data_plugin_updater_check
+	 */
+	public function test_run_with_plugin_updater_errors( $type_flag, $plugin_basename, $expected_file, $code, $error ) {
+		$check_context = new Check_Context( TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/' . $plugin_basename );
```
spacedmonkey: `TESTS_PLUGIN_DIR . '/tests/phpunit/testdata`, should we add this as a const? It seems to be referenced a lot.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+	 *
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_plugin_updaters( Check_Result $result, array $php_files ) {
+
+		$look_for_regex = array(
+			'#\'plugin-update-checker\'#i',
+			'#WP_GitHub_Updater#i',
+			'#WPGitHubUpdater#i',
```
spacedmonkey: Are we staying we don't support github updater anymore?

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+			'#class [A-Z_]+_Plugin_Updater#i',
```
spacedmonkey, line 148: There seems to be lots of example of this on the repo. https://www.wpdirectory.net/search/01H2TFT46P7NR5E16666MSWVWF 

What is this all about?

includes/Checker/Checks/Plugin_Updater_Check.php
felixarntz: @spacedmonkey This test is specifically about plugin review team requirements for plugins hosted on wordpress.org, see https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php. So we should match their requirements for that.

includes/Checker/Checks/Plugin_Updater_Check.php
felixarntz: See https://github.com/10up/plugin-check/pull/189#discussion_r1228408591, this is about not using any updaters since that is prohibited on wordpress.org.

includes/Checker/Checks/Plugin_Updater_Check.php
felixarntz: See https://github.com/10up/plugin-check/pull/189#discussion_r1228408591

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * @since n.e.x.t
+ */
+class Plugin_Updater_Check extends Abstract_File_Check {
+
+	const TYPE_PLUGIN_UPDATE_URI_HEADER = 1;
+	const TYPE_PLUGIN_UPDATER_FILE      = 3;
+	const TYPE_PLUGIN_UPDATERS          = 5;
+	const TYPE_PLUGIN_UPDATER_ROUTINES  = 7;
+	const TYPE_ALL                      = 9; // Same as all of the above with bitwise OR.
```
felixarntz: These numbers do not work as bitwise flags: Any different config value has to be a power of 2, i.e. 2^0, 2^1, 2^2, etc. Only `TYPE_ALL`, which is a group of other values needs to be different, it needs to be all other values with a bitwise OR.

Please review https://www.hendrik-erz.de/post/bitwise-flags-are-beautiful-and-heres-why which is a good intro on how bitwise flags work. Let me know if you have any follow up questions. :)

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+	 *
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_updater_file( Check_Result $result, array $php_files ) {
+
+		$plugin_update_file = self::filter_files_by_regex( $php_files, '/plugin-update-checker\.php$/' );
+
+		if ( $plugin_update_file ) {
+			foreach ( $plugin_update_file as $file ) {
```
felixarntz: Nit-pick: Since it's an array, better to use a plural variable name.
```suggestion
		$plugin_update_files = self::filter_files_by_regex( $php_files, '/plugin-update-checker\.php$/' );

		if ( $plugin_update_files ) {
			foreach ( $plugin_update_files as $file ) {
```

includes/Checker/Checks/Plugin_Updater_Check.php
felixarntz: Why the `i` modifier at the end? These lookups are case-sensitive in https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php#L32, so we should keep them like that also when using a regex for them.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+	 */
+	protected function look_for_plugin_updaters( Check_Result $result, array $php_files ) {
+
+		$look_for_regex = array(
+			'#\'plugin-update-checker\'#i',
+			'#WP_GitHub_Updater#i',
+			'#WPGitHubUpdater#i',
+			'#class [A-Z_]+_Plugin_Updater#i',
+			'#updater\.\w+\.\w{2,5}#i',
+			'#site_transient_update_plugins#i',
```
felixarntz: See above, this value should always be lower-case, but the `i` modifier means it is case-insensitive.

includes/Checker/Checks/Plugin_Updater_Check.php
```diff
@@ -0,0 +1,226 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param array        $php_files List of absolute PHP file paths.
+	 */
+	protected function look_for_updater_routines( Check_Result $result, array $php_files ) {
+
+		$look_for_regex = array(
+			'#auto_update_plugin#i',
```
felixarntz: Same as above, this value is case-sensitive in https://github.com/WordPress/plugin-check/blob/trunk/checks/plugin-updaters.php#L57

felixarntz (commented):
Thanks @mukeshpanchal27, almost looks good to go. The only crucial thing to fix here is how the bitwise flags are used. Please also see the attached article which is some good documentation on that.

includes/Checker/Checks/Plugin_Updater_Check.php
spacedmonkey: That is a confusing check then. What if you run this test again your own plugin that is not hosted on WP.org repo? This is flagging a warning for something that is not an issue.

felixarntz: Not every check has to be run by everyone. This here specifically is a "wordpress.org minimum requirements" check, so it also needs to be available in our plugin, otherwise developers won't be able to sanity-check their plugin for those requirements.

If someone develops a plugin for somewhere outside wordpress.org _and_ is building their own updater, they can always choose to not run this check.

tests/phpunit/Checker/Checks/Plugin_Updater_Check_Tests.php
mukeshpanchal27: Added const in https://github.com/10up/plugin-check/pull/189/commits/ddc2211fd668a0e1b04542bd542436acd2bae7d0

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Thanks @felixarntz for the answers.

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Thanks for the details. In https://github.com/10up/plugin-check/pull/189/commits/d0ddc4a34ba23d71b30d0e017683699a3442a7fc i have corrected it.

includes/Checker/Checks/Plugin_Updater_Check.php
mukeshpanchal27: Nice catch. addressed in https://github.com/10up/plugin-check/pull/189/commits/40cdb93d2b6a2ec5564e3813f4f485e5f51462a9

mukeshpanchal27:
@felixarntz @spacedmonkey, the PR is ready for another round of review. Thank you!

spacedmonkey (approved):

jjgrainger (approved):

felixarntz (approved):
Thanks @mukeshpanchal27, great work!

PR #190, opened 2023-06-09
Add `Plugin_Header_Text_Domain_Check` static check

### LINKED ISSUE #177: Implement header textdomain check
Implements a plugin text domain check based on the [WordPress Plugin Check headers check](https://github.com/WordPress/plugin-check/blob/trunk/checks/header.php). This check can be implemented without using any base class, just needs to implement `Static_Check`.

## Acceptance Criteria
- Class `Plugin_Header_Text_Domain_Check` should exist in `includes/Checker/Checks` folder.
- Class should implement the `Static_Check` interface.
- The check will test the contents of the plugin headers to see if the `Text Domain:` value matches the plugins slug.
    - Get the contents of the plugin main files plugin headers using the [`get_plugin_data()`](https://developer.wordpress.org/reference/functions/get_plugin_data/) core function.
    - Get the value of the `Text Domain:` header.
    - Check the text domain matches the plugins slug.
    - If not, the check should raise a warning.
 
## Test Coverage
- `run()` method raises a warning when the plugin header text domain and slug do not match.
- `run()` method raises no warnings when the text domain and slug match.

mukeshpanchal27:
Closes #177 

### Checklist:

### DIFF
--- a/includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
+++ b/includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
@@ -34,7 +34,7 @@ class Plugin_Header_Text_Domain_Check implements Static_Check {
 			require_once ABSPATH . 'wp-admin/includes/plugin.php';
 		}
 
-		$plugin_main_file = $result->plugin()->abspath();
+		$plugin_main_file = WP_PLUGIN_DIR . '/' . $result->plugin()->basename();
 		$plugin_header    = get_plugin_data( $plugin_main_file );
 		$plugin_slug      = basename( $result->plugin()->path() );
 
--- a/includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
+++ b/includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Checks\Plugin_Header_Text_Domain_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use Exception;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Static_Check;
+
+/**
+ * Check for plugin header text domain.
+ *
+ * @since n.e.x.t
+ */
+class Plugin_Header_Text_Domain_Check implements Static_Check {
+
+	/**
+	 * Amends the given result by running the check on the associated plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 *
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	public function run( Check_Result $result ) {
+
+		if ( ! function_exists( 'get_plugin_data' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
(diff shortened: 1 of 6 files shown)


includes/Plugin_Context.php
```diff
@@ -85,4 +85,15 @@ public function location() {
...
+	/**
+	 * Returns the plugin absolute path of the main file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The absolute path to the plugin main file.
+	 */
+	public function abspath() {
+		return $this->main_file;
+	}
```
felixarntz: See my comment in #189, this shouldn't be needed. You can use `WP_PLUGIN_DIR` plus the `basename()` method.

includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
+	 *                   the check).
+	 */
+	public function run( Check_Result $result ) {
+
+		if ( ! function_exists( 'get_plugin_data' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
+		$plugin_main_file = $result->plugin()->abspath();
```
felixarntz: See below.
```suggestion
		$plugin_main_file = WP_PLUGIN_DIR . '/' . $result->plugin()->basename();
```

tests/phpunit/Checker/Checks/Plugin_Header_Text_Domain_Check_Tests.php
```diff
@@ -0,0 +1,48 @@
...
+
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Plugin_Header_Text_Domain_Check;
+
+class Plugin_Header_Text_Domain_Check_Tests extends WP_UnitTestCase {
+
+	public function test_run_with_errors() {
+		$plugin_header_text_domain_check = new Plugin_Header_Text_Domain_Check();
+		$check_context                   = new Check_Context( TESTS_PLUGIN_DIR . '/tests/phpunit/testdata/plugins/test-plugin-i18n-usage-errors/load.php' );
```
felixarntz: Better to create a new test plugin that specifically has the error we're checking for here instead of using another test plugin file that "accidentally" has the error.

felixarntz (changes requested):
@mukeshpanchal27 Thanks, this is almost good to go. Just a few points of feedback below.

includes/Checker/Checks/Plugin_Header_Text_Domain_Check.php
spacedmonkey: Could we use `$result->plugin()->location();`

mukeshpanchal27: @spacedmonkey, it seems that the function only returns the absolute path of the plugin folder, but we actually need the path to the plugin's main file.

tests/phpunit/Checker/Checks/Plugin_Header_Text_Domain_Check_Tests.php
mukeshpanchal27: Thanks for the feedback. Added new plugin.

mukeshpanchal27:
@felixarntz, the PR is ready for another round of review. Thank you!

spacedmonkey (approved):
SGTM

felixarntz (approved):

PR #209, opened 2023-06-26
Add `Localhost` Static Check

mukeshpanchal27:
Closes #207 

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #211, opened 2023-06-26
Add `No_Unfiltered_Uploads_Check` static check

### LINKED ISSUE #208: Implement check to avoid `ALLOW_UNFILTERED_UPLOADS` usage

See https://github.com/WordPress/plugin-check/blob/allow_unfiltered_uploads-check/checks/check-constants.php, let's implement this as a more specific `No_Unfiltered_Uploads_Check` for now.

### Designs

* Reimplement the above by extending `Abstract_File_Check`.

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #208 

### Checklist:

### DIFF
--- a/includes/Checker/Checks/No_Unfiltered_Uploads_Check.php
+++ b/includes/Checker/Checks/No_Unfiltered_Uploads_Check.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Checks\No_Unfiltered_Uploads_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use WordPress\Plugin_Check\Checker\Check_Result;
+
+/**
+ * Check for detecting "ALLOW_UNFILTERED_UPLOADS" constant in plugin files.
+ *
+ * @since n.e.x.t
+ */
+class No_Unfiltered_Uploads_Check extends Abstract_File_Check {
+
+	/**
+	 * Check the "ALLOW_UNFILTERED_UPLOADS" constant in file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The Check Result to amend.
+	 * @param array        $files  Array of plugin files.
+	 */
+	protected function check_files( Check_Result $result, array $files ) {
+		$php_files = self::filter_files_by_extension( $files, 'php' );
+		$file      = self::file_str_contains( $php_files, 'ALLOW_UNFILTERED_UPLOADS' );
+		if ( $file ) {
+			$result->add_message(
+				true,
+				__( 'ALLOW_UNFILTERED_UPLOADS is not permitted.', 'plugin-check' ),
+				array(
+					'code' => 'allow_unfiltered_uploads_detected',
+					'file' => $file,
+				)
+			);
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


includes/Checker/Checks/No_Unfiltered_Uploads_Check.php
```diff
@@ -0,0 +1,41 @@
...
+	 * Check the "ALLOW_UNFILTERED_UPLOADS" constant in file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The Check Result to amend.
+	 * @param array        $files  Array of plugin files.
+	 */
+	protected function check_files( Check_Result $result, array $files ) {
+		$php_files = self::filter_files_by_extension( $files, 'php' );
+		$file      = self::file_str_contains( $php_files, 'ALLOW_UNFILTERED_UPLOADS' );
```
spacedmonkey, line 29: This should be a check to if is `define( 'ALLOW_UNFILTERED_UPLOADS` exists right? You could check to see if it is defined or use the constant right?

tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
```diff
@@ -0,0 +1,19 @@
...
+ * Author URI: https://make.wordpress.org/performance/
+ * License: GPLv2 or later
+ * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ * Text Domain: test-plugin-unfiltered-uploads-errors
+ *
+ * @package test-plugin-unfiltered-uploads-errors
+ */
+
+// Define constant.
+define( 'ALLOW_UNFILTERED_UPLOADS', true );
```
spacedmonkey: Add a check for something like this 

```
if ( defined( 'ALLOW_UNFILTERED_UPLOADS' ) ){
   return;
}
```

and 

```
if( ALLOW_UNFILTERED_UPLOADS ) {
   return;
}
```

includes/Checker/Checks/No_Unfiltered_Uploads_Check.php
jjgrainger: I believe this is right as the intention is to flag any references to the constant whether it's being defined or used in the codebase.

This also matches the WP plugins implementation referenced in the AC's https://github.com/10up/plugin-check/issues/208

tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
jjgrainger: I agree, @mukeshpanchal27 let's add these examples to make sure the check picks up these cases.

jjgrainger (changes requested):
Thanks @mukeshpanchal27 overall looking good, would be good to add some additional test cases so we capture everything here.

felixarntz (approved):
Thanks @mukeshpanchal27, nothing to add from my end other than the test plugin change that Jonny mentioned. Also, this has to be refreshed with latest `trunk` now that #209 has been merged, to fix the merge conflict.

I'll mark this as approved already, but please address the one point of feedback :)

tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
mukeshpanchal27: Thanks for the feedback. Address in https://github.com/10up/plugin-check/pull/211/commits/5e076c03190b75af9cf1032c66d05ca60b25ebd2

jjgrainger (commented):

jjgrainger (approved):

PR #212, opened 2023-06-26
Disable submit button when plugin is not selected

### LINKED ISSUE #139: Improve error handling if the user clicks on the 'Check it' button without selecting a plugin
When you select the 'Check it' button without selecting a plugin in the dropdown, do you feel we should add a message in the UI?  A message does appear in the console.  See screenshot. 

image-4

jjgrainger:
This PR disables the 'Check it' button when a plugin is not selected in the dropdown.

This has been tested and works with the 'Check this plugin' links on the plugins page.

Closes #139

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -14,8 +14,6 @@
 	}
 
 	// Handle disabling the check it button when a plugin is not selected.
-	pluginsList.addEventListener( 'change', canRunChecks );
-
 	function canRunChecks() {
 		if ( '' === pluginsList.value ) {
 			checkItButton.disabled = true;
@@ -26,6 +24,7 @@
 
 	// Run on page load to test if dropdown is auto populated.
 	canRunChecks();
+	pluginsList.addEventListener( 'change', canRunChecks );
 
 	// When the Check it button is clicked.
 	checkItButton.addEventListener( 'click', ( e ) => {
@@ -33,6 +32,7 @@
 
 		resetResults();
 		checkItButton.disabled = true;
+		pluginsList.disabled = true;
 		spinner.classList.add( 'is-active' );
 
 		getChecksToRun()
@@ -69,6 +69,7 @@
 	function resetForm() {
 		spinner.classList.remove( 'is-active' );
 		checkItButton.disabled = false;
+		pluginsList.disabled = false;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


assets/js/plugin-check-admin.js
```diff
@@ -13,6 +13,21 @@
 		return;
 	}
 
+	// Handle disabling the check it button when a plugin is not selected.
+	pluginsList.addEventListener( 'change', canRunChecks );
```
felixarntz: Nit-pick: Maybe move this together with the `canRunChecks()` call, just to have those two places where the function is called co-located?

assets/js/plugin-check-admin.js
```diff
@@ -13,6 +13,21 @@
...
+
+	function canRunChecks() {
+		if ( '' === pluginsList.value ) {
+			checkItButton.disabled = true;
+		} else {
+			checkItButton.disabled = false;
+		}
+	}
```
felixarntz, line 23: Not directy affiliated with this PR, but now that we're adding this logic to deactivate the button, I see there is a place where our logic is somewhat problematic: If you start running checks, the button gets disabled, but you can still change the plugin dropdown to select another plugin, which would unset the `disabled` on the button.

Could you modify the relevant logic below to also set the _dropdown_ to `disabled` while checks are running (and then unset `disabled` also on the dropdown once checks are done)?

felixarntz (commented):
Thanks @jjgrainger for the PR. Only two small points of feedback here to improve the behavior a bit further. There's one edge-case we should probably address, which should be straightforward.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz I've moved this down with the `canRunChecks()` call.

assets/js/plugin-check-admin.js
jjgrainger: Thanks @felixarntz I've updated to disable the plugin list while checks are running, then re-enable once the checks are complete.

felixarntz (approved):
Thanks @jjgrainger, LGTM!

assets/js/plugin-check-admin.js
```diff
@@ -13,11 +13,26 @@
 		return;
 	}
 
+	// Handle disabling the check it button when a plugin is not selected.
```
mukeshpanchal27:
```suggestion
	// Handle disabling the Check it button when a plugin is not selected.
```

mukeshpanchal27 (approved):
Thanks @jjgrainger, The changes look good to me.

Issue #153, opened 2023-04-25
Rethink the WP-CLI approach


schlessera:
### Is your enhancement related to a problem? Please describe.

Hi performance team!

@spacedmonkey pinged me to see if I have any input on improving the WP-CLI integration.

The current approach to integrating with WP-CLI is both cumbersome and can lead to side effects due to the use of `object-cache.php` as a "hook".

I would suggest a different approach that is safer to use and will ultimately provide a better user experience.

1. Create a WP-CLI package that provides a `plugin check` command. This package is a separate repository and can be installed via something like `wp package install 10up/plugin-check-cli`. As a WP-CLI package, it is already loaded and active before WordPress is loaded, so there is no need for any special handling to get in early. The command can check for the existence of the plugin and make use of its files as required, or come with its own code where it makes sense. This allows for different handling in CLI than for the regular plugin, but also allows for code reuse where applicable (just ensure you can rely on a fixed set of interfaces!). It also gives full control over the WordPress environment and you can already hook into the WordPress lifecycle using `WP_CLI::add_wp_hook()`.
2. From within the plugin, conditionally add a WP-CLI `plugin check` only if that command is not yet registered (to avoid overriding the already installed WP-CLI package). This `plugin check` command does not actually run any checks, though, it only produces a helpful message about the missing WP-CLI package and how the user can install it. It will show something like this:
```shell
$ wp plugin check akismet
Error: The necessary WP-CLI package for running performance checks has not yet been installed.
Please install it using the following command and then run your request again:
  wp package install 10up/plugin-check-cli
```
3. Add version numbers and update checks into both the WP-CLI package and the plugin to ensure that you can detect unmet requirements and prompt users to run updates when required.

The above should be a safer way to achieve a WP-CLI integration with a plugin that goes beyond the boundaries of what a plugin is supposed to do.

That being said, I'd like to note, that I'd generally not recommend have this type of checkers in a plugin, but would rather recommend building it CLI-first and then maybe wrap a plugin UI around it, not the other way around as it is happening here.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #215, opened 2023-06-28
Functional tests for WP-CLI commands

swissspidy:
### Description of the Change

This PR sets up the necessary infrastructure for running Behat tests for the built-in WP-CLI commands.

Also sets up code coverage, which is quite convenient.

### How to test the Change

Run the tests locally or check out the CI build status.

### Changelog Entry

> Added - Functional tests for WP-CLI command

### Checklist:


mukeshpanchal27 (commented):
Thank you, @swissspidy, for the PR. I have reviewed it initially and would like to discuss a few questions:

- Do we have any plans to add new tests in Behat?
- Considering that Behat uses Gherkin syntax, there might be a learning curve for us. How do we plan to address this?
- Are we going to maintain both Unit tests and Behat tests simultaneously?

swissspidy:
Great questions!

> * Do we have any plans to add new tests in Behat?

Right now this PR just lays the foundation for the tests. New tickets can be created to add a few more tests. Not too many tests are needed because the command is rather straightforward and only has a few options.

> * Considering that Behat uses Gherkin syntax, there might be a learning curve for us. How do we plan to address this?

Luckily the syntax is pretty easy to grasp, making it easy to understand and write tests, plus there's plenty of documentation and tutorials out there. And of course I'm happy to do an intro as well if there's interest.

> * Are we going to maintain both Unit tests and Behat tests simultaneously?

Yes, these tests are an addition to the existing PHPUnit-based tests in the repo, focusing on how users interact with the WP-CLI command. There is little overlap. Just like e2e tests and unit/integration tests are not mutually exclusive. So I guess these could be seen as e2e tests for the WP-CLI command :-)

tests/behat/features/plugin-check.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Test that the WP-CLI command works.
+
+  Scenario: Check a non-existent plugin
```
spacedmonkey, line 3: I really wish we had a test of a plugin that works. Just testing failure here feels weird.

swissspidy: Right now this PR just lays the foundation for the tests. New tickets can be created to add a few more tests.

spacedmonkey: How can we ensure the foundation work if the tests do not do anything meaningful?

swissspidy: There was lots of manual testing of the CLI command so far, and the tests setup is standard for WP-CLI and battle-tested. Are you looking for more?

swissspidy:
@felixarntz @westonruter @joemcgill @mukeshpanchal27 Pinging you since this PR has been around for a long time but it's still valuable — this adds some initial Behat tests for the WP-CLI command added by this plugin.

We need this to ensure the command works properly, especially when making changes. Also with #153 in mind.

.github/workflows/behat-test.yml
```diff
@@ -0,0 +1,165 @@
...
+    strategy:
+      fail-fast: true
+      matrix:
+        php:
+        - '7.0'
+        - '7.1'
+        - '7.2'
+        - '7.3'
+        - '7.4'
+        - '8.1'
```
mukeshpanchal27, line 68:
```suggestion
        - '8.0'
        - '8.1'
```

.github/workflows/behat-test.yml
```diff
@@ -0,0 +1,165 @@
...
+        - '8.2'
+        wordpress: [ 'latest' ]
+        include:
+          - php: '8.0'
+            wordpress: 'latest'
+            coverage: true
+          - php: '8.0'
+            wordpress: '6.0'
+          - php: '8.0'
+            wordpress: '6.1'
```
mukeshpanchal27:
```suggestion
```

In #292, we set the minimum WordPress version for the plugin checker to be 6.3. Let's remove this old WP version check.

.github/workflows/behat-test.yml
swissspidy: 8.0 with WP latest is already added below, in the `include` section, with code coverage enabled.

.github/workflows/behat-test.yml
swissspidy: Ah right, I'll set this to 6.3 then.

mukeshpanchal27 (approved):
Thanks @swissspidy, LGTM.

tests/behat/includes/FeatureContext.php
```diff
@@ -0,0 +1,224 @@
...
+
+	/**
+	 * @When /^I (run|try) the WP-CLI command `([^`]+)`$/
+	 */
+	public function when_i_run_the_wp_cli_command( $mode, $command ) {
+		$command = "wp {$command}";
+
+		$with_code_coverage = getenv( 'BEHAT_CODE_COVERAGE' );
+		if ( \in_array( $with_code_coverage, [ true, 'true', 1, '1' ], true ) ) {
+			$command = "{$command} --require={PROJECT_DIR}/tests/behat/utils/maybe-generate-wp-cli-coverage.php";
```
felixarntz, line 93: Does this work when there is an existing `--require` argument? That would be relevant for runtime checks due to the current limitations of having to require the early logic.

felixarntz (commented):
@swissspidy LGTM, just one question before merge.

tests/behat/includes/FeatureContext.php
swissspidy: Yes, WP-CLI supports multiple `--require` arguments.

Also worth noting that this kind of `--require` usage should become obsolete anyway with #153

swissspidy: @felixarntz Please let me know if this answers your question :-)

felixarntz: Thanks, LGTM!

felixarntz (approved):

PR #219, opened 2023-06-30
Add Stability Flag

### LINKED ISSUE #92: Add Stability flag for Plugin Check CLI command
## Description

* Add two constants `STABILITY_STABLE` and `STABILITY_EXPERIMENTAL` to the `Check` interface.
* Add a public method `get_stability()` to check `Check` interface.
* Implement traits `Stable_Check` and `Experimental_Check` that simply implement the above method and return the corresponding constant value.
* Update all existing checks to use the `Stable_Check` trait (since none of our current checks is experimental).
* Add a `INCLUDE_EXPERIMENTAL` bitwise flag to the `Check_Repository` interface. Then update `Default_Check_Repository::get_checks()` to filter out experimental checks from the list _unless_ the passed `$flags` match `INCLUDE_EXPERIMENTAL`.
- Add an optional `--include-experimental` flag to the plugin check CLI command. If set, the above infrastructure should be used to also include experimental checks.

jjgrainger:
Adds a stability flag to checks and a `--include-experimental` flag for the CLI command.

Closes #92


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -156,6 +164,17 @@ final public function set_plugin( $plugin ) {
...
+	/**
+	 * Sets whether to include experimental checks in the process.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool $include_experimental True to include experimental checks. False to exclude.
+	 */
+	final public function set_experimental_flag( $include_experimental ) {
+		$this->include_experimental = $include_experimental;
+	}
```
felixarntz, line 195: The approach of having a method like this doesn't work I think, for the same reason that we had to introduce the abstract `get_check_slugs_param()` method. The way it's implemented now, the parameter wouldn't be picked up if the class is initialized early, in the `object-cache.php` drop-in.

So I think we need to instead go with an abstract method like `get_include_experimental_param() : bool`, which would need to be implemented for both the CLI and AJAX runner (for the latter we could just return `false` for now as there's no support yet for this in WP Admin / AJAX).

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -74,6 +74,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	protected $runtime_environment;
 
+	/**
+	 * Whether to include experimental checks.
+	 *
+	 * @since n.e.x.t
+	 * @var bool
+	 */
+	protected $include_experimental = false;
```
felixarntz: See my other comment, this should probably be `null` initially and then be populated via `get_include_experimental_param()` if needed.

includes/Checker/Check.php
```diff
@@ -15,6 +15,30 @@
...
+	 */
+	const STABILITY_STABLE = 1;
+
+	/**
+	 * Flag for experimental checks.
+	 *
+	 * @since n.e.x.t
+	 * @var int
+	 */
+	const STABILITY_EXPERIMENTAL = 2;
```
felixarntz: I don't think these need to be bitwise flags, as that wouldn't bring a benefit here. We can use simple strings to improve debugging (seeing a string `STABLE` will be more helpful than seeing `1`).

The constants should only exist to not rely on hard-coded strings.
```suggestion
	/**
	 * Stability value for stable checks.
	 *
	 * @since n.e.x.t
	 * @var string
	 */
	const STABILITY_STABLE = 'STABLE';

	/**
	 * Stability value for experimental checks.
	 *
	 * @since n.e.x.t
	 * @var string
	 */
	const STABILITY_EXPERIMENTAL = 'EXPERIMENTAL';
```

includes/Checker/Check.php
```diff
@@ -15,6 +15,30 @@
...
+
+	/**
+	 * Returns the checks stability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return int
+	 */
+	public function get_stability();
```
felixarntz:
```suggestion
	/**
	 * Returns the check's stability.
	 *
	 * @since n.e.x.t
	 *
	 * @return string One of the check stability constant values.
	 */
	public function get_stability();
```

includes/Checker/Experimental_Check.php
```diff
@@ -0,0 +1,24 @@
...
+ * Trait for stable checks.
+ *
+ * @since n.e.x.t
+ */
+trait Experimental_Check {
+	/**
+	 * Returns the checks stability
+	 *
+	 * @return int
+	 */
```
felixarntz: Please update the docs based on my other comment in the interface.

includes/Checker/Stable_Check.php
```diff
@@ -0,0 +1,24 @@
...
+ * Trait for stable checks.
+ *
+ * @since n.e.x.t
+ */
+trait Stable_Check {
+	/**
+	 * Returns the checks stability
+	 *
+	 * @return int
+	 */
```
felixarntz: Please update the docs based on my other comment in the interface.

includes/Checker/Stable_Check.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Trait WordPress\Plugin_Check\Checker\Stable_Check
```
felixarntz: Shouldn't these traits live in `WordPress\Plugin_Check\Traits`?

includes/Checker/Experimental_Check.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Trait WordPress\Plugin_Check\Checker\Experimental_Check
```
felixarntz: Shouldn't these traits live in `WordPress\Plugin_Check\Traits`?

includes/Checker/Default_Check_Repository.php
```diff
@@ -103,6 +103,23 @@ function ( $slug ) use ( $checks ) {
...
+				function ( $check ) {
+					if ( $check->get_stability() === Check::STABILITY_EXPERIMENTAL ) {
+						return false;
+					}
+
+					return $check;
+				},
+				$checks
+			)
+		);
```
felixarntz: I think `array_map()` here is unnecessary since you're not modifying the result?
```suggestion
		return array_filter(
			$checks,
			static function ( $check ) {
				return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
			}
		);
```

tests/phpunit/Checker/Default_Check_Repository_Tests.php
```diff
@@ -115,4 +117,80 @@ public function test_get_checks_throws_exception_for_invalid_check_slug() {
...
 	}
+
+	public function test_get_checks_returns_no_experimental_checks_by_default() {
+		$static_check       = new Static_Check();
+		$runtime_check      = new Runtime_Check();
+		$experimental_check = new Experimental_Static_Check();
+
+		$this->repository->register_check( 'static_check', $static_check );
+		$this->repository->register_check( 'runtime_check', $runtime_check );
+		$this->repository->register_check( 'experimental_static_check', $experimental_check );
```
felixarntz: Nit-pick, but since you register both `Experimental_Static_Check` and `Experimental_Runtime_Check` in all other tests, I'd suggest you do that here too. Shouldn't change the expected result, but the fact that it's currently missing here makes me wonder whether it's intentional for some reason.

felixarntz (changes requested):
@jjgrainger Thanks for implementing this, looks solid for the most part! I added a few comments below, most of which should be quick to address.

The only crucial thing to address is the way the flag is made available to `Abstract_Check_Runner`, as that currently wouldn't work well with the early initialization.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I've updated this to work similar to the other params.

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Updated

includes/Checker/Check.php
jjgrainger: Thanks @felixarntz I've made those updates

includes/Checker/Check.php
jjgrainger: Thanks @felixarntz this has been updated

includes/Checker/Stable_Check.php
jjgrainger: @felixarntz yes, I've moved these now

includes/Checker/Default_Check_Repository.php
jjgrainger: Thanks @felixarntz I've made this update

tests/phpunit/Checker/Default_Check_Repository_Tests.php
jjgrainger: Thanks @felixarntz I've included the check in the test

includes/Checker/AJAX_Runner.php
```diff
@@ -75,4 +75,15 @@ protected function get_check_slugs_param() {
 
 		return $checks;
 	}
+
+	/**
+	 * Returns the include experimental paramater based on the request.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns true to include experimental checks else false.
```
mukeshpanchal27, line 84: it will always remain false.

includes/Checker/CLI_Runner.php
```diff
@@ -88,4 +88,21 @@ protected function get_check_slugs_param() {
...
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns true to include experimental checks else false.
+	 */
+	protected function get_include_experimental_param() {
+		foreach ( $_SERVER['argv'] as $value ) {
+			if ( '--include-experimental' === $value ) {
+				return true;
+			}
+		}
```
mukeshpanchal27:
```suggestion
		if ( in_array( '--include-experimental', $_SERVER['argv'], true ) ) {
			return true;
		}
```

includes/Checker/Check.php
```diff
@@ -15,6 +15,30 @@
...
+	 * @var string
+	 */
+	const STABILITY_EXPERIMENTAL = 'EXPERIMENTAL';
+
+	/**
+	 * Returns the check's stability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string string One of the check stability constant values.
```
mukeshpanchal27:
```suggestion
	 * @return string One of the check stability constant values.
```

includes/Checker/Checks/Plugin_Review_PHPCS_Check.php
```diff
@@ -7,13 +7,16 @@
 
 namespace WordPress\Plugin_Check\Checker\Checks;
 
+use WordPress\Plugin_Check\Traits\Stable_Check;
```
mukeshpanchal27, line 10:
```suggestion
use WordPress\Plugin_Check\Traits\Stable_Check;

```

includes/Traits/Experimental_Check.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Trait WordPress\Plugin_Check\Checker\Experimental_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Traits;
+
+/**
+ * Trait for stable checks.
```
mukeshpanchal27:
```suggestion
 * Trait for experimental checks.
```

mukeshpanchal27 (commented):
Thanks @jjgrainger, Look good to me. Left some feedback.

includes/Checker/AJAX_Runner.php
felixarntz: @mukeshpanchal27 That's temporary though, so better not to document it like that as it could change in the future. Also since this is an abstract method from a parent class, we shouldn't change the documentation here IMO.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -74,6 +74,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 */
 	protected $runtime_environment;
 
+	/**
+	 * Whether to include experimental checks.
+	 *
+	 * @since n.e.x.t
+	 * @var bool
+	 */
+	protected $include_experimental = null;
```
felixarntz: Nit-pick, but the other similar class properties aren't initialized like this, so we can remove it from here too. Uninitialized class properties are `null` anyway.
```suggestion
	protected $include_experimental;
```

includes/Traits/Experimental_Check.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @since n.e.x.t
+ */
+trait Experimental_Check {
+	/**
+	 * Returns the checks stability.
+	 *
+	 * @return string
```
felixarntz: See https://github.com/10up/plugin-check/pull/219#discussion_r1248146160:
```suggestion
	 * Returns the check's stability.
	 *
	 * @since n.e.x.t
	 *
	 * @return string One of the check stability constant values.
```

includes/Traits/Stable_Check.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Trait for stable checks.
+ *
+ * @since n.e.x.t
+ */
+trait Stable_Check {
+	/**
+	 * Returns the checks stability.
+	 *
+	 * @return string
```
felixarntz:
```suggestion
	 * Returns the check's stability.
	 *
	 * @since n.e.x.t
	 *
	 * @return string One of the check stability constant values.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -124,6 +126,7 @@ public function check( $args, $assoc_args ) {
 
 		$checks_to_run = array();
 		try {
+			$runner->set_experimental_flag( $options['include-experimental'] );
```
felixarntz, line 129: Just double-checking, will this always be set? Have you tested that? Unsure about how WP-CLI populates these values for boolean arguments, so I'd like us to verify this doesn't cause a PHP warning if the argument is not provided.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -287,6 +325,11 @@ final public function get_checks_to_run() {
 			$check_flags = Check_Repository::TYPE_ALL;
 		}
 
+		// Check whether to include experimental checks.
+		if ( $this->include_experimental ) {
```
felixarntz: This doesn't work reliably: For the same reason we call the `get_check_slugs()` method above instead of accessing `$this->check_slugs`, we have to call a method that ensures the parameter is retrieved instead of directly accessing `$this->include_experimental`. Otherwise this may still return `null`.

We need to introduce a `get_include_experimental()` method which works equivalent to the `get_check_slugs()` method: If the class property isn't `null`, return that, otherwise return the result of `get_include_experimental_param()`.

felixarntz (commented):
@jjgrainger One crucial change left here in the `Abstract_Check_Runner` class, all other comments are non-blocking nit-picks.

includes/CLI/Plugin_Check_Command.php
jjgrainger: @felixarntz we parse the options with [`get_options`](https://github.com/10up/plugin-check/blob/b2fb12993c6bad7147f6f2466a4c0cea93963743/includes/CLI/Plugin_Check_Command.php#L209-L215) method to make sure the default of false is included. We do the same for `--ignore-errors` and `--ignore-warnings` flags too.

includes/Checker/Experimental_Check.php
jjgrainger: This has been updated

includes/Checker/Stable_Check.php
jjgrainger: This has been updated

includes/Checker/CLI_Runner.php
jjgrainger: Thanks @mukeshpanchal27 updated

includes/Checker/Check.php
jjgrainger: Thanks @mukeshpanchal27 this has been updated

includes/Checker/Checks/Plugin_Review_PHPCS_Check.php
jjgrainger: Thanks @mukeshpanchal27 updated

includes/Traits/Experimental_Check.php
jjgrainger: Thanks @mukeshpanchal27 updated

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I've updated this

includes/Traits/Experimental_Check.php
jjgrainger: Thanks @felixarntz updated

includes/Traits/Stable_Check.php
jjgrainger: Thanks @felixarntz updated

includes/Checker/Abstract_Check_Runner.php
jjgrainger: Thanks @felixarntz I've added the new `get_include_experimental` method

includes/CLI/Plugin_Check_Command.php
felixarntz: Thanks @jjgrainger!

felixarntz (approved):
Thanks @jjgrainger, LGTM!

includes/Traits/Experimental_Check.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Trait for experimental checks.
+ *
+ * @since n.e.x.t
+ */
+trait Experimental_Check {
+	/**
+	 * Returns the checks stability.
+	 *
+	 * @return string One of the check stability constant values.
```
mukeshpanchal27:
```suggestion
	 * Returns the checks stability.
	 *
	 * @since n.e.x.t
	 *
	 * @return string One of the check stability constant values.
```

includes/Traits/Stable_Check.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Trait for stable checks.
+ *
+ * @since n.e.x.t
+ */
+trait Stable_Check {
+	/**
+	 * Returns the checks stability.
+	 *
+	 * @return string One of the check stability constant values.
```
mukeshpanchal27:
```suggestion
	 * Returns the checks stability.
	 *
	 * @since n.e.x.t
	 *
	 * @return string One of the check stability constant values.
```

mukeshpanchal27 (approved):
Thanks @jjgrainger for the changes. Left final nitpick and aprove.

Issue #206, opened 2023-06-22
Allow running only checks of certain categories in WP Admin & WP-CLI


felixarntz:
### Is your enhancement related to a problem? Please describe.

⚠️  **Follow up to / blocked by #203**

Once the foundation for check categories has been implemented, this issue here should be used to expose (optionally) setting one or more categories to run checks for in WP Admin and WP-CLI.

#### For WP-CLI
* Add a new optional parameter `--categories` which expects a comma separated list of category slugs.
* If passed, only checks that fall in one of the categories should be run. Otherwise (default), all checks should be run.

#### For WP Admin
* Add a new `fieldset` with checkboxes for every available category to the WP Admin screen. By default, all of them should be checked.
* The categories checked should be passed along in the AJAX request to get the checks to run, and only checks that fall in one of the categories should be returned.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

felixarntz:
cc @spacedmonkey as this is something you were interested in

jjgrainger:
Thanks @felixarntz AC's look good to me

mukeshpanchal27:
Thanks @felixarntz AC's look good to me. Do we have any UI design for the WP Admin or we can make it simple? cc. @eclarke1

felixarntz:
@mukeshpanchal27 I'd suggest we stick with a regular `fieldset` and check boxes, let's look at some of the relevant WP core settings screens that have such a UI element and style them in a similar way.

PR #220, opened 2023-07-03
Add basic infrastructure for category

### LINKED ISSUE #203: Require every check to have a category

Since there are already at this point different kinds of checks available in the plugin checker, it's going to be crucial to be able to only run specific kinds of checks for specific use-cases. For example, the plugin review team may only be interested in plugin minimum requirement checks for the wordpress.org repository. On the flipside, someone building a plugin for a client may be interested in all checks except that one. There may be people that want to only run security checks. Etc

Right now, you can only run specific checks by passing the check slugs, however as the number of checks will grow further, this will soon become unreasonable as an approach (if it isn't already). We should therefore introduce the concept of **check categories**: Every check should have a category, from a group of predefined categories, and there should be a mechanism to only run checks of a certain category.

This issue should be focused only on implementing the basic infrastructure for categories:
* A place where all available categories are defined
* A modification to the check interface to ensure every check has a category attached
* A way to filter a list of checks to by certain categories.

A follow-up issue should then be responsible for offering a way to actually use this infrastructure via a new optional CLI argument and a new optional UI element in WP Admin.

### Designs

* Introduce a `Check_Categories` class.
* Add class constants (`const`) to the class for all available categories (see below).
* Implement a public static method `get_categories()` which simply returns the array of all categories.
* Implement a public static method `filter_checks_by_categories( array $checks, array $categories )` that filters a given list of check instances to only return those that have one of the provided categories.

For the way that checks declare their category, we should add a public method `get_category()` to the `Check` interface.
* Usage example: `$check->get_category()`
* Implementation example:
```php
class Code_Obfuscation_Check {
    public function get_category() {
        return Check_Categories::CATEGORY_PLUGIN_REPO;
    }
}
```

For the available categories, I think we can start with the following:
* `CATEGORY_GENERAL`
* `CATEGORY_PLUGIN_REPO`
* `CATEGORY_SECURITY`
* `CATEGORY_PERFORMANCE`
* `CATEGORY_ACCESSIBILITY`

### Alternatives considered

For checks to express their category, we could alternatively add a class constant like `CATEGORY` to the `Check interface and set it to some "default" category. Then every check has to define its own override of that class constant.
* Usage example: `Code_Obfuscation_Check::CATEGORY`
* Implementation example:
```php
class Code_Obfuscation_Check {
    const CATEGORY = Check_Categories::CATEGORY_PLUGIN_REPO;
}
```
The benefit would be a bit less code to write for every check (a constant is less than a method). However, this would not work for checks that use separate instances of the same class, which e.g. is relevant for #204. Therefore this is not a good option.

### Code of Conduct

mukeshpanchal27:
Closes #203 

### Checklist:

- [x] All existing tests pass.


includes/Checker/Check.php
```diff
@@ -27,4 +27,11 @@ interface Check {
 	 *                   the check).
 	 */
 	public function run( Check_Result $result );
+
+	/**
+	 * Gets the category for the check.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function get_category();
```
swissspidy: Nit: all the new methods introduced in this PR should have proper `@return` annotations. Once we bump the PHPStan strictness level this will be mandatory.

mukeshpanchal27: Thanks @swissspidy for the feedback. In https://github.com/10up/plugin-check/pull/220/commits/b6a5d2bd849c6e9047a56ea75d13926c877e6324 i have added missing return for the functions.

includes/Checker/Checks/Enqueued_Scripts_In_Footer_Check.php
```diff
@@ -7,13 +7,26 @@
...
 
+	/**
+	 * Gets the category of the check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The category of the check.
+	 */
+	public function get_category() {
+		return Check_Categories::CATEGORY_GENERAL;
```
felixarntz:
```suggestion
		return Check_Categories::CATEGORY_PERFORMANCE;
```

includes/Checker/Checks/Enqueued_Scripts_Size_Check.php
```diff
@@ -49,6 +50,17 @@ public function __construct( $threshold_size = 300000 ) {
...
 
+	/**
+	 * Gets the category of the check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The category of the check.
+	 */
+	public function get_category() {
+		return Check_Categories::CATEGORY_GENERAL;
```
felixarntz:
```suggestion
		return Check_Categories::CATEGORY_PERFORMANCE;
```

includes/Checker/Checks/Enqueued_Styles_Scope_Check.php
```diff
@@ -46,6 +47,17 @@ class Enqueued_Styles_Scope_Check extends Abstract_Runtime_Check implements With
...
 
+	/**
+	 * Gets the category of the check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The category of the check.
+	 */
+	public function get_category() {
+		return Check_Categories::CATEGORY_GENERAL;
```
felixarntz:
```suggestion
		return Check_Categories::CATEGORY_PERFORMANCE;
```

includes/Checker/Checks/Localhost_Check.php
```diff
@@ -16,6 +17,17 @@
...
 
+	/**
+	 * Gets the category of the check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The category of the check.
+	 */
+	public function get_category() {
+		return Check_Categories::CATEGORY_SECURITY;
```
felixarntz:
```suggestion
		return Check_Categories::CATEGORY_PLUGIN_REPO;
```

includes/Checker/Checks/No_Unfiltered_Uploads_Check.php
```diff
@@ -16,6 +17,17 @@
...
 
+	/**
+	 * Gets the category of the check.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The category of the check.
+	 */
+	public function get_category() {
+		return Check_Categories::CATEGORY_SECURITY;
```
felixarntz:
```suggestion
		return Check_Categories::CATEGORY_PLUGIN_REPO;
```

includes/Checker/Check_Categories.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+namespace WordPress\Plugin_Check\Checker;
+
+/**
+ * Abstract Check Runner class.
+ *
+ * @since n.e.x.t
+ */
+class Check_Categories {
```
felixarntz, line 15: Can you add PHPUnit test coverage for this class?

For `get_categories()`, you can probably use a simple check based on https://www.php.net/manual/en/reflectionclass.getconstants.php to ensure the returned array includes all the `CATEGORY_*` class constants.

For `filter_checks_by_categories()`, you can maybe use a data provider to cover multiple scenarios and ensure the expected checks are filtered based on the categories passed.

felixarntz (changes requested):
@mukeshpanchal27 Looks great so far. My only feedback are some category adjustments, and we should add unit tests for the new class.

includes/Checker/Check_Categories.php
swissspidy: We could also just update `get_categories()` to return all class constants. Something like:

```php
public function get_categories() {
		$consts = ( new ReflectionClass( $this ) )->getConstants();

		/**
		 * List of categories.
		 *
		 * @var string[] $categories
		 */
		$categories = array_values(
			array_filter(
				$consts,
				static fn( $key ) => str_starts_with( $key, 'CATEGORY_' ),
				ARRAY_FILTER_USE_KEY
			)
		);

		return $categories;
	}

Comment (felixarntz) [reply includes/Checker/Check_Categories.php:15]: Good idea. In that case maybe we should "cache" that in a static variable though to avoid recomputing this every time `get_categories()` is called?

Comment (mukeshpanchal27) [reply includes/Checker/Check_Categories.php:15]: Thank you both for the feedback. In https://github.com/10up/plugin-check/pull/220/commits/caa2247b5d254c860b65b47d7b65f7bb3438b667 and https://github.com/10up/plugin-check/pull/220/commits/a5a67c7153ebb43f5d82203e6b810a71446a3b0f i have address the review feedback and add unit tests.

File: tests/phpunit/Checker/Check_Categories_Tests.php [hunk 8]
```diff
@@ -0,0 +1,130 @@
...
+		$check_categories = new Check_Categories();
+		$categories       = $check_categories->get_categories();
+
+		// Assert that all the CATEGORY_* constants are included in the returned categories array.
+		$this->assertContains( Check_Categories::CATEGORY_GENERAL, $categories );
+		$this->assertContains( Check_Categories::CATEGORY_PLUGIN_REPO, $categories );
+		$this->assertContains( Check_Categories::CATEGORY_SECURITY, $categories );
+		$this->assertContains( Check_Categories::CATEGORY_PERFORMANCE, $categories );
+		$this->assertContains( Check_Categories::CATEGORY_ACCESSIBILITY, $categories );
+	}
```
felixarntz: Instead of hard-coding all constants here, let's use `ReflectionClass::getConstants()`. Otherwise this test is error-prone, as someone may forget to add a new category here.

tests/phpunit/Checker/Check_Categories_Tests.php
```diff
@@ -0,0 +1,130 @@
...
+
+	public function test_filter_checks_by_categories() {
```
felixarntz: Can you please combine all the `test_filter_checks_*()` methods into a single test method using a data provider? There is a lot of unnecessary duplicate code now, where you could use parameters instead.

For example, have parameters `array $categories, array $all_checks, array $expected_filtered_checks`. Then the data provider method can return those for the 3 different test scenarios you have here.

felixarntz (commented):
@mukeshpanchal27 A bit of feedback on the added tests below.

tests/phpunit/Checker/Check_Categories_Tests.php
mukeshpanchal27: Thank you, @felixarntz, for your assistance. Previously, I attempted to add a test using a data provider. However, I encountered an error stating `Class 'WordPress\Plugin_Check\Test_Data\Category_Check_One' not found` within the data provider. Consequently, I have now included the necessary class files by adding a "require" statement in the data provider.

tests/phpunit/Checker/Check_Categories_Tests.php
```diff
@@ -0,0 +1,124 @@
...
+
+		$checks = $this->repository->get_checks();
+
+		$check_categories = new Check_Categories();
+		$filtered_checks  = $check_categories->filter_checks_by_categories( $checks, $categories );
+
+		$this->assertEquals( $expected_filtered_checks, $filtered_checks );
+	}
+
+	public function filterChecksByCategoriesDataProvider() {
```
swissspidy: Nit: we don't use camelCase elsewhere. And other data providers typically have a `data_` prefix, e.g. `data_forbidden_file_types`. I'd suggest renaming this to `data_checks_by_categories`

mukeshpanchal27: Nice catch. Don't know how i mis this.

felixarntz (commented):
Thanks @mukeshpanchal27, code-wise this looks good to go now. However, there has been some discussion on the idea whether to support multiple categories per check, and we're leaning towards doing that, see https://github.com/10up/plugin-check/issues/203#issuecomment-1625690563.

Could you update this PR accordingly? It's mostly about applying the same change throughout, to change `get_category() : string` to `get_categories() : array`. You could then in all checks return an array with the one category that is currently there.

mukeshpanchal27:
@felixarntz @spacedmonkey @joemcgill In https://github.com/10up/plugin-check/pull/220/commits/2a444b4f5a095420a44e1152b329c0f5359dcb14 i have added support for multiple category. Could you please take a look so i can start working on blocker #206 issue?

includes/Checker/Check_Categories.php
```diff
@@ -0,0 +1,72 @@
...
+		return $categories;
+	}
+
+	/**
+	 * Returns an array of checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $checks     An array of Check instances.
+	 * @param array $categories An array of available categories.
```
felixarntz:
```suggestion
	 * @param array $categories An array of categories to filter by.
```

includes/Checker/Check_Categories.php
```diff
@@ -0,0 +1,72 @@
...
+	 * @return array Filtered $checks list.
+	 */
+	public static function filter_checks_by_categories( array $checks, array $categories ) {
+		return array_filter(
+			$checks,
+			static function( $check ) use ( $categories ) {
+				return array_intersect( $check->get_categories(), $categories );
```
felixarntz: This is a clever implementation, maybe clarify with a comment what its purpose is? It may also help cast it to a bool just for clarification.
```suggestion
				// Return true if at least one of the check categories is among the filter categories.
				return (bool) array_intersect( $check->get_categories(), $categories );
```

felixarntz (approved):

Two non-blocking nit-picks below.

jjgrainger (approved):
Thanks @mukeshpanchal27 approved but recommend addressing the last few nit-picks and updating the branch with trunk

PR #227, opened 2023-07-10
Validate that every check has at least one category assigned

felixarntz:
### Description of the Change

Follow-up to #220 (see #203): Given that we changed the definition of the interface to allow returning multiple categories, we should still ensure every check has at least one category. When using a string, that was mandated by the interface, but because we're now using an array, it technically allows returning an empty array.

This PR ensures at least one category is associated with every check as follows:

* Throw an exception when a registered check doesn't have any categories.
* Document on the method that at least one category will be returned.

### Checklist:

### DIFF
--- a/includes/Checker/Default_Check_Repository.php
+++ b/includes/Checker/Default_Check_Repository.php
@@ -44,7 +44,13 @@ class Default_Check_Repository implements Check_Repository {
 	 */
 	public function register_check( $slug, Check $check ) {
 		if ( ! $check instanceof Runtime_Check && ! $check instanceof Static_Check ) {
-			throw new Exception( __( 'Check must be an instance of Runtime_Check or Static_Check.', 'plugin-check' ) );
+			throw new Exception(
+				sprintf(
+					/* translators: %s: The Check slug. */
+					__( 'Check with slug "%s" must be an instance of Runtime_Check or Static_Check.', 'plugin-check' ),
+					$slug
+				)
+			);
 		}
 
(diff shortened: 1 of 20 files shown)


felixarntz:
@mukeshpanchal27 @jjgrainger This is just a quick follow-up to #220 that I realized after merging. This was first mentioned in https://github.com/10up/plugin-check/issues/203#issuecomment-1625690563.

includes/Checker/Default_Check_Repository.php
```diff
@@ -44,7 +44,13 @@ class Default_Check_Repository implements Check_Repository {
...
 	public function register_check( $slug, Check $check ) {
 		if ( ! $check instanceof Runtime_Check && ! $check instanceof Static_Check ) {
-			throw new Exception( __( 'Check must be an instance of Runtime_Check or Static_Check.', 'plugin-check' ) );
+			throw new Exception(
+				sprintf(
+					/* translators: %s: The Check slug. */
+					__( 'Check with slug "%s" must be an instance of Runtime_Check or Static_Check.', 'plugin-check' ),
+					$slug
+				)
+			);
```
felixarntz, line 53: This change is somewhat unrelated, but it adds the slug to the message to make the exception message more helpful. Without it, you may not know which check caused the exception.

jjgrainger (approved):
Thanks @felixarntz LGTM 👍

mukeshpanchal27 (approved):

PR #228, opened 2023-07-12
Allow running only checks of certain categories in WP Admin & WP-CLI

### LINKED ISSUE #206: Allow running only checks of certain categories in WP Admin & WP-CLI

mukeshpanchal27:
Closes #206

**WP Admin screen:**
Screenshot 2023-07-12 at 2 51 02 PM

**WP_CLI command**
`npm run wp-env run cli -- wp plugin check {plugin} --require=./wp-content/plugins/plugin-check/cli.php --categories=security,general`

### Checklist:

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -16,7 +16,7 @@
 		! pluginsList ||
 		! resultsContainer ||
 		! spinner ||
-		! categoriesList
+		! categoriesList.length
 	) {
 		console.error( 'Missing form elements on page' );
 		return;
@@ -102,10 +102,6 @@
 			'action',
 			pluginCheck.actionSetUpRuntimeEnvironment
 		);
-
-		for ( let i = 0; i < data.categories.length; i++ ) {
-			pluginCheckData.append( 'categories[]', data.categories[ i ] );
-		}
 
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			pluginCheckData.append( 'checks[]', data.checks[ i ] );
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -5,11 +5,19 @@
 	const pluginsList = document.getElementById(
 		'plugin-check__plugins-dropdown'
 	);
-	const categoriesList = document.querySelectorAll( 'input[name=categories]' );
+	const categoriesList = document.querySelectorAll(
+		'input[name=categories]'
+	);
 	const templates = {};
 
 	// Return early if the elements cannot be found on the page.
-	if ( ! checkItButton || ! pluginsList || ! resultsContainer || ! spinner || ! categoriesList ) {
+	if (
+		! checkItButton ||
+		! pluginsList ||
+		! resultsContainer ||
+		! spinner ||
+		! categoriesList
+	) {
 		console.error( 'Missing form elements on page' );
 		return;
 	}
@@ -169,7 +177,10 @@
 
 		for ( let i = 0; i < categoriesList.length; i++ ) {
 			if ( categoriesList[ i ].checked ) {
-				pluginCheckData.append( 'categories[]', categoriesList[ i ].value );
+				pluginCheckData.append(
+					'categories[]',
+					categoriesList[ i ].value
+				);
 			}
 		}
 
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -9,7 +9,7 @@
 	const templates = {};
 
 	// Return early if the elements cannot be found on the page.
-	if ( ! checkItButton || ! pluginsList || ! resultsContainer || ! spinner ) {
+	if ( ! checkItButton || ! pluginsList || ! resultsContainer || ! spinner || ! categoriesList ) {
 		console.error( 'Missing form elements on page' );
 		return;
 	}
@@ -95,9 +95,9 @@
 			pluginCheck.actionSetUpRuntimeEnvironment
 		);
 
-			for ( let i = 0; i < data.categories.length; i++ ) {
-				pluginCheckData.append( 'categories[]', data.categories[ i ] );
-			}
+		for ( let i = 0; i < data.categories.length; i++ ) {
+			pluginCheckData.append( 'categories[]', data.categories[ i ] );
+		}
 
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			pluginCheckData.append( 'checks[]', data.checks[ i ] );
@@ -168,7 +168,7 @@
 		pluginCheckData.append( 'action', pluginCheck.actionGetChecksToRun );
 
 		for ( let i = 0; i < categoriesList.length; i++ ) {
-			if ( categoriesList[i].checked ) {
+			if ( categoriesList[ i ].checked ) {
 				pluginCheckData.append( 'categories[]', categoriesList[ i ].value );
 			}
 		}
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -5,6 +5,7 @@
 	const pluginsList = document.getElementById(
 		'plugin-check__plugins-dropdown'
 	);
+	const categoriesList = document.querySelectorAll( 'input[name=categories]' );
 	const templates = {};
 
 	// Return early if the elements cannot be found on the page.
@@ -34,6 +35,9 @@
 		checkItButton.disabled = true;
 		pluginsList.disabled = true;
 		spinner.classList.add( 'is-active' );
+		for ( let i = 0; i < categoriesList.length; i++ ) {
+			categoriesList[ i ].disabled = true;
+		}
 
 		getChecksToRun()
 			.then( setUpEnvironment )
@@ -70,6 +74,9 @@
 		spinner.classList.remove( 'is-active' );
 		checkItButton.disabled = false;
 		pluginsList.disabled = false;
+		for ( let i = 0; i < categoriesList.length; i++ ) {
+			categoriesList[ i ].disabled = false;
+		}
 	}
 
 	/**
@@ -87,6 +94,10 @@
 			'action',
 			pluginCheck.actionSetUpRuntimeEnvironment
 		);
+
+			for ( let i = 0; i < data.categories.length; i++ ) {
+				pluginCheckData.append( 'categories[]', data.categories[ i ] );
+			}
 
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			pluginCheckData.append( 'checks[]', data.checks[ i ] );
@@ -155,6 +166,12 @@
 		pluginCheckData.append( 'nonce', pluginCheck.nonce );
 		pluginCheckData.append( 'plugin', pluginsList.value );
 		pluginCheckData.append( 'action', pluginCheck.actionGetChecksToRun );
+
+		for ( let i = 0; i < categoriesList.length; i++ ) {
+			if ( categoriesList[i].checked ) {
+				pluginCheckData.append( 'categories[]', categoriesList[ i ].value );
+			}
+		}
 
 		return fetch( ajaxurl, {
 			method: 'POST',
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -108,12 +108,17 @@ final class Admin_AJAX {
 			);
 		}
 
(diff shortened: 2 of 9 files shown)


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -194,6 +211,24 @@ final public function set_experimental_flag( $include_experimental ) {
...
+	 * @since n.e.x.t
+	 *
+	 * @param array $categories An array of categories for filtering.
+	 *
+	 * @throws Exception Thrown if the getegories does not match the original request parameter.
+	 */
+	final public function filter_checks_by_specific_categories( $categories ) {
+		if ( $categories !== $this->get_categories_param() ) {
+			throw new Exception(
+				__( 'Invalid flag: The include-experimental flag does not match the original request parameter.', 'plugin-check' )
```
swissspidy: I'd recommend using `sprintf()` here for `include-experimental` so that translators don't have to translate this piece.

```php
sprintf(
	/* translators: %s: include-experimental */
	__( 'Invalid flag: The %s flag does not match the original request parameter.', 'plugin-check' ),
	'include-experimental'
)
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -186,14 +205,40 @@ final public function set_experimental_flag( $include_experimental ) {
...
+	 * @param array $categories An array of categories for filtering.
+	 *
+	 * @throws Exception Thrown if the getegories does not match the original request parameter.
+	 */
+	final public function filter_checks_by_specific_categories( $categories ) {
+		if ( $categories !== $this->get_categories_param() ) {
+			throw new Exception(
+				sprintf(
+					/* translators: %s: categories */
+					__( 'Invalid categories: The %s does not match the original request parameter.', 'plugin-check' ),
```
swissspidy:
```suggestion
					__( 'Invalid categories: The %s flag does not match the original request parameter.', 'plugin-check' ),
```

includes/Checker/Abstract_Check_Runner.php
mukeshpanchal27: Thanks @swissspidy, Updated in https://github.com/10up/plugin-check/pull/228/commits/5c679dc04e4ae6259a95ec6f07ddfb7a6fe6efc5

swissspidy (approved):
Looks like it does the trick! 👍

jjgrainger (approved):
Thanks @mukeshpanchal27 LGTM

includes/Admin/Admin_AJAX.php
```diff
@@ -108,12 +108,17 @@ public function set_up_environment() {
...
+		$categories = filter_input( INPUT_POST, 'categories', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
+		$categories = is_null( $categories ) ? array() : $categories;
+		$checks     = filter_input( INPUT_POST, 'checks', FILTER_DEFAULT, FILTER_FORCE_ARRAY );
+		$checks     = is_null( $checks ) ? array() : $checks;
+		$plugin     = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
 		try {
 			$runner->set_check_slugs( $checks );
 			$runner->set_plugin( $plugin );
+			$runner->filter_checks_by_specific_categories( $categories );
```
felixarntz: These changes look like they're happening in the wrong place. The categories are only important to filter the checks to run, which should happen in the `get_checks_to_run()` AJAX request, which is the very first one. Subsequent requests should only receive the actual list of checks, so the categories shouldn't be relevant for those requests anymore.

I think the changes in this method should be reverted, since at this point we should already know the proper list of checks.

includes/Admin/Admin_AJAX.php
```diff
@@ -221,8 +228,9 @@ public function get_checks_to_run() {
 
 		wp_send_json_success(
 			array(
-				'plugin' => $plugin_basename,
-				'checks' => array_keys( $checks_to_run ),
+				'plugin'     => $plugin_basename,
+				'checks'     => array_keys( $checks_to_run ),
+				'categories' => $categories,
```
felixarntz: This change should be unnecessary, the categories shouldn't be returned from here. The purpose of the `get_checks_to_run()` AJAX callback is to return the full list of checks to run, which should already be filtered by the categories per your logic added to `Abstract_Check_Runner::get_checks_to_run()`. So returning the categories isn't needed.

Instead, in the `try catch` clause right above here, you need to set the check categories in the runner via the method for that (which currently has the misleading name `filter_checks_by_specific_categories`, it should really be called `set_categories`, as mentioned in my other comment).

includes/Admin/Admin_Page.php
```diff
@@ -150,12 +152,15 @@ private function get_available_plugins() {
...
 	public function render_page() {
-		global $available_plugins, $selected_plugin_basename;
+		global $available_plugins, $selected_plugin_basename, $categories;
 
 		$available_plugins = $this->get_available_plugins();
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
+		$check_categories = new Check_Categories();
+		$categories       = $check_categories->get_categories();
```
felixarntz: Static methods don't require an instance.
```suggestion
		$categories = Check_Categories::get_categories();
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -186,14 +205,40 @@ final public function set_experimental_flag( $include_experimental ) {
...
+	/**
+	 * Sets categories for filtering the checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $categories An array of categories for filtering.
+	 *
+	 * @throws Exception Thrown if the getegories does not match the original request parameter.
+	 */
+	final public function filter_checks_by_specific_categories( $categories ) {
```
felixarntz: The method name here is not right. This sets the check categories, it doesn't filter anything. It should be called `set_categories()`.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -186,14 +205,40 @@ final public function set_experimental_flag( $include_experimental ) {
...
+		if ( $categories !== $this->get_categories_param() ) {
```
felixarntz: Similar to the other similar methods, this check should only be made if the class instance was initialized early.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -186,14 +205,40 @@ final public function set_experimental_flag( $include_experimental ) {
 		if ( $this->initialized_early ) {
 			if ( $include_experimental !== $this->get_include_experimental_param() ) {
 				throw new Exception(
-					__( 'Invalid flag: The include-experimental flag does not match the original request parameter.', 'plugin-check' )
+					sprintf(
+						/* translators: %s: include-experimental */
+						__( 'Invalid flag: The %s flag does not match the original request parameter.', 'plugin-check' ),
```
felixarntz: Good idea to consolidate the strings, just one rewording suggestion:
```suggestion
						__( 'Invalid flag: The %s value does not match the original request parameter.', 'plugin-check' ),
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -186,14 +205,40 @@ final public function set_experimental_flag( $include_experimental ) {
...
+	 * @param array $categories An array of categories for filtering.
+	 *
+	 * @throws Exception Thrown if the getegories does not match the original request parameter.
+	 */
+	final public function filter_checks_by_specific_categories( $categories ) {
+		if ( $categories !== $this->get_categories_param() ) {
+			throw new Exception(
+				sprintf(
+					/* translators: %s: categories */
+					__( 'Invalid categories: The %s flag does not match the original request parameter.', 'plugin-check' ),
```
felixarntz:
```suggestion
					__( 'Invalid categories: The %s value does not match the original request parameter.', 'plugin-check' ),
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -330,7 +375,16 @@ final public function get_checks_to_run() {
...
 		}
 
-		return $this->check_repository->get_checks( $check_flags, $check_slugs );
+		$checks = $this->check_repository->get_checks( $check_flags, $check_slugs );
+
+		// Filters the checks by specific categories.
+		$get_categories = $this->get_categories();
+		if ( $get_categories ) {
+			$check_categories = new Check_Categories();
+			$checks           = $check_categories->filter_checks_by_categories( $checks, $get_categories );
```
felixarntz: A bit of an odd variable name, why not just `$categories`? Also, similar to the above, no need to instantiate the class when the methods are static.
```suggestion
		$categories = $this->get_categories();
		if ( $categories ) {
			$checks = Check_Categories::filter_checks_by_categories( $checks, $categories );
```

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -398,6 +452,21 @@ final public function get_include_experimental() {
...
 	}
 
+	/**
+	 * Returns an array of categories for filtering the checks.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of categories.
+	 */
+	final public function get_categories() {
```
felixarntz: This method should be protected, there's no public usage of it.

I missed that in the `get_include_experimental()` method above, could you update that to become protected as well?
```suggestion
	final protected function get_categories() {
```

includes/Checker/CLI_Runner.php
```diff
@@ -103,4 +103,24 @@ protected function get_include_experimental_param() {
...
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of categories.
+	 */
+	protected function get_categories_param() {
+		$categories = array();
+
+		foreach ( $_SERVER['argv'] as $value ) {
+			if ( false !== strpos( $value, '--categories=' ) ) {
+				$categories = explode( ',', str_replace( '--categories=', '', $value ) );
```
felixarntz: You use `wp_parse_list()` above, so the same function should be used here as well instead of manual `explode()`.

assets/js/plugin-check-admin.js
```diff
@@ -88,6 +103,10 @@
 			pluginCheck.actionSetUpRuntimeEnvironment
 		);
 
+		for ( let i = 0; i < data.categories.length; i++ ) {
+			pluginCheckData.append( 'categories[]', data.categories[ i ] );
+		}
+
```
felixarntz: See my other comment, this should be removed. The proper list of checks is already known at this point and passed below, so we don't need to pass categories anymore.

assets/js/plugin-check-admin.js
```diff
@@ -5,10 +5,19 @@
...
 	const templates = {};
 
 	// Return early if the elements cannot be found on the page.
-	if ( ! checkItButton || ! pluginsList || ! resultsContainer || ! spinner ) {
+	if (
+		! checkItButton ||
+		! pluginsList ||
+		! resultsContainer ||
+		! spinner ||
+		! categoriesList
```
felixarntz: Since this is an array, I think this needs to rather check for the length.
```suggestion
		! categoriesList.length
```

felixarntz (changes requested):
@mukeshpanchal27 This looks good UI-wise, but there are some problems with how the AJAX flow is implemented. The categories should only be relevant to pass to the first request, which is to get the checks to run. After that, we already know which checks to run, so the other AJAX requests don't need to receive the categories anymore.

mukeshpanchal27:
Thank you, @felixarntz, for your feedback. I have made the necessary changes in https://github.com/10up/plugin-check/pull/228/commits/b42cb528fc1f759a14251dd98d615cd2018e043a. Could you please review the updated version?

felixarntz (approved):
@mukeshpanchal27 Great work, LGTM!

Issue #194, opened 2023-06-15
Plugin checker QA testing


mukeshpanchal27:
### Is your enhancement related to a problem? Please describe.

QA testing will review the user interfaces for the Plugin Checker plugin. This will involve testing both the WP Admin interface and the CLI command.

Testing will be done against a number of different scenarios to ensure the plugin and checks work correctly under different circumstances. As well as identifying any bugs within the plugin, QA should also make recommendations on any improvements that can be made to accessibility and overall usability.

## Scenarios

The following scenarios outline the different environment setups to be tested against. For each of the scenarios below there should be 2 tests, 1) where the Test Plugin contains errors/warnings that the check will pick up, 2) where the Test Plugin contains no errors/warnings.

1. **Minimal Plugins Installed / Test Plugin inactive** - This is the most minimal environment to test against. There should only be the Plugin Checker and Test plugin installed. In this scenario only static checks should be run.
2. **Minimal Plugins Installed / Test Plugin active** - Same as the Scenario 1, except the Test Plugin is active. In this scenario all checks (both static and runtime) should be run.
3. **Minimal Plugins Installed / Test([Performance Lab](https://wordpress.org/plugins/performance-lab/)) Plugin active** - Same as the Scenario 1, except the Test Plugin is active. In this scenario only static checks should be run because PL plugin override object cache so runtime check will be skipped in this case.
4. **Multiple Plugins installed / No plugins active** - This scenario includes additional plugins that won't be checked by the plugin checker. In this scenario, only *static* checks should run against the Test Plugin and no others.
5. **Multiple Plugins installed / Only test plugin active** - Same as Scenario 3, but this time both static and runtime checks should be run.
6. **Multiple Plugins installed / Only test([Performance Lab](https://wordpress.org/plugins/performance-lab/)) plugin active** - Same as Scenario 3, but this time static checks should be run because PL plugin override object cache so runtime check will be skipped in this case.
7. **Multiple Plugins installed / All plugins active** - Same as Scenario 4, but checks/errors are only run against the Test Plugin.

Note: There is one known issue that we are already tracking for the Elementor plugin. Please refer to https://github.com/10up/plugin-check/issues/185. If you encounter this issue, please ignore it. Thank you!

## Test Plugins
The following plugin was custom developed by us for testing purposes only.

[...]

PR #236, opened 2023-07-25
Plugin Checker Deactivation Causes Error in CLI Command

mukeshpanchal27:
During QA testing, it was discovered that if the Plugin Checker is deactivated and the CLI command `npm run wp-env run cli -- wp plugin check {plugin} --require=./wp-content/plugins/plugin-check/cli.php` is executed, it shows an error. The error is related to the behaviour discussed in the following comment on GitHub: https://github.com/10up/plugin-check/issues/194#issuecomment-1647538986. As a result of this finding, the decision was made to throw an error when anyone attempts to use the CLI command while the Plugin Checker is not activated.

**Proposed Solution:**
To address the issue, the possible code changes need to be added to the `cli.php` file. However, due to the `cli.php` file running early in WordPress, certain functions cannot be used there. As a workaround, the necessary code has been added to the `includes/CLI/Plugin_Check_Command.php` file.

**Steps to Reproduce:**
1. Install any plugin that you want to check.
2. Deactivate the "Plugin Checker" from `Admin Panel > Plugins`.
3. Execute the following CLI command:
   ```
   npm run wp-env run cli -- wp plugin check {plugin} --require=./wp-content/plugins/plugin-check/cli.php
   ```
   Replace `{plugin}` with the slug of the plugin you want to check.
4. Observe the error that occurs.

**Expected behaviour:**
After applying the changes from the pull request, the CLI command should display the error message: `Error: plugin-check/plugin-check.php is not active.` when attempting to use it while the Plugin Checker is not activated.

### Checklist:

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -114,11 +114,7 @@ final class Plugin_Check_Command {
 		 */
 		if ( is_plugin_inactive( $this->plugin_context->basename() ) ) {
 			WP_CLI::error(
-				sprintf(
-					// translators: Plugin Checker slug.
-					__( '%s is not active.', 'plugin-check' ),
-					$this->plugin_context->basename()
-				)
+				__( 'Plugin Checker is not active.', 'plugin-check' )
 			);
 		}
 
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -105,6 +105,22 @@ final class Plugin_Check_Command {
 	 * @SuppressWarnings(PHPMD.NPathComplexity)
 	 */
 	public function check( $args, $assoc_args ) {
+		/*
+		 * Bail early if the Plugin Checker is not activated.
+		 *
+		 * If the Plugin Checker is not activated, it will throw an error and
+		 * CLI commands won't be usable.
+		 */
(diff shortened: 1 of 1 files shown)


includes/CLI/Plugin_Check_Command.php
```diff
@@ -103,8 +103,25 @@ public function __construct( Plugin_Context $plugin_context ) {
...
+		 * If the Plugin Checker is not activated, it will throw an error and
+		 * CLI commands won't be usable.
+		 */
+		if ( is_plugin_inactive( $this->plugin_context->basename() ) ) {
+			WP_CLI::error(
+				sprintf(
+					// translators: Plugin Checker slug.
+					__( '%s is not active.', 'plugin-check' ),
+					$this->plugin_context->basename()
+				)
```
swissspidy: The translator comment must actually reference the placeholder, i.e. `// translators: %s: Plugin Checker slug.`

However, why not just simplify the error message?

```suggestion
				__( 'Plugin Checker is not active.', 'plugin-check' )
```

swissspidy (approved):
Overall looks good.

Reminds me to look into #153 soon

felixarntz (approved):
Thanks @mukeshpanchal27, LGTM, would be great if you could incorporate @swissspidy's suggestion.

mukeshpanchal27:
Thank you, @swissspidy, for the suggestion. We will definitely take a look at item 153 as soon as possible. Your input is appreciated!

PR #237, opened 2023-07-31
Ignore `.git`, `node_modules`, and `vendor` directories from the checks

### LINKED ISSUE #224: Investigate why the memory limit issue is occurring when testing the plugin checker plugin.
### Describe the bug

Check discussion here https://github.com/10up/plugin-check/issues/213#issuecomment-1609786902

### Steps to Reproduce

Follow steps describe here https://github.com/10up/plugin-check/issues/213#issuecomment-1610717623

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #224

### Checklist:

### DIFF
--- a/includes/Utilities/Plugin_Request_Utility.php
+++ b/includes/Utilities/Plugin_Request_Utility.php
@@ -134,4 +134,29 @@ class Plugin_Request_Utility {
 
 		static::$runner = null;
 	}
+
+	/**
+	 * Gets the directories to ignore using the filter.
+	 *
+	 * @since n.e.x.t
+	 */
+	public static function get_directories_to_ignore() {
+		// By default, ignore the '.git', 'vendor' and 'node_modules' directories.
+		$default_ignore_directories = array(
+			'.git',
+			'vendor',
+			'node_modules',
+		);
+
+		/**
+		 * Filters the directories to ignore.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param array $default_ignore_directories An array of directories to ignore.
+		 */
+		$directories_to_ignore = (array) apply_filters( 'wp_plugin_check_ignore_directories', $default_ignore_directories );
+
+		return $directories_to_ignore;
+	}
 }
--- a/tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
+++ b/tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
@@ -205,6 +205,11 @@ class Plugin_Request_Utility_Tests extends WP_UnitTestCase {
 		$this->assertEquals( $custom_ignore_directories, $result );
 
 		// Remove the filter to avoid interfering with other tests.
-		remove_filter( $filter_name );
+        remove_filter(
+			$filter_name,
+			static function () use ( $custom_ignore_directories ) {
+            	return $custom_ignore_directories;
+        	}
+		);
 	}
 }
--- a/tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
+++ b/tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
@@ -6,6 +6,10 @@
  */
 
 use WordPress\Plugin_Check\Checker\AJAX_Runner;
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks;
+use WordPress\Plugin_Check\Checker\Checks\I18n_Usage_Check;
 use WordPress\Plugin_Check\Checker\CLI_Runner;
 use WordPress\Plugin_Check\Test_Data\Runtime_Check;
 use WordPress\Plugin_Check\Test_Utils\Traits\With_Mock_Filesystem;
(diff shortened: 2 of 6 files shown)


felixarntz (commented):
@mukeshpanchal27 Thanks for implementing the PR! I think this code works fine for now, though there are also cases where plugins actually use `vendor` or `node_modules` directories even in the distributed plugin bundle.

I think it would be good to make this behavior conditional. For example, we could apply a filter on the array of directory names to ignore, e.g. `wp_plugin_check_ignore_directories`, implemented in a single method to get those directories. Then this method could be used in the two places here.

The default behavior here makes sense, but with such a filter it could then be adjusted as needed.

mukeshpanchal27:
@felixarntz, I have implemented the filter as per your feedback. Could you please take a look? Thanks!

felixarntz (approved):
Thanks @mukeshpanchal27, that looks great! +1 for the unit tests you added 🎉

mukeshpanchal27:
Thanks @felixarntz.

@joemcgill Could you please take a look.

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
```diff
@@ -171,4 +171,45 @@ public function test_destroy_runner_with_no_runner() {
 
 		$this->assertNull( $runner );
 	}
+
+	public function test_default_ignore_directories() {
```
joemcgill, line 179: These tests are both nice additions. It would be great if we could add an additional integration test that can run against an example plugin in the [testdata](https://github.com/10up/plugin-check/tree/trunk/tests/phpunit/testdata/plugins) directory to ensure that ignored files are actually being ignored.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -134,4 +134,29 @@ public static function destroy_runner() {
...
+	 * Gets the directories to ignore using the filter.
+	 *
+	 * @since n.e.x.t
+	 */
+	public static function get_directories_to_ignore() {
+		// By default, ignore the '.git', 'vendor' and 'node_modules' directories.
+		$default_ignore_directories = array(
+			'.git',
+			'vendor',
+			'node_modules',
```
joemcgill, line 148: This is a good initial list, but after seeing [your comment on the original issue](https://github.com/10up/plugin-check/issues/224#issuecomment-1658130784) after running this, I wonder if we should expand this to include other hidden files, or update the checks to ensure they're only running against the types of files we want to check, e.g. PHP?

joemcgill (commented):
Nice to see this is having the desired effect on the memory issues. Left a few comments for your consideration. Thanks, @mukeshpanchal27!

tests/phpunit/Utilities/Plugin_Request_Utility_Tests.php
mukeshpanchal27: Thanks @joemcgill I have added unit test as requested.

includes/Utilities/Plugin_Request_Utility.php
mukeshpanchal27: I personally don't think we have to add that because it's the result of the `File_Type_Check` check, and I ran checks in the development version of the plugin, and it makes sense to show those errors.

In the future, we can add possible values if we receive support tickets for that, or users can ignore them using the provided filter.

I would like to get @felixarntz feedback here.

felixarntz: The primary reason to exclude these by default is that they are _directories_ that are usually not used for actual plugin code, plus they're potentially huge and thus costly to analyze. Hidden files are different because there's no real cost to covering them, plus ignoring them may actually be problematic since they _should_ explicitly be discoverable for the `File_Type_Check` to work correctly and flag them.

So I think leaving it to these few, commonly large directories for now is good.

Regarding your point on which type of files to check, that is a good point, however that is already implemented at a higher level, the `Abstract_File_Check` class already provides methods (that internally use this code here) which allow you to limit to certain file types.

joemcgill: Sounds good to me. Thanks all.

tests/phpunit/testdata/plugins/test-plugin-ignore-directories/custom_directory/error.php
```diff
@@ -0,0 +1,18 @@
...
+ // This will cause a WordPress.WP.I18n.MissingTranslatorsComment error as it has no translators comment.
+ sprintf(
+     __( 'Your city is %s.', 'test-plugin-check-errors' ),
+     $city
+ );
+ 
+ $text_domain = 'test-plugin-check-errors';
+ 
+ // This will cause a WordPress.WP.I18n.NonSingularStringLiteralDomain error as a variable is used for the text-domain.
+ esc_html__( 'Hello World!', $text_domain );
```
mukeshpanchal27:
```suggestion
 esc_html__( 'Hello World!', $text_domain );
 
```

mukeshpanchal27:
@joemcgill or @swissspidy, could one of you please review this PR when you have a moment? Thanks.

swissspidy (approved):

PR #238, opened 2023-08-01
Fix errors detected by Plugin Checker

mukeshpanchal27:
This pull request addresses several issues detected by the Plugin Checker when validating the plugin itself. The changes in this PR aim to resolve potential problems and improve overall code quality.

Please refer to https://github.com/10up/plugin-check/issues/224#issuecomment-1658130784 for a detailed list of errors identified by the Plugin Checker during the validation process.

### Checklist:

### DIFF
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -36,7 +36,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 			return false;
 		}
 
-		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
+		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( ! isset( $_REQUEST['action'] ) || 'plugin_check_run_checks' !== $_REQUEST['action'] ) {
 			return false;
@@ -55,7 +55,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 	 * @throws Exception Thrown if the plugin parameter is empty.
 	 */
 	protected function get_plugin_param() {
-		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
+		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( ! isset( $_REQUEST['plugin'] ) ) {
 			throw new Exception(
@@ -63,7 +63,7 @@ class AJAX_Runner extends Abstract_Check_Runner {
 			);
 		}
 
-		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
+		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		return sanitize_text_field( $_REQUEST['plugin'] );
 	}
--- a/includes/Checker/AJAX_Runner.php
+++ b/includes/Checker/AJAX_Runner.php
@@ -36,6 +36,8 @@ class AJAX_Runner extends Abstract_Check_Runner {
 			return false;
 		}
 
+		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( ! isset( $_REQUEST['action'] ) || 'plugin_check_run_checks' !== $_REQUEST['action'] ) {
 			return false;
 		}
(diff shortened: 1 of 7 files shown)


mukeshpanchal27:
@felixarntz @joemcgill In https://github.com/10up/plugin-check/pull/238/commits/d7012f46537db22bb5d5fe32d9f75eba11ce9e89 i added ignore for warning `file_get_contents() is discouraged. Use wp_remote_get() for remote URLs instead.` Do we need to update function?

templates/admin-page.php
```diff
@@ -43,10 +43,10 @@
 					<tr>
 						<td>
 							<fieldset>
-								<legend class="screen-reader-text"><?php esc_html_e( $category ); ?></legend>
+								<legend class="screen-reader-text"><?php echo esc_attr( $category ); ?></legend>
```
swissspidy: This is not inside an attribute, so `echo esc_html` is more applicable

templates/admin-page.php
```diff
@@ -43,10 +43,10 @@
...
 								<label for="<?php echo esc_attr( $category ); ?>">
 									<input type="checkbox" id="<?php echo esc_attr( $category ); ?>" name="categories" value="<?php echo esc_attr( $category ); ?>" checked="checked" />
-									<?php esc_html_e( ucfirst( str_replace( '_', ' ', $category ) ) ); ?>
+									<?php echo esc_attr( ucfirst( str_replace( '_', ' ', $category ) ) ); ?>
```
swissspidy: This is not inside an attribute, so `echo esc_html` is more applicable

templates/admin-page.php
```diff
@@ -43,10 +43,10 @@
 					<tr>
 						<td>
 							<fieldset>
-								<legend class="screen-reader-text"><?php esc_html_e( $category ); ?></legend>
+								<legend class="screen-reader-text"><?php echo esc_html( $category ); ?></legend>
```
joemcgill, line 46: `esc_html_e` seems appropriate in both cases here. Is there a reason why these are being changed?

includes/Checker/AJAX_Runner.php
```diff
@@ -36,6 +36,8 @@ public static function is_plugin_check() {
 			return false;
 		}
 
+		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
```
felixarntz:
```suggestion
		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
```

includes/Checker/AJAX_Runner.php
```diff
@@ -53,12 +55,16 @@ public static function is_plugin_check() {
 	 * @throws Exception Thrown if the plugin parameter is empty.
 	 */
 	protected function get_plugin_param() {
+		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
```
felixarntz:
```suggestion
		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
```

includes/Checker/AJAX_Runner.php
```diff
@@ -53,12 +55,16 @@ public static function is_plugin_check() {
...
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( ! isset( $_REQUEST['plugin'] ) ) {
 			throw new Exception(
 				__( 'Invalid plugin: Plugin parameter must not be empty.', 'plugin-check' )
 			);
 		}
 
+		// PHPCS ignore reason: Nonce check is not required in this case since the logic is directly tied to an Ajax call.
```
felixarntz:
```suggestion
		// PHPCS ignore reason: Nonce check is already happening before this logic in `Admin_AJAX` class.
```

templates/admin-page.php
felixarntz: Agreed, this doesn't look wrong to me.

felixarntz (commented):
@mukeshpanchal27 Just a minor suggestion to clarify the comment about nonce verification, since it _should_ happen, but it does already happen elsewhere (in `Admin_AJAX` class).

Other than that, looks good except for @joemcgill's existing point.

templates/admin-page.php
mukeshpanchal27: When using `esc_html_e`, the error `WordPress.WP.I18n.NonSingularStringLiteralText` is shown by PHPCS, which is related to how WordPress handles translation functions for plural and singular text strings. The problem arises from using the `esc_html_e()` function with a variable as the text parameter.

joemcgill: Thanks for the explanation, @mukeshpanchal27. Of course you can't translate a variable. This change looks correct 👍🏻

joemcgill (approved):

templates/admin-page.php
felixarntz: Oh 🤦 You're right, thanks for clarifying!

felixarntz (approved):
Thank you for the update @mukeshpanchal27, LGTM!

Issue #195, opened 2023-06-20
Add `readme.md` check in `Plugin_Readme_Check`


mukeshpanchal27:
### Is your enhancement related to a problem? Please describe.

Follow up: https://github.com/10up/plugin-check/pull/191#discussion_r1235104550

We have to add `readme.md` check in `Plugin_Readme_Check`.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #239, opened 2023-08-01
Incorrect display of `readme.txt` instead of `readme.md` in Plugin Readme check

mukeshpanchal27:
In issue #195, it was allowed to use a `readme.md` file for plugin. However, when someone uses the `readme.md` file and runs the plugin checks, the plugin incorrectly displays `readme.txt` as the file name instead of the actual `readme.md` file.

### Checklist:

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -12,7 +12,7 @@ use WordPress\Plugin_Check\Checker\Check_Result;
 use WordPress\Plugin_Check\Traits\Stable_Check;
 
 /**
- * Check the plugins readme.txt file and contents.
+ * Check the plugins readme file and contents.
  *
  * @since n.e.x.t
  */
@@ -34,7 +34,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	}
 
 	/**
-	 * Check the readme.txt file.
+	 * Check the readme file.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -45,7 +45,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		// Find the readme file.
 		$readme = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
 
-		// If the readme.txt does not exist, add a warning and skip other tests.
+		// If the readme file does not exist, add a warning and skip other tests.
 		if ( empty( $readme ) ) {
 			$result->add_message(
 				false,
@@ -59,18 +59,18 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 			return;
 		}
 
-		// Check the readme.txt for default text.
+		// Check the readme file for default text.
 		$this->check_default_text( $result, $readme );
 
-		// Check the readme.txt for a valid license.
+		// Check the readme file for a valid license.
 		$this->check_license( $result, $readme );
 
-		// Check the readme.txt for a valid version.
+		// Check the readme file for a valid version.
 		$this->check_stable_tag( $result, $readme );
 	}
 
 	/**
-	 * Checks the readme.txt for default text.
+	 * Checks the readme file for default text.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -78,24 +78,38 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	 * @param array        $files  Array of plugin files.
 	 */
 	private function check_default_text( Check_Result $result, array $files ) {
-		if (
-			self::file_str_contains( $files, 'Here is a short description of the plugin.' ) ||
-			self::file_str_contains( $files, 'Tags: tag1' ) ||
-			self::file_str_contains( $files, 'Donate link: http://example.com/' )
-		) {
+		$default_text_patterns = array(
+			'Here is a short description of the plugin.',
+			'Tags: tag1',
+			'Donate link: http://example.com/',
+		);
+
+		$file = '';
+		foreach ( $default_text_patterns as $pattern ) {
+			$file = self::file_str_contains( $files, $pattern );
+			if ( $file ) {
+				break;
+			}
+		}
+
+		if ( $file ) {
 			$result->add_message(
 				false,
-				__( 'The readme.txt appears to contain default text.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'The %s appears to contain default text.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
 				array(
 					'code' => 'default_readme_text',
-					'file' => $result->plugin()->path( '/readme.txt' ),
+					'file' => $file,
 				)
 			);
 		}
 	}
 
 	/**
-	 * Checks the readme.txt for a valid license.
+	 * Checks the readme file for a valid license.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -105,7 +119,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	private function check_license( Check_Result $result, array $files ) {
 		$matches = array();
 		// Get the license from the readme file.
-		self::file_preg_match( '/(License:|License URI:)\s*(.+)*/i', $files, $matches );
+		$file = self::file_preg_match( '/(License:|License URI:)\s*(.+)*/i', $files, $matches );
 
 		if ( empty( $matches ) ) {
 			return;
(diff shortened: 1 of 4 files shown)


joemcgill:
I'm curious about the initial intent for this check. I think it's still a requirement for a plugin to have a `readme.txt` file, even if it also includes a `readme.md`, since the header in the txt version is the one used for parsing out plugin info like version, tested up to, etc. 

See: https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information

felixarntz:
@joemcgill I thought so too, but someone else (I don't recall, maybe it was @swissspidy or @westonruter?) pointed out that wordpress.org, despite not well documenting it, actually supports using `readme.md` as well as the _actual_ .org readme file. So if you don't include `readme.txt` but include `readme.md`, wordpress.org will try to parse its contents apparently. Of course that only works if you format `readme.md` correctly for that purpose, not if it's a general readme e.g. for just GitHub.

westonruter:
> I thought so too, but someone else (I don't recall, maybe it was @swissspidy or @westonruter?) pointed out that wordpress.org

I think it was actually @johnbillion: https://johnblackbourn.com/readme-md-plugin-on-wordpress-org/

For the AMP plugin we initially tried just going with `README.md` alone in https://github.com/ampproject/amp-wp/pull/5807, but in https://github.com/ampproject/amp-wp/pull/5815 we re-introduced the `readme.txt` but [created it](https://github.com/ampproject/amp-wp/blob/cebab46e081f55815f4be83529b89feee34f3020/bin/transform-readme.php) at build time out of the `README.md` due to various WordPress-specific issues with the format.

joemcgill:
That's really interesting, I didn't realize this was supported. Still, I'm curious if the lack of a README.txt file would get flagged in the plugin review process, and if so, is something we should still include as a check in our plugin?

felixarntz:
I don't think we need to specifically check for `readme.txt`, as either works. This is also confirmed by the other plugin checker code, which supports either file extension in the same way: https://github.com/WordPress/plugin-check/blob/trunk/plugin.php#L76

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -59,43 +59,57 @@ protected function check_files( Check_Result $result, array $files ) {
...
-				__( 'The readme.txt appears to contain default text.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'The %s appears to contain default text.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
 				array(
 					'code' => 'default_readme_text',
-					'file' => $result->plugin()->path( '/readme.txt' ),
+					'file' => $file,
```
felixarntz: This needs to also be the path relative to the plugin's directory, similar to other checks.
```suggestion
					'file' => str_replace( $result->plugin()->path(), '', $file ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -59,43 +59,57 @@ protected function check_files( Check_Result $result, array $files ) {
...
+
+		if ( $file ) {
 			$result->add_message(
 				false,
-				__( 'The readme.txt appears to contain default text.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'The %s appears to contain default text.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
```
felixarntz: I think we can simplify this since the concrete file path is part of the data below anyway.
```suggestion
				__( 'The readme appears to contain default text.', 'plugin-check' ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -115,17 +129,21 @@ private function check_license( Check_Result $result, array $files ) {
...
-				__( 'Your plugin has an invalid license declared. Please update your readme.txt with a valid SPDX license identifier.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'Your plugin has an invalid license declared. Please update your %s with a valid SPDX license identifier.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
 				array(
 					'code' => 'invalid_license',
-					'file' => $result->plugin()->path( '/readme.txt' ),
+					'file' => $file,
```
felixarntz: See above:
```suggestion
					'file' => str_replace( $result->plugin()->path(), '', $file ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -115,17 +129,21 @@ private function check_license( Check_Result $result, array $files ) {
 		if ( ! preg_match( '/^([a-z0-9\-\+\.]+)(\sor\s([a-z0-9\-\+\.]+))*$/i', $matches[2] ) ) {
 			$result->add_message(
 				false,
-				__( 'Your plugin has an invalid license declared. Please update your readme.txt with a valid SPDX license identifier.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'Your plugin has an invalid license declared. Please update your %s with a valid SPDX license identifier.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
```
felixarntz: See above:
```suggestion
				__( 'Your plugin has an invalid license declared. Please update your readme with a valid SPDX license identifier.', 'plugin-check' ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -147,12 +166,12 @@ private function check_stable_tag( Check_Result $result, array $files ) {
 				__( "It's recommended not to use 'Stable Tag: trunk'.", 'plugin-check' ),
 				array(
 					'code' => 'trunk_stable_tag',
-					'file' => $result->plugin()->path( '/readme.txt' ),
+					'file' => $file,
```
felixarntz: See above:
```suggestion
					'file' => str_replace( $result->plugin()->path(), '', $file ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -161,10 +180,14 @@ private function check_stable_tag( Check_Result $result, array $files ) {
...
-				__( 'The Stable Tag in your readme.txt file does not match the version in your main plugin file.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'The Stable Tag in your %s file does not match the version in your main plugin file.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
 				array(
 					'code' => 'stable_tag_mismatch',
-					'file' => $result->plugin()->path( '/readme.txt' ),
+					'file' => $file,
```
felixarntz: See above:
```suggestion
					'file' => str_replace( $result->plugin()->path(), '', $file ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -161,10 +180,14 @@ private function check_stable_tag( Check_Result $result, array $files ) {
 		) {
 			$result->add_message(
 				false,
-				__( 'The Stable Tag in your readme.txt file does not match the version in your main plugin file.', 'plugin-check' ),
+				sprintf(
+					/* translators: %s: readme file */
+					__( 'The Stable Tag in your %s file does not match the version in your main plugin file.', 'plugin-check' ),
+					str_replace( $result->plugin()->path( '/' ), '', $file )
+				),
```
felixarntz: See above:
```suggestion
				__( 'The Stable Tag in your readme file does not match the version in your main plugin file.', 'plugin-check' ),
```

felixarntz (changes requested):
Thanks @mukeshpanchal27, a few minor bits of feedback below to make the file paths consistent with other checks (relative to the plugin's directory).

mukeshpanchal27:
Thanks all the the review and details. The PR is ready for review.

mukeshpanchal27:
The Plugin Checker currently checks all files, including in subfolders. However, we should modify the check to only consider files in the root directory for `readme.txt` and `readme.md` checks. 

WordPress allows and parse `readme.txt` and `readme.md` files in root directory of the plugin. Therefore, we will adjust the Plugin Checker accordingly to focus only on root directory files.

For instance, when we checked the WooCommerce plugin, the Plugin Checker returned an array of files as follows:

```
/var/www/html/wp-content/plugins/woocommerce/Array
(
    [1511] => /var/www/html/wp-content/plugins/woocommerce/packages/action-scheduler/readme.txt
    [1668] => /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/assets/js/base/README.MD
    [3531] => /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/readme.txt
    [3962] => /var/www/html/wp-content/plugins/woocommerce/readme.txt
)
```

The specific error highlighted was:

<h4>FILE: packages/action-scheduler/readme.txt</h4>

Line | Column | Type | Code | Message
-- | -- | -- | -- | --
0 | 0 | WARNING | stable_tag_mismatch | The Stable Tag in your readme file does not match the version in your main plugin file.

We will update the Plugin Checker to only report errors for `readme.txt` and `readme.md` files that are directly in the root directory of the plugin and not in subfolders.

@felixarntz @joemcgill If the above makes sense, then I will open the follow-up issue and start working on it. Thanks.

felixarntz:
Thanks @mukeshpanchal27, that's a great catch! +1 to opening a follow-up issue and PR to fix it so that the plugin checker only considers readme files in the root directory.

felixarntz (approved):
LGTM, let's open a follow-up PR for the other problem you discovered above. Thank you @mukeshpanchal27!

joemcgill (approved):

PR #241, opened 2023-08-03
Modify Plugin Checker to only check root directory readme files

### LINKED ISSUE #240: Modify Plugin Checker to only check root directory readme files
### Describe the bug

Followup issue https://github.com/10up/plugin-check/pull/239#issuecomment-1661953900

The Plugin Checker currently examines all files, including those in subfolders, to validate readme.txt and readme.md files. However, this behaviour should be adjusted to focus solely on files in the root directory of the plugin for readme checks.

WordPress allows and parses readme.txt and readme.md files in the root directory of the plugin. Therefore, the Plugin Checker should be updated to consider only these root directory files.

For example, when testing the WooCommerce plugin, the Plugin Checker returned an array of files as follows:

```
/var/www/html/wp-content/plugins/woocommerce/Array
(
    [1511] => /var/www/html/wp-content/plugins/woocommerce/packages/action-scheduler/readme.txt
    [1668] => /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/assets/js/base/README.MD
    [3531] => /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/readme.txt
    [3962] => /var/www/html/wp-content/plugins/woocommerce/readme.txt
)
```

The specific error highlighted was:

#### FILE: packages/action-scheduler/readme.txt

Line | Column | Type | Code | Message
-- | -- | -- | -- | --
0 | 0 | WARNING | stable_tag_mismatch | The Stable Tag in your readme file does not match the version in your main plugin file.

To address this issue, the Plugin Checker should be updated to only report errors for readme.txt and readme.md files that exist directly in the root directory of the plugin and not in any subfolders. This adjustment will ensure accurate validation of these critical files.

### Steps to Reproduce

1. Install and activate WooCommerce Plugin
2. Run checks using Plugin Checker
3. It show `stable_tag_mismatch` error from sub directory file `packages/action-scheduler/readme.txt`

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #240

### Checklist:


mukeshpanchal27:
One thing I encountered while working on this PR is whether any plugins have both `readme.txt` and `readme.md` files. If they do, should we display an error for both files?

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -76,32 +85,29 @@ protected function check_files( Check_Result $result, array $files ) {
 	 *
 	 * @param Check_Result $result The Check Result to amend.
 	 * @param array        $files  Array of plugin files.
+	 * @param string       $path   Plugin relative path.
 	 */
-	private function check_default_text( Check_Result $result, array $files ) {
+	private function check_default_text( Check_Result $result, array $files, $path ) {
```
felixarntz: I don't think we need the `$path` parameter? IMO it's cleaner to keep using `$result->plugin()->path()` below. There is no meaningful cost to calling that function, and the fewer parameters necessary, the better.

felixarntz (commented):
Thanks @mukeshpanchal27, LGTM. Just one recommendation.

felixarntz:
@mukeshpanchal27 
> One thing I encountered while working on this PR is whether any plugins have both `readme.txt` and `readme.md` files. If they do, should we display an error for both files?

I don't think we need to flag this. A plugin may contain a `readme.md` file for GitHub, which is not a problem. However, what I _would_ suggest is that, if there is a `readme.txt`, we should ignore the `readme.md` - since the `readme.txt` is the typical place and only has that purpose, while `readme.md` could be either for wordpress.org readme, or GitHub.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -42,8 +42,17 @@ public function get_categories() {
...
+		$root_readme          = array(
+			$plugin_relative_path . 'readme.txt',
+			$plugin_relative_path . 'readme.md',
+		);
+
 		// Find the readme file.
-		$readme = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+		$readme_list = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+
+		$readme = array_intersect( $root_readme, $readme_list );
```
felixarntz: Actually, there is a problem here: We have to consider the file names regardless of case, for example, it could also be `README.md`. It would be great to find an alternative solution where we don't have to hardcode all the file names and use an exact string match.

Potentially you could use `array_filter()` to ensure that only readme files are considered which are directly in the plugin root?

Additionally, to your question, we need to add logic here to ignore any `.md` versions _if_ there are `.txt` versions.

felixarntz (changes requested):
Actually, there is one issue that I found here with the implementation.

includes/Checker/Checks/Plugin_Readme_Check.php
mukeshpanchal27: Thanks for feedback. Address in https://github.com/10up/plugin-check/pull/241/commits/cae8aace9aee62ae530b2f92f1fe8455f532a894

includes/Checker/Checks/Plugin_Readme_Check.php
mukeshpanchal27: Thanks @felixarntz for the feedback. In https://github.com/10up/plugin-check/pull/241/commits/3c15ebeff992ca809841dc35e179f29735399d71 I address this. Ready for review.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -42,8 +42,41 @@ public function get_categories() {
...
 	 */
 	protected function check_files( Check_Result $result, array $files ) {
+
+		$plugin_relative_path = $result->plugin()->path();
+
 		// Find the readme file.
-		$readme = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+		$readme_list = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+
+		// Filter the readme files that located at rootx.
```
swissspidy:
```suggestion
		// Filter the readme files located at root.
```

swissspidy (approved):

felixarntz (approved):

PR #255, opened 2023-09-21
Include assets for WP.org

bordoni:
### Description of the Change

Uploaded to the Repo already.

image

I took some of the existing [Dashicons](https://icon-sets.iconify.design/dashicons/) modified a bit and made the assets,
I've used the WordPress blue and Grey for the header and just the blue for the Icon.

### Checklist:


felixarntz (commented):
Thanks @bordoni!

Quick question, why are those added to `.assets`? Based on https://github.com/10up/action-wordpress-plugin-deploy#optional-environment-variables, shouldn't they be in a `.wordpress-org` directory?

Separate question: I assume we should also add those assets to the `trunk` branch?

bordoni:
Oh yeah I will do the move! Thanks for the review

bordoni:
Done.

felixarntz (approved):
LGTM!

One follow up question: Are the `icon-128x128.png` and `icon-256x256.png` files intentionally missing? Not sure if the `icon.svg` works everywhere, so we may want to add those two versions as well. Not a blocker though.

bordoni:
I didn't know about the `.svg` not working in some places.
I will add the others in a separate PR.

Issue #253, opened 2023-09-11
Achieve check feature parity with plugin review team's current tool


felixarntz:
### Is your enhancement related to a problem? Please describe.

Reviewing https://github.com/WordPress/plugin-check/commits/trunk, several check related changes have landed in the past 2 weeks that should be incorporated into our plugin checker for parity.

Each of the below top-level bullet points should preferably be addressed in its own PR:
* [x] Update severity of certain checks in `Plugin_Readme_Check` to "error" (see https://github.com/WordPress/plugin-check/commit/c5d539b6d248dba81fcbb2218207c7302aaf91c4).
* [x] Update Plugin Check Review PHPCS XML file related to `Plugin_Review_PHPCS_Check` with latest changes from https://github.com/WordPress/plugin-check/commit/593983160d936e35369930a3a1b07d3dbb538e6e, https://github.com/WordPress/plugin-check/commit/ee56bc209ffca0f552fde9e8da18de9356816d65, and https://github.com/WordPress/plugin-check/commit/542ea3c704f1886237db86a832bb64fac0cea6e2.
* [x] Add application files support to `File_Type_Check` (see https://github.com/WordPress/plugin-check/commit/b081cf7aa0ccc32e158d803b8905a66914c3f086).
* [x] Implement new check for plugin trademark violations (see https://github.com/WordPress/plugin-check/commit/f15d41c26ece3cc760c90e87aaf4a4966b873784).

cc @eclarke1 @joemcgill @bordoni

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bordoni:
Thank you for logging this here.
I will start porting over the code into the branch later today, so we can reference the old code from this repo.

bordoni:
All of the code from our legacy plugin repository is now living here on the [`legacy-plugin`](https://github.com/10up/plugin-check/tree/legacy-plugin) branch.

_Sorry for the random mentions that were made once the commits were ported, since the hashtag mentions were to the old issues and PRs._

I will manually release version `0.2.0` this week to the WP.org, once we the Plugin Review team is ok with a launch date.

bordoni:
We are live with version `0.2.0` on WordPress.org.

https://wordpress.org/plugins/plugin-check/

mukeshpanchal27:
@felixarntz, I am closing this issue as we have successfully merged all the top-level bullet point changes into the trunk.

PR #257, opened 2023-09-21
Update readme to match svn and improve it minorly

aaronjorbin:
### Description of the Change
The Readme doesn't match the one in SVN.  The SVN one also says `This is very handy for theme developers, or anybody looking to make sure that their plugin supports`, I changed that to say it's very handy for **plugin** developers. 

### How to test the Change
Proofread and make sure I copy and pasted correctly. 

### Credits
Just me. 

### Checklist:

-  [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).


bordoni:
@aaronjorbin thank you so much for the contribution, the not matching is because we are running a separate branch for the pre-version 1.0.0 which will be based on the work the performance team.

Current branch with the code: `legacy-plugin`

A little bit more context on #253.

Would you mind if I pointed your change to that branch? So we can get that readme fix?

aaronjorbin:
@bordoni Thanks, that totally makes sense. I'm going to close this and open a new one since as you see above, everything goes a little haywire since this branch was based off main.

PR #276, opened 2023-09-23
Create a way to remove certain readme warnings, contributors_ignored.

bordoni:
### Description of the Change
A Couple of users mentioned that they are seeing problems on their reports related to a warning on their `readme.txt`.

After doing some digging on the readme parser I realized that this is a warning that should only be active on WordPress.org, so I add a filter to allow us to enable it on that environment but by default it's now ignored.

### DIFF
--- a/checks/readme.php
+++ b/checks/readme.php
@@ -1,6 +1,7 @@
 <?php
 namespace WordPressdotorg\Plugin_Check\Checks;
 use WordPressdotorg\Plugin_Check\{Error, Guideline_Violation, Message, Notice, Warning};
+use WordPressdotorg\Plugin_Directory\Readme\Parser;
 
 class Readme extends Check_Base {
 	/**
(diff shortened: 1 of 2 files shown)


checks/readme.php
```diff
@@ -76,13 +77,31 @@ function check_for_default_text() {
...
+
+		/**
+		 * Filter the list of ignored readme parser warnings.
+		 *
+		 * @since 0.2.2
+		 *
+		 * @param array  $ignored_warnings Array of ignored warning keys.
+		 * @param Parser $readme The readme object.
+		 */
+		$ignored_warnings = apply_filters( 'plugin_check_readme_warnings_ignored', $ignored_warnings, $this->readme );
```
EvanHerman: Just for good measure, we should probably type cast this as an array, incase users return a string.
```suggestion
		$ignored_warnings = (array) apply_filters( 'plugin_check_readme_warnings_ignored', $ignored_warnings, $this->readme );
```

PR #278, opened 2023-09-24
Do not flag `localhost` usage in comments

### LINKED ISSUE #267: Do not raise error "Do not use Localhost in your code" when "localhost" in comment

I'm given the following error:

```
Do not use Localhost in your code. Detected: http://localhost
```

However, my code doesn't use `localhost` at all. The only occurrence is inside a comment (from a 3rd-party package):

```php
/**
 * Returns the path being requested relative to the executed script.
 *
 * The path info always starts with a /.
 *
 * Suppose this request is instantiated from /mysite on localhost:
 *
 *  * http://localhost/mysite              returns an empty string
 *  * http://localhost/mysite/about        returns '/about'
 *  * http://localhost/mysite/enco%20ded   returns '/enco%20ded'
 *  * http://localhost/mysite/about?var=1  returns '/about'
 *
 * @return string The raw path (i.e. not urldecoded)
 */
public function getPathInfo(): string
{
    return $this->pathInfo ??= $this->preparePathInfo();
}
```

I guess the error should not be raised when the match happens inside a comment.

### Designs

Screenshot 2023-09-22 at 12 01 33 PM

_No response_

### Code of Conduct

nielslange:
### Description of the Change

In #267, @leoloso reported that the error "Do not use Localhost in your code" also gets triggered when the string "localhost" is used within a comment. This PR updates the corresponding regex to ignore the string "localhost" within comments.

Closes #267

### How to test the Change

1. Download and activate this [helper-plugin.zip](https://github.com/10up/plugin-check/files/12708275/helper-plugin.zip)
2. Go to `/wp-admin/tools.php?page=plugin-check` and check the helper plugin before applying this fix.
3. Verify that the error "Do not use Localhost/127.0.0.1 in your code." appears.
4. Apply this PR and check the helper plugin again.
5. Verify that the error "Do not use Localhost/127.0.0.1 in your code." no longer appears.

### Changelog Entry

> Fixed - “localhost” in comments no longer raises error.

### Credits

Props @nielslange

### Checklist:

### DIFF
--- a/includes/Checker/Checks/Localhost_Check.php
+++ b/includes/Checker/Checks/Localhost_Check.php
@@ -43,7 +43,8 @@ class Localhost_Check extends Abstract_File_Check {
 	 */
 	protected function check_files( Check_Result $result, array $files ) {
 		$php_files = self::filter_files_by_extension( $files, 'php' );
-		$file      = self::file_preg_match( '#https?://(localhost|127.0.0.1)#', $php_files );
+		$file      = self::file_preg_match( '#(?<!://)(?<!\* )https?://(localhost|127.0.0.1)#', $php_files );
+
 		if ( $file ) {
 			$result->add_message(
 				true,
--- a/includes/Checker/Checks/Localhost_Check.php
+++ b/includes/Checker/Checks/Localhost_Check.php
@@ -43,17 +43,31 @@ class Localhost_Check extends Abstract_File_Check {
 	 */
 	protected function check_files( Check_Result $result, array $files ) {
 		$php_files = self::filter_files_by_extension( $files, 'php' );
-		$file      = self::file_preg_match( '#(?<!://)(?<!\* )https?://(localhost|127.0.0.1)#', $php_files );
 
-		if ( $file ) {
-			$result->add_message(
-				true,
-				__( 'Do not use Localhost/127.0.0.1 in your code.', 'plugin-check' ),
-				array(
-					'code' => 'localhost_code_detected',
-					'file' => $file,
-				)
-			);
+		foreach ( $php_files as $file ) {
+			$code   = file_get_contents( $file );
+			$tokens = token_get_all( $code );
+
+			foreach ( $tokens as $token ) {
+				if ( is_array( $token ) ) {
+					if ( in_array( $token[0], array( T_COMMENT, T_DOC_COMMENT ), true ) ) {
+						continue;
+					}
+
+					if ( preg_match( '#https?://(localhost|127.0.0.1)#', $token[1] ) ) {
+						$result->add_message(
+							true,
+							__( 'Do not use Localhost/127.0.0.1 in your code.', 'plugin-check' ),
+							array(
+								'code' => 'localhost_code_detected',
+								'file' => $file,
+							)
+						);
+						break;
+					}
+				}
+			}
 		}
 	}
+
 }
--- a/tests/phpunit/Checker/Checks/Localhost_Check_Tests.php
+++ b/tests/phpunit/Checker/Checks/Localhost_Check_Tests.php
@@ -29,4 +29,16 @@ class Localhost_Check_Tests extends WP_UnitTestCase {
 		$this->assertArrayHasKey( 'code', $errors['load.php'][0][0][0] );
 		$this->assertEquals( 'localhost_code_detected', $errors['load.php'][0][0][0]['code'] );
 	}
+
+	public function test_run_without_errors() {
+		$localhost_check = new Localhost_Check();
+		$check_context   = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-localhost-without-errors/load.php' );
+		$check_result    = new Check_Result( $check_context );
+
+		$localhost_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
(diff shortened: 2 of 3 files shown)


bordoni:
Thank you @nielslange, the code you are creating your Pull Request against is the new Codebase, and it wont be released on the plugin for a little bit so we cannot close the #267 using this PR unless the code you base your PR is from the `legacy-plugin` branch of this repository.

Sorry for the confusion there.

Now talking about the code, I wonder how much of a performance difference this particular approach has compared to the existing one? We need to consider that, since checks will run on plugins that are 10MB big.

includes/Checker/Checks/Localhost_Check.php
```diff
@@ -43,16 +43,31 @@ public function get_categories() {
...
-				__( 'Do not use Localhost/127.0.0.1 in your code.', 'plugin-check' ),
-				array(
-					'code' => 'localhost_code_detected',
-					'file' => $file,
-				)
-			);
+
+		foreach ( $php_files as $file ) {
+			$code   = file_get_contents( $file );
+			$tokens = token_get_all( $code );
```
swissspidy, line 49: I think this would be the first sniff using `token_get_all()`. While it's probably fine, I wonder whether this could be covered by a dedicated PHPCS sniff that can detect disallowed strings more precisely. Maybe such a sniff already exists?

includes/Checker/Checks/Localhost_Check.php
```diff
@@ -43,16 +43,31 @@ public function get_categories() {
...
+								'file' => $file,
+							)
+						);
+						break;
+					}
+				}
+			}
 		}
 	}
+
```
mukeshpanchal27, line 72:
```suggestion
```

Remove extra line.

includes/Checker/Checks/Localhost_Check.php
```diff
@@ -43,16 +43,31 @@ public function get_categories() {
 	 */
 	protected function check_files( Check_Result $result, array $files ) {
```
mukeshpanchal27, line 44: IMO we should update the code like below:
- Find all the files which have localhost/127.0.0.1 in code base.
- Run `token_get_all` one filter files.
- Check localhost/127.0.0.1 in tokens.

Code:
```
protected function check_files( Check_Result $result, array $files ) {
	$php_files = self::filter_files_by_extension( $files, 'php' );
	$file      = self::file_preg_match( '#https?://(localhost|127.0.0.1)#', $php_files );
	if ( $file ) {
		$code   = file_get_contents( $file );
		$tokens = token_get_all( $code );

		foreach ( $tokens as $token ) {
			if ( is_array( $token ) ) {
				if ( in_array( $token[0], array( T_COMMENT, T_DOC_COMMENT ), true ) ) {
					continue;
				}

				if ( preg_match( '#https?://(localhost|127.0.0.1)#', $token[1] ) ) {
					$result->add_message(
						true,
						__( 'Do not use Localhost/127.0.0.1 in your code.', 'plugin-check' ),
						array(
							'code' => 'localhost_code_detected',
							'file' => $file,
						)
					);
					break;
				}
			}
		}
	}
}
```

mukeshpanchal27 (commented):
Thank you, @nielslange, for the pull request. I apologize for the delayed review. I have provided some feedback for your consideration.

In response to @swissspidy's feedback, I invested some time in attempting to identify the issue. However, rather than utilizing a "sniff", we can readily achieve the desired result through the code.

includes/Checker/Checks/Localhost_Check.php
ernilambar: Such dedicated PHPCS sniff does not exist. I suggested for such sniff here https://github.com/PHPCSStandards/PHPCSExtra/issues/302 but it has been closed as this would too generalized and very difficult to maintain and support. So I believe PHPCS approach is no go this scenario.

ernilambar:
I used `time` command to find the time taken for completion of the command. But it seems `time` gives different output for same command run consecutively. So, I ran commands 3 times and calculate the average time. This does not seem to be reliable test but I did not find much significant drops in performance when trying with three big WordPress plugins.

Before PR:

* WooCommerce - 7.54s
* Yoast SEO - 2.67s
* EDD - 4.99s

After PR:

* WooCommerce - 6.45s
* Yoast SEO - 3.12s
* EDD - 3.76s

ernilambar:
While checking this PR, I found another issue. https://github.com/WordPress/plugin-check/issues/362

swissspidy:
I previously suggested using a dedicated PHPCS sniff for this, and I think https://github.com/WPTT/WPThemeReview/blob/15684d0852fe90d807c2ae7746dea1302b74b4bd/WPThemeReview/Sniffs/Privacy/ShortenedURLsSniff.php could be a great inspiration here.

ernilambar:
Since we are now adding custom sniff in PCP itself, I would also recommend to create custom sniff for this issue. Check will be more robust with sniff I believe.

davidperezgar:
Yes, I'm agree with @ernilambar

ernilambar:
@nielslange Thanks for the PR and contribution. Closing this now as sniff approach looked more appropriate now.

Closing in favor of https://github.com/WordPress/plugin-check/pull/743

PR #282, opened 2023-09-25
Add application files support to `File_Type_Check`

mukeshpanchal27:
This PR address third bullet point in #253 issue.

### Checklist:


includes/Checker/Checks/File_Type_Check.php
```diff
@@ -178,6 +182,32 @@ protected function look_for_hidden_files( Check_Result $result, array $files ) {
...
+			foreach ( $application_files as $file ) {
+				$result->add_message(
+					$is_error,
+					__( 'Application files are not permitted.', 'plugin-check' ),
+					array(
+						'code' => 'application_detected',
+						'file' => str_replace( $result->plugin()->path(), '', $file ),
+					)
+				);
+			}
```
felixarntz: Curious why this should only be an error in production? Is that the case for the code from https://github.com/WordPress/plugin-check/commit/b081cf7aa0ccc32e158d803b8905a66914c3f086? We don't do that in the other tests in this check, which is why I'm asking.

It would also allow us to use the existing reusable function, getting rid of some code here.
```suggestion
			foreach ( $application_files as $file ) {
				$this->add_result_error_for_file( $result, $file, 'application_detected', __( 'Application files are not permitted.', 'plugin-check' ) );
			}
```

tests/phpunit/Checker/Checks/File_Type_Check_Tests.php
```diff
@@ -79,6 +79,30 @@ public function test_run_with_vcs_dir_errors() {
...
+		$check = new File_Type_Check( File_Type_Check::TYPE_APPLICATION );
+		$check->run( $check_result );
+
+		if ( ( ! defined( 'WP_DEBUG' ) || ! WP_DEBUG ) && 'production' === wp_get_environment_type() ) {
+			$problems      = $check_result->get_errors();
+			$problem_count = $check_result->get_error_count();
+		} else {
+			$problems      = $check_result->get_warnings();
+			$problem_count = $check_result->get_warning_count();
+		}
```
felixarntz: See above, curious why you went with this choice?

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -178,6 +182,32 @@ protected function look_for_hidden_files( Check_Result $result, array $files ) {
...
+	/**
+	 * Looks for application files and amends the given result with an error if found.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @param array        $files  List of absolute file paths.
+	 */
+	protected function look_for_application_files( Check_Result $result, array $files ) {
+		$application_files = self::filter_files_by_extensions( $files, array( 'a', 'bin', 'bpk', 'deploy', 'dist', 'distz', 'dmg', 'dms', 'DS_Store', 'dump', 'elc', 'exe', 'iso', 'lha', 'lrf', 'lzh', 'o', 'obj', 'phar', 'pkg', 'sh', 'so' ) );
```
felixarntz: Nit-pick, but since this is such a long array, better to break it into separate lines. At a minimum, we should put the array on its own line to improve readability.
```suggestion
		$application_files = self::filter_files_by_extensions(
			$files,
			array( 'a', 'bin', 'bpk', 'deploy', 'dist', 'distz', 'dmg', 'dms', 'DS_Store', 'dump', 'elc', 'exe', 'iso', 'lha', 'lrf', 'lzh', 'o', 'obj', 'phar', 'pkg', 'sh', 'so' )
		);
```

felixarntz (commented):
@mukeshpanchal27 This looks great, just a few minor points of feedback.

includes/Checker/Checks/File_Type_Check.php
mukeshpanchal27: Thanks @felixarntz for the feedback. Yes i refer the code from upstream. I have updated in https://github.com/10up/plugin-check/pull/282/commits/e4d41159865c805a3e428d09f45d03cc79cc985e

felixarntz (approved):
Thanks @mukeshpanchal27 for the update, LGTM!

mukeshpanchal27:
@joemcgill, @swissspidy, @spacedmonkey, or @bordoni, could one of you please review this PR so that we can proceed with merging it? Thanks!

swissspidy (approved):

PR #285, opened 2023-10-03
Add `Trademarks_Check`

mukeshpanchal27:
This PR added last point for the issue #253.

### Checklist:

### DIFF
--- a/includes/Checker/Checks/Trademarks_Check.php
+++ b/includes/Checker/Checks/Trademarks_Check.php
@@ -164,7 +164,7 @@ class Trademarks_Check extends Abstract_File_Check {
 	const TYPE_README = 1;
 	const TYPE_NAME   = 2;
 	const TYPE_SLUG   = 4;
-	const TYPE_ALL     = 7; // Same as all of the above with bitwise OR.
+	const TYPE_ALL    = 7; // Same as all of the above with bitwise OR.
 
 	/**
 	 * Bitwise flags to control check behavior.
--- a/includes/Checker/Checks/Trademarks_Check.php
+++ b/includes/Checker/Checks/Trademarks_Check.php
@@ -244,11 +244,14 @@ class Trademarks_Check extends Abstract_File_Check {
 			$readme_list,
 			function ( $file ) use ( $plugin_relative_path ) {
 				$file = str_replace( $plugin_relative_path, '', $file );
-				if ( ! strpos( $file, '/' ) ) {
-					return true;
-				}
+				return ! str_contains( $file, '/' );
 			}
 		);
+
+		// If the readme file does not exist, then skip test.
+		if ( empty( $potential_readme_files ) ) {
+			return;
+		}
 
 		// Find the .txt versions of the readme files.
 		$readme_txt = array_filter(
@@ -258,31 +261,17 @@ class Trademarks_Check extends Abstract_File_Check {
 			}
 		);
 
-		// Find the .md versions of the readme files.
-		$readme_md = array_filter(
-			$potential_readme_files,
-			function ( $file ) {
-				return preg_match( '/^readme\.md$/i', basename( $file ) );
-			}
-		);
-
-		// If there's a .txt version, ignore .md versions.
-		$readme = ( ! empty( $readme_txt ) ) ? $readme_txt : $readme_md;
-
-		// If the readme file does not exist, then skip test.
-		if ( empty( $readme ) ) {
-			return;
-		}
+		$readme = $readme_txt ? $readme_txt : $potential_readme_files;
 
 		$matches = array();
 		// Get the plugin name from readme file.
-		$file = self::file_preg_match( '/===(.*)===/i', $files, $matches );
+		$file = self::file_preg_match( '/===(.*)===/i', $readme, $matches );
 
-		if ( ! $file ) {
+		if ( ! $file || ! isset( $matches[1] ) ) {
 			return;
 		}
 
-		$name = isset( $matches[1] ) ? $matches[1] : '';
+		$name = $matches[1];
 
 		try {
 			$this->validate_name_has_no_trademarks( $name );
@@ -342,11 +331,7 @@ class Trademarks_Check extends Abstract_File_Check {
 	 * @throws Exception Thrown if we found trademarked term in plugin name.
 	 */
 	private function validate_name_has_no_trademarks( $plugin_name ) {
-		if ( empty( $plugin_name ) ) {
-			return;
-		}
-
-		$check = $this->has_trademarked_slug( $plugin_name );
+		$check = $this->has_trademarked_slug( sanitize_title_with_dashes( $plugin_name ) );
 		if ( ! $check ) {
 			return;
 		}
@@ -393,10 +378,6 @@ class Trademarks_Check extends Abstract_File_Check {
 	 * @throws Exception Thrown if we found trademarked term in plugin slug.
 	 */
 	private function validate_slug_has_no_trademarks( $plugin_slug ) {
-		if ( empty( $plugin_slug ) ) {
-			return;
-		}
-
 		$check = $this->has_trademarked_slug( $plugin_slug );
 		if ( ! $check ) {
 			return;
@@ -443,8 +424,10 @@ class Trademarks_Check extends Abstract_File_Check {
 	 * @return string|false The trademark slug if found, false otherwise.
 	 */
 	private function has_trademarked_slug( $slug ) {
-		// We work on slugs for this check.
-		$slug = sanitize_title_with_dashes( $slug );
+		// Bail early if the plugin slug not provided.
+		if ( empty( $slug ) ) {
+			return false;
+		}
 
 		$has_trademarked_slug = false;
 
--- a/includes/Checker/Checks/Trademarks_Check.php
+++ b/includes/Checker/Checks/Trademarks_Check.php
@@ -161,9 +161,9 @@ class Trademarks_Check extends Abstract_File_Check {
 		'woo',
 	);
 
-	const CHECK_README = 1;
-	const CHECK_NAME   = 2;
-	const CHECK_SLUG   = 4;
+	const TYPE_README = 1;
+	const TYPE_NAME   = 2;
+	const TYPE_SLUG   = 4;
 	const TYPE_ALL     = 7; // Same as all of the above with bitwise OR.
 
 	/**
@@ -209,17 +209,17 @@ class Trademarks_Check extends Abstract_File_Check {
 	protected function check_files( Check_Result $result, array $files ) {
 
 		// Check the trademarks in readme file plugin name.
-		if ( $this->flags & self::CHECK_README ) {
+		if ( $this->flags & self::TYPE_README ) {
 			$this->check_for_readme( $result, $files );
 		}
 
 		// Check the trademarks in plugin name.
-		if ( $this->flags & self::CHECK_NAME ) {
+		if ( $this->flags & self::TYPE_NAME ) {
 			$this->check_for_name( $result );
 		}
 
 		// Check the trademarks in plugin slug.
-		if ( $this->flags & self::CHECK_SLUG ) {
+		if ( $this->flags & self::TYPE_SLUG ) {
 			$this->check_for_slug( $result );
 		}
 	}
--- a/includes/Checker/Checks/Trademarks_Check.php
+++ b/includes/Checker/Checks/Trademarks_Check.php
@@ -7,6 +7,7 @@
 
 namespace WordPress\Plugin_Check\Checker\Checks;
 
+use Exception;
 use WordPress\Plugin_Check\Checker\Check_Categories;
 use WordPress\Plugin_Check\Checker\Check_Result;
 use WordPress\Plugin_Check\Traits\Stable_Check;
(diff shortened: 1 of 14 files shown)


mukeshpanchal27:
@felixarntz @spacedmonkey @joemcgill @bordoni and @swissspidy PR is ready for review.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,482 @@
...
+
+	/**
+	 * Checks the trademarks in readme file plugin name.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The Check Result to amend.
+	 * @param array        $files  Array of plugin files.
+	 */
+	private function check_for_readme( Check_Result $result, array $files ) {
```
swissspidy: The following code is basically duplicated from the readme check.

Should we just move the trademark check to the `Readme_Check`? Or must it be separate?

If separate, can we perhaps find a way to reuse the logic to retrieve the readme file?

tests/phpunit/Checker/Checks/Trademarks_Check_Tests.php
```diff
@@ -0,0 +1,109 @@
...
+		$this->assertEmpty( $warnings );
+
+		$this->assertEquals( 0, $check_result->get_error_count() );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
+	}
+
+	/**
+	 * @dataProvider data_trademarks_check
+	 */
+	public function test_trademarks_with_different_scenarios( $type_flag, $plugin_basename, $expected_file, $expected_code ) {
```
swissspidy: Instead of passing `$expected_file` and `$expected_code` and doing this conditional check below, it might be easier to read if the data provider returned an array of `$expected_errors`, so that we could compare the length & content of the expected errors with the actual errors & error count. WDYT?

includes/Checker/Checks/Trademarks_Check.php
mukeshpanchal27: Thank you, @swissspidy, for bringing this up. Would it be possible to handle this refactoring in a follow-up PR to avoid making changes to other files within this current PR?

swissspidy: I think first we would need to decide on the approach: keep two checks (but with shared code), or merge the checks into one. If the latter, it doesn't really make sense to do it in a follow-up PR. Let's hear others :)

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,482 @@
...
+	 * Determines if we find a trademarked term.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result   The Check Result to amend.
+	 * @param string       $file     The plugin file.
+	 * @param string       $input    The plugin name or slug.
+	 * @param string       $preamble The preamble message.
+	 */
+	private function look_for_trademark( $result, $file, $input, $preamble ) {
```
felixarntz: This method does too many things at once and therefore becomes overly complicated. A few problems I see in this method:
* I think we should simplify the `$input` parameter. It is confusing that it can be either a slug or a name, especially because this will lead to confusing error messages below. If you pass a slug for `$input`, the messages will still say "Your plugin name - ...", although it is actually a plugin slug. We should probably rename this parameter to `$plugin_name` and the method to `look_for_trademarked_name()`. For the slug check, we should then use another method like `look_for_trademarked_slug()`, and move the common parts into lower-level methods that both can share.
* The `$preamble` parameter isn't necessary. As far as I can tell, the message for it is always the same, so we should just put it directly into the function.
* The messages should probably also be different for the slug specific check so that they are not confusing about what the problem is.

Maybe you can further simplify this by not requiring the `$result` and `$file` parameters even and simply throwing exceptions for any trademark problems. For example as follows:
* `validate_name_has_no_trademarks( $plugin_name )` (throws exceptions with the specific messages for name problems)
* `validate_slug_has_no_trademarks( $plugin_slug )` (throws exceptions with the specific messages for slug problems)
* Implement the `add_result_error_for_file()` method here that other checks have too.
* In the other methods, you can then use something like:
```php
try {
	$this->validate_name_has_no_trademarks( $plugin_name );
	// Or use $this->validate_slug_has_no_trademarks( $plugin_slug );
} catch ( Exception $e ) {
	$this->add_result_error_for_file( $result, $file, 'trademarked_term', $e->getMessage() );
}
```

This is so little logic that it isn't really duplicate code.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,482 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The Check Result to amend.
+	 * @param array        $files  Array of plugin files.
+	 */
+	protected function check_files( Check_Result $result, array $files ) {
+
+		// Check the trademarks in readme file plugin name.
+		if ( $this->flags & self::TYPE_README ) {
+			$this->check_for_readme( $result, $files );
```
felixarntz: Maybe rename to clarify: Something like `check_for_name_in_readme()`?

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,482 @@
...
+		}
+
+		// Check the trademarks in plugin name.
+		if ( $this->flags & self::TYPE_NAME ) {
+			$this->check_for_name( $result );
```
felixarntz: Maybe `check_for_name_in_main_file()`?

includes/Checker/Checks/Trademarks_Check.php
felixarntz: I agree we should avoid the duplicate code here, though I don't feel strongly whether to do it now or separately. There are a few other things that come up while working on checks that should be abstracted out (e.g. into a trait), and I think I'm personally leaning towards doing that in individual issues as it would touch other checks as well.

The most important thing for this kind of feedback is that, if we don't address it here, we should open a dedicated follow up issue so that it's not forgotten.

felixarntz (commented):
This looks good so far @mukeshpanchal27. A bit of feedback below.

includes/Checker/Checks/Trademarks_Check.php
mukeshpanchal27: Thanks both for the feedback. Open followup issue #286

tests/phpunit/Checker/Checks/Trademarks_Check_Tests.php
mukeshpanchal27: Thank you, @swissspidy, for your feedback. Upon further examination, I've identified some challenges with the data provider in our current unit tests. We are currently testing three different scenarios:

1. A readme plugin name check with an error.
2. A plugin name check with an error.
3. A plugin name with 'for WooCommerce' without an error.

includes/Checker/Checks/Trademarks_Check.php
mukeshpanchal27: Thank you for the valuable feedback. I have updated the code in accordance with your guidance, which can be found at https://github.com/WordPress/plugin-check/pull/285/commits/c43ee3cdf220dbf7c769ac1175095b5048e3e31f.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+		$readme_list = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+
+		// Filter the readme files located at root.
+		$potential_readme_files = array_filter(
+			$readme_list,
+			function ( $file ) use ( $plugin_relative_path ) {
+				$file = str_replace( $plugin_relative_path, '', $file );
+				if ( ! strpos( $file, '/' ) ) {
+					return true;
+				}
```
felixarntz: A bit cleaner, we should also return `false` here. Also, maybe better to use `str_contains()`.
```suggestion
				return ! str_contains( $file, '/' );
```

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+			}
+		);
+
+		// If there's a .txt version, ignore .md versions.
+		$readme = ( ! empty( $readme_txt ) ) ? $readme_txt : $readme_md;
+
+		// If the readme file does not exist, then skip test.
+		if ( empty( $readme ) ) {
+			return;
+		}
```
felixarntz: I think we can simplify this logic:
1. Move the `empty()` check to the top of this block, checking that for `$potential_readme_files`. If that list is empty, we can bail early. If it's not empty, we know _for sure_ there is either `readme.txt` or `readme.md`.
2. Remove the `$readme_md = array_filter( ...` logic. We don't need to check that, because if `$readme_txt` is empty, we know that `$potential_readme_files` only contains a `readme.md`.

So I think you can do something like:
```php
if ( $readme_txt ) {
	$readme = $readme_txt;
} else {
	$readme = $potential_readme_files;
}
```

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+
+		$matches = array();
+		// Get the plugin name from readme file.
+		$file = self::file_preg_match( '/===(.*)===/i', $files, $matches );
```
felixarntz: We shouldn't pass _all_ `$files` here, but only the relevant readme file.
```suggestion
		$file = self::file_preg_match( '/===(.*)===/i', $readme, $matches );
```

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+
+		if ( ! $file ) {
+			return;
+		}
+
+		$name = isset( $matches[1] ) ? $matches[1] : '';
```
felixarntz: This check strictly isn't needed, since `$matches[1]` should always be set unless `! $file`, in which case this logic isn't reached. But still +1 for defensive coding here. My only suggestion then would be to bail early in that case too, since an empty name wouldn't make sense to check.
```suggestion
		if ( ! $file || ! isset( $matches[1] ) ) {
			return;
		}

		$name = $matches[1];
```

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+	 * Whether the plugin uses a trademark in the slug.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $slug The plugin slug.
+	 * @return string|false The trademark slug if found, false otherwise.
+	 */
+	private function has_trademarked_slug( $slug ) {
+		// We work on slugs for this check.
+		$slug = sanitize_title_with_dashes( $slug );
```
felixarntz: I think we should remove this from here, as the method expects a slug anyway. This function should only be called when passing a plugin _name_ to `has_trademarked_slug()`.

Instead, here we could add the check to bail early (with `return false`) if `empty( $slug )`.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_name The plugin name.
+	 *
+	 * @throws Exception Thrown if we found trademarked term in plugin name.
+	 */
+	private function validate_name_has_no_trademarks( $plugin_name ) {
+		if ( empty( $plugin_name ) ) {
+			return;
+		}
```
felixarntz: Instead of having an `empty()` check here _and_ in `validate_slug_has_no_trademarks()`, you could centralize that in `has_trademarked_slug()`.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param string $plugin_slug The plugin slug.
+	 *
+	 * @throws Exception Thrown if we found trademarked term in plugin slug.
+	 */
+	private function validate_slug_has_no_trademarks( $plugin_slug ) {
+		if ( empty( $plugin_slug ) ) {
+			return;
+		}
```
felixarntz: See above, this could be removed from here in favor of adding it in `has_trademarked_slug()`.

includes/Checker/Checks/Trademarks_Check.php
```diff
@@ -0,0 +1,543 @@
...
+	 * @param string $plugin_name The plugin name.
+	 *
+	 * @throws Exception Thrown if we found trademarked term in plugin name.
+	 */
+	private function validate_name_has_no_trademarks( $plugin_name ) {
+		if ( empty( $plugin_name ) ) {
+			return;
+		}
+
+		$check = $this->has_trademarked_slug( $plugin_name );
```
felixarntz: See my other comment below:
```suggestion
		$check = $this->has_trademarked_slug( sanitize_title_with_dashes( $plugin_name ) );
```

felixarntz (commented):
@mukeshpanchal27 Apologies for the delay in review on this one. This is looking really good, I left a few follow up comments, but I think it's close to merge.

mukeshpanchal27:
Thank you, @felixarntz, for your valuable feedback. I will implement the necessary changes related to the README when we start working on the `Find_Readme` trail, as discussed in the following comment: https://github.com/WordPress/plugin-check/issues/286#issuecomment-1760083800.

felixarntz (approved):

@swissspidy Can you please take another look at this next week when you get a chance?

tests/phpunit/Checker/Checks/Trademarks_Check_Tests.php
swissspidy: What I was suggesting here is to update the data provider to return the expected error for each file instead of just expected error codes. This way we don't need the conditional checks in the test itself, making the test more readable. Does that make sense to you?

swissspidy (approved):
Still a suggestion to improve the data provider, but no objections otherwise

tests/phpunit/Checker/Checks/Trademarks_Check_Tests.php
mukeshpanchal27: @swissspidy In https://github.com/WordPress/plugin-check/pull/285/commits/eb3f667a507f64283218c8a35ff1defa305a5970 i have update data provider as requested.

PR #295, opened 2023-10-18
Refactoring and Code Duplication Reduction for `Plugin_Readme_Check` and `Trademarks_Check`

### LINKED ISSUE #286: Refactoring and Code Duplication Reduction for `Plugin_Readme_Check` and `Trademarks_Check`
In pull request #285, we introduced the `Trademarks_Check` check, which reuses code identical to that found in the `Plugin_Readme_Check`. This duplicate code is a potential source of maintenance challenges and can introduce bugs. It is essential that we address this issue promptly to improve the overall quality of our codebase.

mukeshpanchal27:
Closes #286

This pull request introduces two trait classes, which enable us to eliminate code duplications and reuse them throughout the codebase:
1. `Amend_Check_Result`
2. `Find_Readme`

### Checklist:

### DIFF
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -61,8 +61,6 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 	 *
 	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
 	 *                   the check).
-	 *
-	 * @SuppressWarnings(PHPMD.NPathComplexity)
 	 */
 	final public function run( Check_Result $result ) {
 		// Include the PHPCS autoloader.
@@ -127,25 +125,15 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 			}
 
 			foreach ( $file_results['messages'] as $file_message ) {
-				if ( strtoupper( $file_message['type'] ) === 'ERROR' ) {
-					$this->add_result_error_for_file(
-						$result,
-						$file_message['message'],
-						$file_message['source'],
-						$file_name,
-						$file_message['line'],
-						$file_message['column']
-					);
-				} else {
-					$this->add_result_warning_for_file(
-						$result,
-						$file_message['message'],
-						$file_message['source'],
-						$file_name,
-						$file_message['line'],
-						$file_message['column']
-					);
-				}
+				$this->add_result_message_for_file(
+					$result,
+					strtoupper( $file_message['type'] ) === 'ERROR',
+					$file_message['message'],
+					$file_message['source'],
+					$file_name,
+					$file_message['line'],
+					$file_message['column']
+				);
 			}
 		}
 	}
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -11,6 +11,7 @@ use Exception;
 use PHP_CodeSniffer\Runner;
 use WordPress\Plugin_Check\Checker\Check_Result;
 use WordPress\Plugin_Check\Checker\Static_Check;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
 use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
 /**
@@ -19,6 +20,8 @@ use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
  * @since n.e.x.t
  */
 abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
+
+	use Amend_Check_Result;
 
 	/**
 	 * List of allowed PHPCS arguments.
@@ -58,6 +61,8 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 	 *
 	 * @throws Exception Thrown when the check fails with a critical error (unrelated to any errors detected as part of
 	 *                   the check).
+	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
 	 */
 	final public function run( Check_Result $result ) {
 		// Include the PHPCS autoloader.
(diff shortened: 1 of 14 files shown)


includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -122,16 +127,25 @@ final public function run( Check_Result $result ) {
...
+				} else {
+					$this->add_result_warning_for_file(
+						$result,
+						$file_message['message'],
+						$file_message['source'],
+						$file_name,
+						$file_message['line'],
+						$file_message['column']
+					);
+				}
```
swissspidy: Code like this would be much simpler if we just had 1 method with a `$severity` parameter, instead of two separate methods that are nearly identical.

That would also improve `Result::add_message()` as it no longer uses a bool flag but a proper enum for the severity.

felixarntz: Good point @swissspidy. I would suggest we introduce a third message in the trait, maybe `add_result_message_for_file()`, with the additional parameter. We can keep the `add_result_error_for_file()` and `add_result_warning_for_file()` as convenience wrappers, and they would just call the new `add_result_message_for_file()` with the corresponding extra parameter.

That would avoid the almost duplicated code between the two functions, would simplify code like the one here, while still provide more intuitive methods for checks that _only_ will add errors or warnings and therefore don't need the extra parameter differentiation.

felixarntz: Regarding whether to use a `$severity` parameter instead of the `bool $error`, I'm not sure we need that. While I agree boolean parameters aren't a great practice in general, the approach of this `bool $error` parameter was taken from PHPCodeSniffer's codebase (which heavily inspired `Check_Result` as a whole).

Since we only have errors and warnings, I'm not sure we need a more detailed severity at this point. If we introduced that, I think we should also change `Check_Result` to no longer have that boolean parameter, in order to have a consistent codebase, and that would increase the scope of this issue. So I'd prefer to leave that out and consider separately.

includes/Traits/Amend_Check_Result.php
```diff
@@ -0,0 +1,68 @@
...
+			false,
+			$message,
+			array(
+				'code'   => $code,
+				'file'   => str_replace( $result->plugin()->path(), '', $file ),
+				'line'   => $line,
+				'column' => $column,
+			)
+		);
+	}
```
felixarntz, line 75: See above: Instead of the duplicate code between these two methods, let's add a third one that allows to provide the additional parameter for whether it's an error or warning. Then the two methods here can become simple wrapper methods for the third method.

felixarntz (commented):
@mukeshpanchal27 Thank you for the PR. It looks great, my only feedback relates to @swissspidy's existing feedback, I think we can reduce a bit of duplicate code.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
mukeshpanchal27: Thanks for the feedback. Addressed in https://github.com/WordPress/plugin-check/pull/295/commits/95e863e339a7c7d757aa2d2076b11cadcc0e21b7

swissspidy (approved):

includes/Checker/Checks/Code_Obfuscation_Check.php
```diff
@@ -94,7 +96,12 @@ protected function check_files( Check_Result $result, array $files ) {
 	protected function look_for_zendguard( Check_Result $result, array $php_files ) {
 		$obfuscated_file = self::file_preg_match( '/(<\?php \@Zend;)|(This file was encoded by)/', $php_files );
 		if ( $obfuscated_file ) {
-			$this->add_result_error_for_file( $result, $obfuscated_file, 'Zend Guard' );
+			$this->add_result_error_for_file(
+				$result,
+				__( 'Code Obfuscation tools are not permitted. Detected: Zend Guard', 'plugin-check' ),
```
felixarntz: Better to keep this a single translation string regardless of the specific tool. The tool name shouldn't matter for the translation. In other words, maybe let's keep using the following approach here and below:
```php
sprintf(
	/* translators: %s: tool name */
	__( 'Code Obfuscation tools are not permitted. Detected: %s', 'plugin-check' ),
	$tool_name
)
```

includes/Checker/Checks/File_Type_Check.php
```diff
@@ -152,13 +164,12 @@ function ( $directory ) use ( $directories ) {
...
+				$this->add_result_message_for_file(
+					$result,
 					$is_error,
 					__( 'Version control checkouts should not be present.', 'plugin-check' ),
-					array(
-						'code' => 'vcs_present',
-						'file' => str_replace( $result->plugin()->path(), '', $dir ),
-					)
+					'vcs_present',
+					str_replace( $result->plugin()->path(), '', $dir )
```
felixarntz: The `str_replace()` is handled in the central method now and therefore not needed here.
```suggestion
					$dir
```

includes/Traits/Amend_Check_Result.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @param bool         $error   Whether it is an error or notice.
+	 * @param string       $message Error message.
+	 * @param string       $code    Error code.
+	 * @param string       $file    Absolute path to the file where the issue was found.
+	 * @param int          $line    The line on which the message occurred. Default is 0 (unknown line).
+	 * @param int          $column  The column on which the message occurred. Default is 0 (unknown column).
+	 */
+	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0 ) {
+		$result->add_message(
+			$error ? true : false,
```
felixarntz: Nit-pick:
```suggestion
			(bool) $error,
```

felixarntz (approved):
@mukeshpanchal27 Thank you for the updates, LGTM!

A few minor nit-picks below, if you can take a look, but not strictly blockers.

includes/Checker/Checks/Code_Obfuscation_Check.php
mukeshpanchal27: Thanks for the review. Updated in https://github.com/WordPress/plugin-check/pull/295/commits/28e8242ed5481fe6bf7bbe73ce04a6d762f922be

PR #298, opened 2023-10-23
Add link to Plugin Editor for each found issue to jump user to file and line of code

### LINKED ISSUE #251: Add link to Plugin Editor for each found issue to jump user to file and line of code

Currently when running PHPCS checks in the admin, it will mention the file and line number that an issue occurs on. However, it could go further by adding a link to open that line of code in the plugin editor so that the user could fix the issue right there (such as if they are a developer).

We implemented this in the AMP plugin, so that when an AMP validation error occurs the user can jump to the responsible code. We also provide a filter so that the user's own IDE can be opened to the responsible line of code.

Here's the logic for constructing the editor link: https://github.com/ampproject/amp-wp/blob/956c8d298764e4b22b936347d7c8b591b2fadfc2/includes/validation/class-amp-validation-error-taxonomy.php#L2356-L2473

Here's the logic for jumping to the line of code since core doesn't support that itself: https://github.com/ampproject/amp-wp/blob/956c8d298764e4b22b936347d7c8b591b2fadfc2/includes/validation/class-amp-validation-error-taxonomy.php#L1034-L1062

### Designs

_No response_

_No response_

### Code of Conduct

mukeshpanchal27:
Closes #251

### Checklist:


mukeshpanchal27:
This pull request adds the Plugin Editor link for all error/notice situations. However, there are certain cases where we need to exclude it:

- If a plugin contains `.phar` or `.gitignore` files, the Plugin Editor does not support these formats, and clicking the link will display a `Sorry, that file cannot be edited` message.
- If a plugin does not have a readme file, the plugin will display a `no_plugin_readme` message, but that file is not present in the folder, so it will show the same error as mentioned above.

I would appreciate hearing your thoughts on this matter before proceeding with the exclusion of these files or any other files that may require exclusion.

templates/results-row.php
```diff
@@ -14,5 +14,10 @@
 	<td>
 		{{data.message}}
 	</td>
+	<td>
+		<# if ( data.link ) { #>
+			<a href="{{data.link}}" title="<?php _e( 'View file in the plugin file editor.', 'plugin-check' ); ?>" aria-label="<?php _e( 'View file in the plugin file editor.', 'plugin-check' ); ?>" target="_blank"><?php _e( 'View in code editor', 'plugin-check' ); ?></a>
```
swissspidy: `title` attribute is bad for a11y, let's remove it.

mukeshpanchal27: Thanks for the review. Removed in https://github.com/WordPress/plugin-check/pull/298/commits/f323f66f638430e109ec07c63545a27fba93dae7

felixarntz (changes requested):
@mukeshpanchal27 This mostly looks good, though there's one critical problem as this is missing `current_user_can( 'edit_plugins' )` checks (also see https://github.com/WordPress/plugin-check/pull/262#discussion_r1334628438).

We must only show these links if the current user has the permission to edit plugins via the built-in file editor - which is something that is often not the case as many environments disable it.

swissspidy:
> We must only show these links if the current user has the permission to edit plugins via the built-in file editor - which is something that is often not the case as many environments disable it.

Should we still show the link if the link has been filtered to point to another editor like VS Code?

felixarntz:
@swissspidy 
> > We must only show these links if the current user has the permission to edit plugins via the built-in file editor - which is something that is often not the case as many environments disable it.
> 
> Should we still show the link if the link has been filtered to point to another editor like VS Code?

Good point, that would make sense.

templates/results-row.php
```diff
@@ -14,5 +14,10 @@
 	<td>
 		{{data.message}}
 	</td>
+	<td>
+		<# if ( data.link ) { #>
+			<a href="{{data.link}}" aria-label="<?php _e( 'View file in the plugin file editor.', 'plugin-check' ); ?>" target="_blank"><?php _e( 'View in code editor', 'plugin-check' ); ?></a>
```
westonruter:
```suggestion
			<a href="{{data.link}}" aria-label="<?php esc_attr_e( 'View file in the plugin file editor.', 'plugin-check' ); ?>" target="_blank"><?php esc_html_e( 'View in code editor', 'plugin-check' ); ?></a>
```

mukeshpanchal27:
@felixarntz @swissspidy, in the commit at [https://github.com/WordPress/plugin-check/pull/298/commits/733479ba08131f263610689f37454f24c47f71e1](https://github.com/WordPress/plugin-check/pull/298/commits/733479ba08131f263610689f37454f24c47f71e1), I have addressed all your concerns. Additionally, I have also added checks for the `DISALLOW_FILE_EDIT` constant, which many plugins use to disable the plugin editors. I have adjusted the result table accordingly. 

Could you please review these changes once more? Thank you!

includes/Traits/File_Editor_URL.php
```diff
@@ -0,0 +1,120 @@
...
+				}
+			}
+		}
+
+		// Fall back to using the plugin editor if no external editor is offered.
+		if (
+			! $edit_url
+			&& ! ( defined( 'DISALLOW_FILE_EDIT' ) && true === DISALLOW_FILE_EDIT )
+			&& current_user_can( 'edit_plugins' )
+		) {
```
felixarntz: We don't need to check for any of the `DISALLOW_FILE_EDIT` or `DISALLOW_FILE_MODS` constants here as they will automatically be covered by the `current_user_can( 'edit_plugins' )` check. See the [`map_meta_cap()` function](https://developer.wordpress.org/reference/functions/map_meta_cap/).
```suggestion
		if ( ! $edit_url && current_user_can( 'edit_plugins' ) ) {
```

templates/results-row.php
```diff
@@ -14,5 +14,10 @@
 	<td>
 		{{data.message}}
 	</td>
+	<# if ( data.link ) { #>
+		<td>
+			<a href="{{data.link}}" aria-label="<?php esc_attr_e( 'View file in the plugin file editor.', 'plugin-check' ); ?>" target="_blank"><?php esc_html_e( 'View in code editor', 'plugin-check' ); ?></a>
```
felixarntz: What is the benefit of an `aria-label` here over the regular link anchor text? Is that needed for accessibility? I may not be up to date with the requirements there, so it would be great if you could clarify.

I am concerned about the `_target="blank"` here from an accessibility perspective. If we want to keep it, we should make it obvious that this link opens in another window/tab with relevant screen reader text.

templates/results-row.php
```diff
@@ -14,5 +14,10 @@
 	<td>
 		{{data.message}}
 	</td>
+	<# if ( data.link ) { #>
+		<td>
```
felixarntz: I don't think we can output the `<td>` conditionally, it either needs to be always there or never. Otherwise, we'll create invalid table markup for cases where `data.link` is false-y.

I think maybe we should handle this differently: `data.link` is realistically always going to be set for every error/warning, _except_ when the current user cannot `edit_plugins`. So maybe we should check this more holistically before output the individual table as a whole.

templates/results-table.php
```diff
@@ -14,7 +14,7 @@
 			<td>
 				<?php esc_html_e( 'Code', 'plugin-check' ); ?>
 			</td>
-			<td>
+			<td colspan="2">
```
felixarntz: See my comment above, I'm not sure this works reliably.

I think we need to check holistically:
* If _any_ of the links within the table data is not empty, we should output this column in the table. Most likely that'll mean _all_ links within the table aren't empty.
* Otherwise, we should skip the column for the entire table.
* In that case, we should also probably add a table header column, e.g. "Link" or "Edit Link".

felixarntz (commented):
Thanks @mukeshpanchal27 for making the updates, sharing a bit of follow up feedback below.

We may need to holistically determine for each table whether or not to include a column for the edit link, by checking for whether _any_ of the rows include a non-empty link.

templates/results-table.php
mukeshpanchal27: Thank you, @felixarntz, for your feedback and explanation. Since we are managing the table through two different files, we should consider updating the table through JavaScript, even though it may not be my preferred method. I've attempted different approaches but haven't been successful, so I'll revisit the issue. Thanks!

felixarntz: Thanks @mukeshpanchal27. I think we could do it something like this:
* In `renderFileResults()`, check both `errors` and `warnings` for whether any of the items in them includes a non-empty `link`. Then we can set a new boolean variable `hasLinks` based on that.
* This variable we can then pass to the `plugin-check-results-table` template as another field, and based on that render or not render the additional column.
* We can then also pass the variable as a parameter to the two `renderResultRows()` calls.
* Then in that function we pass it as another field to the `plugin-check-results-row` template, and based on that always render or not render the additional column, regardless of what `link` is. Realistically, either all rows should have a `link` or none should have a `link`.

templates/results-row.php
mukeshpanchal27: Thanks for bringing this to my attention. I have removed the `aria-label` attribute and added screen reader text in https://github.com/WordPress/plugin-check/pull/298/commits/5df8da4698ccff1d0998f8435cbae17ee8a91435, following the approach used in WordPress.

templates/results-table.php
mukeshpanchal27: Thank you, @felixarntz. I have made the necessary updates based on your suggestions in https://github.com/WordPress/plugin-check/pull/298/commits/5df8da4698ccff1d0998f8435cbae17ee8a91435

templates/results-row.php
```diff
@@ -14,5 +14,13 @@
 	<td>
 		{{data.message}}
 	</td>
+	<# if ( data.hasLinks ) { #>
+		<td>
+			<a href="{{data.link}}"<?php echo ( ! has_filter( 'wp_plugin_check_validation_error_source_file_editor_url_template' ) ) ? ' target="_blank"' : '' ?>>
+				<?php esc_html_e( 'View in code editor', 'plugin-check' ); ?>
+				<span class="screen-reader-text"><?php esc_html_e( '(opens in a new tab)', 'plugin-check' ); ?></span>
+			</a>
```
felixarntz: I think this still needs to be wrapped in a `data.link` if check, to make sure we're not rendering an `<a>` tag that doesn't go anywhere. Only the `<td>` must be rendered regardless of `data.link`.

templates/results-row.php
```diff
@@ -14,5 +14,13 @@
 	<td>
 		{{data.message}}
 	</td>
+	<# if ( data.hasLinks ) { #>
+		<td>
+			<a href="{{data.link}}"<?php echo ( ! has_filter( 'wp_plugin_check_validation_error_source_file_editor_url_template' ) ) ? ' target="_blank"' : '' ?>>
```
felixarntz: Why only make the link external when it's another editor, and not when it's the plugin file editor? I'm not sure about it, but am thinking opening in a new tab would be a better experience as otherwise you would lose where you were in the original plugin checker report.

My original feedback on accessibility was primarily about the screen reader text that we need to have on external links. Note that right now this is inconsistent either way, since the screen reader text should only be there when the link also opens an external window (i.e. has `target="_blank"`).

I would suggest to simply remove the condition here, as it may be unreliable anyway - different links may or may not use the filter, we don't know that just by checking whether there is a filter.
```suggestion
			<a href="{{data.link}}" target="_blank">
```

felixarntz (commented):
@mukeshpanchal27 Two points of follow up feedback regarding the markup.

templates/results-row.php
mukeshpanchal27: Thanks for the feedback. I have updated in https://github.com/WordPress/plugin-check/pull/298/commits/55d8780ec5afb884020e8f13d22002e50caf4c65

mukeshpanchal27:
@felixarntz @swissspidy @westonruter @joemcgill, I have addressed all the feedback, and the PR is now ready for review.

templates/results-row.php
```diff
@@ -14,5 +14,15 @@
 	<td>
 		{{data.message}}
 	</td>
+	<# if ( data.hasLinks ) { #>
+		<td>
+			<# if ( data.link ) { #>
```
swissspidy, line 19: Do we need both conditions? If `data.link` is true, then certainly `data.hasLinks` must be true also.

swissspidy (approved):
One question, otherwise looks good

templates/results-row.php
mukeshpanchal27: Thank you, @swissspidy, for your feedback. Regarding the `data.hasLinks` issue, it is always true. However, as an additional check, we've implemented the suggested approach as per the discussion at https://github.com/WordPress/plugin-check/pull/298#discussion_r1373364037.

swissspidy: Surely `data.hasLinks` is not always true, otherwise what is the point?

What I am suggesting is

```
<# if ( data.link ) { #>
		<td>
<a href="{{data.link}}" target="_blank">
					<?php esc_html_e( 'View in code editor', 'plugin-check' ); ?>
					<span class="screen-reader-text"><?php esc_html_e( '(opens in a new tab)', 'plugin-check' ); ?></span>
				</a>
		</td>
	<# } #>
```

Because if `data.link` is true, `data.hasLinks` has to be true as well.

felixarntz: @swissspidy The opposite is not true though. `data.link` may be `false`, but `data.hasLinks` could still be `true`. That's why the separate checks here are needed.

swissspidy: Argh, got it, brain fart 🤦

includes/Traits/File_Editor_URL.php
```diff
@@ -0,0 +1,116 @@
...
+	 *
+	 * @param Check_Result $result   The check result to amend, including the plugin context to check.
+	 * @param string       $filename Error file name.
+	 * @param int          $line     Line number of error.
+	 * @return string|null File editor URL or null if not available.
+	 */
+	protected function get_file_editor_url( Check_Result $result, $filename, $line ) {
+		if ( ! isset( $filename, $line ) ) {
+			return null;
+		}
```
westonruter: The phpdoc for `$filename` and `$line` don't indicate that they can be nullable. Therefore, is this condition valid? If it is, then should they be modified to be nullable in the phpdoc?

includes/Traits/File_Editor_URL.php
```diff
@@ -0,0 +1,116 @@
...
+		 * For a template to be considered, the string '{{file}}' must be present in the filtered value.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param string|null $editor_url_template Editor URL template. default null.
+		 */
+		$editor_url_template = apply_filters( 'wp_plugin_check_validation_error_source_file_editor_url_template', null );
+
+		// Supply the file path to the editor template.
+		if ( null !== $editor_url_template && str_contains( $editor_url_template, '{{file}}' ) ) {
```
westonruter: Better to check if this is a string, because if someone provided an array then a fatal error would ensue:

```suggestion
		if ( is_string( $editor_url_template ) && str_contains( $editor_url_template, '{{file}}' ) ) {
```

assets/js/plugin-check-admin.js
```diff
@@ -378,36 +378,63 @@
...
-	 * @param {string} type    The result type. Either ERROR or WARNING.
 	 * @param {Object} results The results object.
-	 * @param {Object} table   The HTML table to append a result row to.
+	 * @return {boolean} True if there are links, false otherwise.
+	 */
+	function hasLinksInResults( results ) {
+		for ( const line in results ) {
+			for ( const column in results[ line ] ) {
+				for ( let i = 0; i < results[ line ][ column ].length; i++ ) {
+					if ( results[ line ][ column ][ i ].link ) {
```
westonruter, line 411: I'm not familiar with the data structure, but the code here could perhaps be made less verbose/repetitious with `for`/`of` loops. Not important however.

westonruter (changes requested):

assets/js/plugin-check-admin.js
felixarntz: Not sure this could be optimized. Of course we could consider using nested array functions, but I think that would be more of a personal preference that wouldn't make this code faster or slower.

includes/Traits/File_Editor_URL.php
felixarntz: Fair point, I think we should remove this. If we include checks like this just because someone _could_ pass `null` (due to lack of parameter type hints), we would need to do that in almost _every_ function.

I do however have another related suggestion: Several of our checks are _not_ line-specific, as in the message data doesn't include a line (or rather have the line set to `0`). I think it would make sense to give `$line` a default value of `0` to clarify that it is optional, and then the method logic could be adjusted to not pass `line` if it doesn't have an actual (i.e. not `0`) value.

felixarntz: Per my other comment:
```suggestion
	 * @param int          $line     Optional. Line number of error. Default 0 (no specific line).
	 * @return string|null File editor URL or null if not available.
	 */
	protected function get_file_editor_url( Check_Result $result, $filename, $line = 0 ) {
```

includes/Traits/File_Editor_URL.php
```diff
@@ -0,0 +1,116 @@
...
+		// Fall back to using the plugin editor if no external editor is offered.
+		if ( ! $edit_url && current_user_can( 'edit_plugins' ) ) {
+			return add_query_arg(
+				array(
+					'plugin' => rawurlencode( $result->plugin()->basename() ),
+					'file'   => rawurlencode( $plugin_slug . '/' . $filename ),
+					'line'   => $line,
+				),
+				admin_url( 'plugin-editor.php' )
+			);
```
felixarntz: Per my other comment, since not all our checks provide specific lines:
```suggestion
			$query_args = array(
				'plugin' => rawurlencode( $result->plugin()->basename() ),
				'file'   => rawurlencode( $plugin_slug . '/' . $filename ),
			);
			if ( $line ) {
				$query_args['line'] = $line;
			}
			return add_query_arg(
				$query_args,
				admin_url( 'plugin-editor.php' )
			);
```

assets/js/plugin-check-admin.js
westonruter: For example:

```suggestion
		for ( const lineResult of results ) {
			for ( const columnResult of lineResult ) {
				for ( const result of columnResult ) {
					if ( columnResult.link ) {
```

Not really an optimization, but more a reduction in verbosity. 

The names of my variables could surely be improved.

includes/Traits/File_Editor_URL.php
mukeshpanchal27: Addressed in https://github.com/WordPress/plugin-check/pull/298/commits/0ef24763e1fde59359c84e1789b54bd2fcba7c50

assets/js/plugin-check-admin.js
mukeshpanchal27: Thank you for the feedback. This code change will result in the exclusion of certain code sniff results, even if the code contains the necessary link.

includes/Traits/File_Editor_URL.php
```diff
@@ -0,0 +1,116 @@
...
+						array(
+							'{{file}}',
+							'{{line}}',
+						),
+						array(
+							rawurlencode( $file_path ),
+							$line,
+						),
+						$editor_url_template
+					);
```
felixarntz, line 95: Here we're currently still passing `line` unconditionally, but I think changing that will require more thinking. I tend to think that overall these two filters are overly complicated to use and could be replaced with a single filter. But let's solve that via a separate issue to unblock this PR, which is more important than this specific extension point.

felixarntz (approved):
Thank you @mukeshpanchal27 for the updates, LGTM!

One other comment below, but I think we should solve that in a separate issue/PR.

includes/Traits/File_Editor_URL.php
mukeshpanchal27: Open followup issue #314.

mukeshpanchal27:
@westonruter, whenever you have some moment, could you please review this PR so that we can proceed with merging it? Thanks!

westonruter (approved):

PR #304, opened 2023-10-24
Implement persisting the enabled check categories checkboxes in user settings

### LINKED ISSUE #299: Implement persisting the enabled check categories checkboxes in user settings
The checkboxes for which check categories to run are currently non-persistent, which leads to poor UX: It is common for specific use-cases to rerun the same categories of checks over and over again. For this purpose, any time that the user changes the enabled checkboxes, this configuration should be silently persisted/saved so that the selection remains the same even after a page refresh.

For reference, see the checkboxes:
WordPress plugin checker UI in WP Admin

Preferably, the configuration should be saved into the user's settings:
* We can probably handle this on the client-side by relying on WordPress's global JS function `window.setUserSetting( name, value )` function (in the `utils` script which is enqueued throughout all WP Admin).
* Note that the function only accepts letters, numbers, hyphens, and underscores. So we should probably rely on the check category slugs and separate them with some special combination of allowed characters, e.g. `--`.
    * For example, if "Plugin repo" and "Security" are checked, we would store `plugin_repo--security`.
* We can then retrieve this value on the server side via the PHP function `get_user_setting()`, so there we'll have to split the string into its different values and then only mark the checkboxes as enabled based on the setting.

mukeshpanchal27:
Closes #299

### Checklist:

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -35,6 +35,31 @@
 	canRunChecks();
 	pluginsList.addEventListener( 'change', canRunChecks );
 
+	/**
+	 * Saves the user settings.
+	 *
+	 * @since n.e.x.t
+	 */
+	function saveUserSettings() {
+		const selectedCategories = [];
+
+		// Assuming you have a list of category checkboxes, find the selected ones.
+		categoriesList.forEach( function ( checkbox ) {
+			if ( checkbox.checked ) {
+				selectedCategories.push( checkbox.value );
+			}
+		} );
+
+		// Join the selected category slugs with '__' and save it as a user setting.
+		const settingValue = selectedCategories.join( '__' );
+		window.setUserSetting( 'category_preferences', settingValue );
+	}
+
+	// Attach the saveUserSettings function when a category checkbox is clicked.
+	categoriesList.forEach( function ( checkbox ) {
+		checkbox.addEventListener( 'change', saveUserSettings );
+	} );
+
 	// When the Check it button is clicked.
 	checkItButton.addEventListener( 'click', ( e ) => {
 		e.preventDefault();
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -47,7 +47,10 @@
 
 		// Join the selected category slugs with '__' and save it as a user setting.
 		const settingValue = selectedCategories.join( '__' );
-		window.setUserSetting( 'plugin_check_category_preferences', settingValue );
+		window.setUserSetting(
+			'plugin_check_category_preferences',
+			settingValue
+		);
 	}
 
 	// Attach the saveUserSettings function when a category checkbox is clicked.
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -35,11 +35,6 @@
 	canRunChecks();
 	pluginsList.addEventListener( 'change', canRunChecks );
 
-	/**
-	 * Saves the user settings.
-	 *
-	 * @since n.e.x.t
-	 */
 	function saveUserSettings() {
 		const selectedCategories = [];
 
@@ -52,7 +47,7 @@
 
 		// Join the selected category slugs with '__' and save it as a user setting.
 		const settingValue = selectedCategories.join( '__' );
-		window.setUserSetting( 'category_preferences', settingValue );
+		window.setUserSetting( 'plugin_check_category_preferences', settingValue );
 	}
 
 	// Attach the saveUserSettings function when a category checkbox is clicked.
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -165,7 +165,9 @@ final class Admin_Page {
 
 		$categories = Check_Categories::get_categories();
 
-		$user_settings = get_user_setting( 'category_preferences' );
+		// Get user settings for category preferences and set a default value to check all categories by default.
+		$user_settings = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
+		$user_settings = explode( '__', $user_settings );
 
 		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'templates/admin-page.php';
 	}
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -162,7 +162,7 @@ final class Admin_Page {
 
 		// Get user settings for category preferences and set a default value to check all categories by default.
 		$user_settings = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
-		$user_settings = explode( '__', $user_settings );
+		$user_settings = 'all_categories' === $user_settings ? $categories : explode( '__', $user_settings );
 
 		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'templates/admin-page.php';
 	}
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -161,8 +161,8 @@ final class Admin_Page {
 		$categories = Check_Categories::get_categories();
 
 		// Get user settings for category preferences and set a default value to check all categories by default.
-		$user_settings = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
-		$user_settings = 'all_categories' === $user_settings ? $categories : explode( '__', $user_settings );
+		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
+		$user_enabled_categories = 'all_categories' === $user_enabled_categories ? $categories : explode( '__', $user_enabled_categories );
 
 		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'templates/admin-page.php';
 	}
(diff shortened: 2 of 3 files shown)


assets/js/plugin-check-admin.js
```diff
@@ -35,6 +35,31 @@
...
+
+		// Assuming you have a list of category checkboxes, find the selected ones.
+		categoriesList.forEach( function ( checkbox ) {
+			if ( checkbox.checked ) {
+				selectedCategories.push( checkbox.value );
+			}
+		} );
+
+		// Join the selected category slugs with '__' and save it as a user setting.
+		const settingValue = selectedCategories.join( '__' );
```
mukeshpanchal27, line 49: @felixarntz, in the pull request, I opted to use `__` instead of `--` as the latter was causing inconsistencies when viewing the Plugin Check page on various browsers.

assets/js/plugin-check-admin.js
```diff
@@ -35,6 +35,31 @@
...
+		window.setUserSetting( 'category_preferences', settingValue );
```
swissspidy: `category_preferences` is very generic. Let's prefix this with `plugin_check_`.

assets/js/plugin-check-admin.js
```diff
@@ -35,6 +35,31 @@
 	canRunChecks();
 	pluginsList.addEventListener( 'change', canRunChecks );
 
+	/**
+	 * Saves the user settings.
+	 *
+	 * @since n.e.x.t
```
swissspidy: Probably not needed. This function is not exposed to global scope or anything.

assets/js/plugin-check-admin.js
swissspidy: Maybe a pipe `|` character as a delimiter? Seems more natural

templates/admin-page.php
```diff
@@ -45,7 +45,18 @@
...
+										if ( $user_settings ) {
+											$selected_categories = explode( '__', $user_settings );
+											if ( in_array( $category, $selected_categories, true ) ) {
+												$checked = true;
+											}
+										} else {
+											$checked = true;
+										}
+									?>
+									<input type="checkbox" id="<?php echo esc_attr( $category ); ?>" name="categories" value="<?php echo esc_attr( $category ); ?>"<?php echo ( $checked ? ' checked="checked"' : '' ); ?> />
```
swissspidy: You can use the `checked()` helper function for adding the `checked` attribute.

assets/js/plugin-check-admin.js
felixarntz: Other functions have those docs, so let's revisit that holistically.

assets/js/plugin-check-admin.js
felixarntz: @swissspidy We can only use `__` or `--`, as no other special characters are accepted for user settings.

includes/Admin/Admin_Page.php
```diff
@@ -150,16 +150,23 @@ private function get_available_plugins() {
...
-		global $available_plugins, $selected_plugin_basename, $categories;
+		global $available_plugins, $selected_plugin_basename, $categories, $user_settings;
 
 		$available_plugins = $this->get_available_plugins();
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
 		$categories = Check_Categories::get_categories();
 
+		$user_settings = get_user_setting( 'category_preferences' );
```
felixarntz: Let's handle the "exploding" into an array here, rather than in every `foreach` loop in the template.

templates/admin-page.php
```diff
@@ -45,7 +45,18 @@
...
+									<?php
+										$checked = false;
+										if ( $user_settings ) {
+											$selected_categories = explode( '__', $user_settings );
+											if ( in_array( $category, $selected_categories, true ) ) {
+												$checked = true;
+											}
+										} else {
+											$checked = true;
+										}
```
felixarntz: See my comment above, doing this for every category in the template isn't efficient, we should rather handle it when retrieving the user setting, so that `$user_settings` is already the array of checked categories.

Then we can simplify this here to:
```suggestion
										$checked = in_array( $category, $user_settings, true );
```

includes/Admin/Admin_Page.php
```diff
@@ -150,16 +150,23 @@ private function get_available_plugins() {
...
 	 * @since n.e.x.t
+	 *
+	 * @global array  $available_plugins        The list of available plugins.
+	 * @global string $selected_plugin_basename The selected plugin basename.
+	 * @global array  $categories               An array of categories.
+	 * @global mixed  $user_settings            The user interface setting value.
 	 */
 	public function render_page() {
-		global $available_plugins, $selected_plugin_basename, $categories;
+		global $available_plugins, $selected_plugin_basename, $categories, $user_settings;
```
felixarntz: Not strictly related to this PR, but do these actually need to be globals? If they are only used in the `admin-page.php` template that is `require_once`d right below in this same method, I think we should be able to use simply the local variables?

If that's the case, we should remove the `global` from this line to not pollute global scope unnecessarily.

felixarntz (changes requested):
Thanks @mukeshpanchal27 for implementing the PR. It looks great so far, sharing a few points of additional feedback to Pascal's below.

assets/js/plugin-check-admin.js
swissspidy: TIL!

mukeshpanchal27: @swissspidy `~` also not supported.

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks for the review and suggestion. I have updated it in https://github.com/WordPress/plugin-check/pull/304/commits/678e5341421ec207b56bd1c6c2cd4db93a3fb157

includes/Admin/Admin_Page.php
mukeshpanchal27: Thanks for raising, Nice catch. Addressed in https://github.com/WordPress/plugin-check/pull/304/commits/678e5341421ec207b56bd1c6c2cd4db93a3fb157

templates/admin-page.php
mukeshpanchal27: Thank you for the review.

For the default checks, I use the default value for the `get_user_setting` function, so we can select all categories by default if the user setting is not configured.

includes/Admin/Admin_Page.php
mukeshpanchal27: Thank you for bringing this up. PHPMD flagged the error, so in the commit https://github.com/WordPress/plugin-check/pull/304/commits/8a7bea21d60e18d0e5c9bbbb9149d1828ca9ba17, I removed the global variable and added `SuppressWarnings` annotations in the function.

assets/js/plugin-check-admin.js
mukeshpanchal27: Thanks for the review. I've opened a separate issue #305, to address this matter.

templates/admin-page.php
```diff
@@ -45,7 +45,7 @@
 							<fieldset>
 								<legend class="screen-reader-text"><?php echo esc_html( $category ); ?></legend>
 								<label for="<?php echo esc_attr( $category ); ?>">
-									<input type="checkbox" id="<?php echo esc_attr( $category ); ?>" name="categories" value="<?php echo esc_attr( $category ); ?>" checked="checked" />
+									<input type="checkbox" id="<?php echo esc_attr( $category ); ?>" name="categories" value="<?php echo esc_attr( $category ); ?>" <?php checked( in_array( 'all_categories', $user_settings, true ) || in_array( $category, $user_settings, true ) ); ?> />
```
felixarntz: More efficient than two `in_array()` checks would be to make `$user_settings` fall back to simply contain all available categories for the case that no value is stored in the user settings.

Then we can simplify this here to:
```suggestion
									<input type="checkbox" id="<?php echo esc_attr( $category ); ?>" name="categories" value="<?php echo esc_attr( $category ); ?>" <?php checked( in_array( $category, $user_settings, true ) ); ?> />
```

includes/Admin/Admin_Page.php
```diff
@@ -150,16 +150,20 @@ private function get_available_plugins() {
...
-
 		$available_plugins = $this->get_available_plugins();
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
 		$categories = Check_Categories::get_categories();
 
+		// Get user settings for category preferences and set a default value to check all categories by default.
+		$user_settings = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
+		$user_settings = explode( '__', $user_settings );
```
felixarntz: See my other comment, we can simplify this to avoid the custom `all_categories` value:
```suggestion
		$user_settings = 'all_categories' === $user_settings ? $categories : explode( '__', $user_settings );
```

On another note: Maybe use a more descriptive variable name? `$user_settings` is what's used under the hood to store the value, but it doesn't describe well what the variable contains. Maybe `$user_enabled_categories` or `$user_checked_categories`?

felixarntz (commented):
Thanks @mukeshpanchal27, this is almost good to go. Two last points of follow up feedback.

mukeshpanchal27:
@felixarntz @swissspidy, I have addressed all the feedback, and the PR is now ready for review.

swissspidy (approved):

felixarntz (approved):

PR #306, opened 2023-10-25
Plugin deploy workflow for `trunk`

### LINKED ISSUE #302: Implement plugin deploy workflow in `trunk`
Once the new `trunk` branch is being used to continue development, we will need to be able to deploy new releases of the plugin.

For this purpose, we should be able to simply copy the workflow from https://github.com/WordPress/plugin-check/blob/legacy-plugin/.github/workflows/deploy.yml into `trunk`.

As part of the same PR, we should also copy over the `.wordpress-org` directory, and we should add a `.gitattributes` file so that only the necessary files are included in the deployed ZIP.

mukeshpanchal27:
Closes #302 

### Checklist:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v3
+
+      - name: WordPress Plugin Deploy
+        id: deploy
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        with:
+          generate-zip: true
+        env:
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
```
mukeshpanchal27, line 24: For added security, I've commented out the secrets. Once the pull request is ready to be merged, I will remove the comments.

felixarntz (approved):

swissspidy (approved):

PR #311, opened 2023-10-31
CLI: add option to exclude directories

### LINKED ISSUE #7: Add a way to ignore directories
Filter `wp_plugin_check_ignore_directories` to exclude directories such as `vendor_prefixed/`

### LINKED ISSUE #309: CLI: add option to ignore folders
Right now specific folders can only be excluded via a filter. A dedicated CLI arg should be provided for that.

swissspidy:
Successfully tested with `wp plugin check wordpress-seo --exclude-directories=vendor_prefixed,admin`

Fixes #309 
Fixes https://github.com/swissspidy/wp-plugin-check-action/issues/7

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -87,10 +87,14 @@ final class Plugin_Check_Command {
 	 * [--include-experimental]
 	 * : Include experimental checks.
 	 *
+	 * [--exclude-directories=<directories>]
+	 * : Additional directories to exclude from checks
+	 * By default, `.git`, `vendor` and `node_modules` directories are excluded.
+	 *
 	 * ## EXAMPLES
 	 *
 	 *   wp plugin check akismet
-	 *   wp plugin check akismet --checks=escaping
+	 *   wp plugin check akismet --checks=late_escaping
 	 *   wp plugin check akismet --format=json
 	 *
 	 * @subcommand check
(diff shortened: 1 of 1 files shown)


includes/CLI/Plugin_Check_Command.php
```diff
@@ -87,10 +87,14 @@ public function __construct( Plugin_Context $plugin_context ) {
...
 	 *
+	 * [--exclude-directories=<directories>]
+	 * : Additional directories to exclude from checks
+	 * By default, `.git`, `vendor` and `node_modules` directories are excluded.
+	 *
 	 * ## EXAMPLES
 	 *
 	 *   wp plugin check akismet
-	 *   wp plugin check akismet --checks=escaping
+	 *   wp plugin check akismet --checks=late_escaping
```
mukeshpanchal27, line 97: Nice catch.

mukeshpanchal27 (approved):
Thanks @swissspidy for the PR. Could you please update doc for [--checks](https://github.com/WordPress/plugin-check/blob/trunk/includes/CLI/Plugin_Check_Command.php#L63) args?

`Only runs checks provided as an argument in comma-separated values, e.g. enqueued-scripts, escaping. Otherwise runs all checks.`

Should be

`Only runs checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping. Otherwise runs all checks.`

swissspidy:
> Could you please update doc for [--checks](https://github.com/WordPress/plugin-check/blob/trunk/includes/CLI/Plugin_Check_Command.php?rgh-link-date=2023-10-31T09%3A55%3A12Z#L63) args?

Done, thanks!

felixarntz (approved):
Thanks @swissspidy, LGTM!

We should also enable that in the UI, as it shouldn't only be possible via CLI. Can you open a follow up issue to log that so that we can implement that separately?

swissspidy:
Done: #313

Issue #308, opened 2023-10-30
CLI: make it easier to exclude specific checks


swissspidy:
Right now it's only possible to pass an allowlist, not a denylist

swissspidy:
Related example: WooCommerce would want to disable the trademark check, because they are allowed to use their own trademark.

Another plugin might want  to disable the `i18n_usage` or `late_escaping` checks because they already use PHPCS for that and would be otherwise torpedoed by a massive amount of duplicate errors.

swissspidy:
Not sure how to best implement this, doesn't seem to be as straightforward as I thought it would be.

@mukeshpanchal27 @felixarntz any thoughts?

felixarntz:
@swissspidy It would certainly be possible, but we would need to be careful to follow the same pattern as other parameters so that it is centrally handled and works with UI and WP-CLI.

I think the foundation would need to be in `Abstract_Check_Runner`. There's a `get_check_slugs()` method with corresponding `get_check_slugs_param()` and `set_check_slugs()`. We would need to implement the same set of methods for the denylist, e.g. `*_check_ignore_slugs*()`.

We could then retrieve this denylist via `get_check_ignore_slugs()` in `Abstract_Check_Runner::get_checks_to_run()`, and then we'd have two options:
* Either pass it as a new parameter in the `Check_Repository::get_checks()` call. This would technically be a breaking change as it breaks the `Check_Repository` interface, but if we think that's the best implementation, it would probably be fine, since I'm sure nobody has written their own `Check_Repository` yet at this point.
* Or we simply use the list _after_ the `Check_Repository::get_checks()` call to remove those checks from the returned map. This would be completely trivial from a BC perspective, but is arguably less reliable as it wouldn't be implemented in the `Check_Repository` which is the foundation for retrieving checks.

felixarntz:
@swissspidy I actually just started looking into a related improvement. If you give me 1 hour or so, I'll open a PR with a related simplification that will improve this.

felixarntz:
@swissspidy The PR I implemented in #315 should make implementing this issue a bit more straightforward while improving the API for better separation of concerns.

My last "either or" consideration in https://github.com/WordPress/plugin-check/issues/308#issuecomment-1787582087 could then become a new `Check_Collection::exclude( array $check_slugs )` method, simple yet centralized. LMKWYT.

swissspidy:
Now that #315 is merged...

[...]

PR #315, opened 2023-10-31
Implement `Check_Collection` to separate concerns for check filtering from `Check_Repository`

felixarntz:
This PR introduces a `Check_Collection` interface, primarily with the purpose of providing enhanced filtering of checks to run, while decoupling that from the `Check_Repository` responsibilities.

The `Check_Repository` was introduced as the central place to register checks, and while a limited flag based selection mechanism makes sense in the scope of that class, the parameter for filtering down to specific checks wasn't really well thought through, which becomes more obvious when considering #308:
* We shouldn't just add more and more parameters to `Check_Repository::get_checks()` to allow further ways to filter.
* At the same time, we should arbitrarily implement filtering partially in `Check_Repository` and partially elsewhere.

The `Check_Collection` approaches relies on sort of a "builder" pattern where `Check_Repository::get_checks()` returns a collection that allows further filtering and largely can be used like an array. A `to_array()` method and a `to_map()` method are present to return the under-the-hood array representation. Internally, the checks are stored in the collection with their slugs to provide the necessary filtering and lookup capabilities.

### DIFF
--- a/includes/Checker/Default_Check_Repository.php
+++ b/includes/Checker/Default_Check_Repository.php
@@ -82,13 +82,10 @@ class Default_Check_Repository implements Check_Repository {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param int   $flags       The check type flag.
-	 * @param array $check_slugs An array of check slugs to return.
-	 * @return array An array of check instances.
-	 *
-	 * @throws Exception Thrown when invalid flag is passed, or Check slug does not exist.
+	 * @param int $flags The check type flag.
+	 * @return Check_Collection Check collection providing an indexed array of check instances.
 	 */
-	public function get_checks( $flags = self::TYPE_ALL, array $check_slugs = array() ) {
+	public function get_checks( $flags = self::TYPE_ALL ) {
 		$checks = array();
 
 		if ( $flags & self::TYPE_STATIC ) {
@@ -99,37 +96,19 @@ class Default_Check_Repository implements Check_Repository {
 			$checks += $this->runtime_checks;
 		}
 
-		// Filter out the specific check slugs requested.
-		if ( ! empty( $check_slugs ) ) {
-			$checks = array_map(
-				function ( $slug ) use ( $checks ) {
-					if ( ! isset( $checks[ $slug ] ) ) {
-						throw new Exception(
-							sprintf(
-								/* translators: %s: The Check slug. */
-								__( 'Check with the slug "%s" does not exist.', 'plugin-check' ),
-								$slug
-							)
-						);
-					}
-
-					return $checks[ $slug ];
-				},
-				$check_slugs
-			);
-		}
-
 		// Return all checks, including experimental if requested.
 		if ( $flags & self::INCLUDE_EXPERIMENTAL ) {
-			return $checks;
+			return new Default_Check_Collection( $checks );
 		}
 
 		// Remove experimental checks before returning.
-		return array_filter(
-			$checks,
-			static function ( $check ) {
-				return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
-			}
+		return new Default_Check_Collection(
+			array_filter(
+				$checks,
+				static function ( $check ) {
+					return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
+				}
+			)
 		);
 	}
 }
--- a/includes/Checker/Default_Check_Repository.php
+++ b/includes/Checker/Default_Check_Repository.php
@@ -102,13 +102,10 @@ class Default_Check_Repository implements Check_Repository {
 		}
 
 		// Remove experimental checks before returning.
-		return new Default_Check_Collection(
-			array_filter(
-				$checks,
-				static function ( $check ) {
-					return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
-				}
-			)
+		return ( new Default_Check_Collection( $checks ) )->filter(
+			static function ( $check ) {
+				return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
+			}
 		);
 	}
 }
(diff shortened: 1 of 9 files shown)


felixarntz:
It's worth noting that this PR also fixes a bug: Before this, `Check_Repository::get_checks()` would return a map (`$slug => $check`) when no `$check_slugs` were passed, but an indexed array of `$check` objects when `$check_slugs` were passed. This was simply due to an oversight. It can be seen especially in the tests updated in this PR: The one test where a slug was passed to `Check_Repository::get_checks()` is the only scenario which now uses `to_array()`, and it is the only place where an indexed array was expected before in the test - which didn't make sense and was probably implemented like that simply to make the test pass.

Now, it returns the collection and then the caller can decide whether they need the map or array format, via `to_map()` / `to_array()`.

includes/Checker/Default_Check_Repository.php
```diff
@@ -99,37 +96,19 @@ public function get_checks( $flags = self::TYPE_ALL, array $check_slugs = array(
 			$checks += $this->runtime_checks;
 		}
 
-		// Filter out the specific check slugs requested.
-		if ( ! empty( $check_slugs ) ) {
-			$checks = array_map(
-				function ( $slug ) use ( $checks ) {
-					if ( ! isset( $checks[ $slug ] ) ) {
-						throw new Exception(
```
mukeshpanchal27, line 107: Before the PR changes( in trunk ), when we ran the check for an unregistered checks, it would display an error in the CLI. However, with the current changes, we no longer receive any errors in this situation. We should determine whether this is the intended behaviour or if there are plans to add error messages for unregistered items later.

Screenshot 2023-11-01 at 10 40 05 AM

tests/phpunit/Checker/Default_Check_Repository_Tests.php
```diff
@@ -116,14 +116,11 @@ public function test_get_checks_returns_checks_via_slug() {
 		$this->repository->register_check( 'static_check', $static_check );
 		$this->repository->register_check( 'runtime_check', $runtime_check );
 
-		$this->assertSame( array( $static_check ), $this->repository->get_checks( Check_Repository::TYPE_ALL, array( 'static_check' ) ) );
-	}
-
-	public function test_get_checks_throws_exception_for_invalid_check_slug() {
```
mukeshpanchal27, line 122: Check above comment.

mukeshpanchal27 (changes requested):
Thank you, @felixarntz, for the PR. It looks quite solid from my perspective. I've left some feedback to seek clarification on certain points.

includes/Checker/Default_Check_Repository.php
```diff
@@ -99,37 +96,19 @@ public function get_checks( $flags = self::TYPE_ALL, array $check_slugs = array(
...
-			static function ( $check ) {
-				return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
-			}
+		return new Default_Check_Collection(
+			array_filter(
+				$checks,
+				static function ( $check ) {
+					return $check->get_stability() !== Check::STABILITY_EXPERIMENTAL;
+				}
+			)
```
swissspidy: Would it make sense to add a `filter()` method on the `Check_Collection`?

Then we could do `return (new Default_Check_Collection( $checks ) )->filter( ... )`

includes/Checker/Default_Check_Repository.php
felixarntz: @mukeshpanchal27 Great point, I made this change consciously though, I just forgot to add a comment about it on the PR. It seemed weird to me to throw an exception in a filtering function, e.g. it could just silently fail.

Now that I think about it again, I am not sure though. Do we want to be strict and fail for such requests? Or do we want to be forgiving? The latter makes it a bit simpler to use, but it could also lead to false expectations, e.g. you want to run only two specific checks one of which e.g. no longer exists or was renamed, but because there's no exception you think the check just passed.

WDYT @mukeshpanchal27 @swissspidy?

tests/phpunit/Checker/Default_Check_Repository_Tests.php
felixarntz: Let's discuss this in https://github.com/WordPress/plugin-check/pull/315#discussion_r1378399550. If we decide to keep the exception, I'll bring back the test.

includes/Checker/Default_Check_Repository.php
felixarntz: Added in https://github.com/WordPress/plugin-check/pull/315/commits/6b9df7fbe80fca772eb13737a29e9a43a750e206

includes/Checker/Default_Check_Repository.php
swissspidy: UX-wise I'd prefer a more strict message and not failing silently, exactly to prevent issues like that with renames etc.

mukeshpanchal27: I'm also in favour of strict messaging. Without clear messages about renaming or deprecation checks, end users may not fully understand whether their plugin has passed all the necessary checks.

felixarntz: @swissspidy @mukeshpanchal27 I have implemented a new `Check_Collection::require()` method for this purpose. This allows us to decouple filtering by specific checks from _requiring_ specific checks to be present.

In our central logic, both functions are now called, i.e. requiring the given checks to be present but then also limiting to only run those. While we could have combined both things into one method, I think the approach of two separate methods allows for more flexibility. For example, there could be a scenario where you really just want to filter on a collection that may _already_ be a subset of checks, so then checking that all slugs are valid may not be applicable.

I've also added proper unit tests for the new class (and as part of that replaced two tests that were previously part of `Default_Check_Repository_Tests`).

mukeshpanchal27 (approved):
Thanks @felixarntz for the changes. LGTM from my end.

swissspidy (approved):

PR #318, opened 2023-11-06
CLI: Add `--exclude-checks` support

### LINKED ISSUE #308: CLI: make it easier to exclude specific checks

swissspidy:
Fixes #308

Successfully tested using something like this:

```
wp plugin check web-stories # Returns some errors
wp plugin check web-stories --exclude-checks=late_escaping,plugin_readme,plugin_review_phpcs # No errors
```

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -62,6 +62,10 @@ final class Plugin_Check_Command {
 	 * [--checks=<checks>]
 	 * : Only runs checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping. Otherwise runs all checks.
 	 *
+	 * [--exclude-checks=<checks>]
+	 * : Exclude checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping.
+	 * Applies after evaluating `--checks`.
+	 *
 	 * [--format=<format>]
 	 * : Format to display the results. Options are table, csv, and json. The default will be a table.
 	 * ---
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -37,6 +37,14 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	protected $check_slugs;
 
 	/**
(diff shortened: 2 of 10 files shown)


includes/CLI/Plugin_Check_Command.php
```diff
@@ -62,6 +62,10 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * [--checks=<checks>]
 	 * : Only runs checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping. Otherwise runs all checks.
 	 *
+	 * [--exclude-checks=<checks>]
```
felixarntz, line 65: You call this `ignore-checks` / `ignore-slugs` everywhere else in this PR. I don't feel strongly about either way, but we should go with a consistent term throughout. It makes sense that the parameter uses "checks" while the code uses "slugs", but I mean we should only use the term "ignore" or "exclude" throughout.

I personally prefer `exclude-checks` I think, but either way WFM.

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -109,6 +117,14 @@ abstract protected function get_plugin_param();
 	 * @return array An array of Check slugs.
 	 */
 	abstract protected function get_check_slugs_param();
+	/**
```
felixarntz: Nit-pick:
```suggestion
	abstract protected function get_check_slugs_param();

	/**
```

felixarntz (commented):
@swissspidy Almost LGTM, only I think we should choose a consistent name for the new argument.

includes/CLI/Plugin_Check_Command.php
swissspidy: Makes sense, let's settle on _exclude_

felixarntz (approved):
Thanks @swissspidy, LGTM!

swissspidy:
@adamsilverstein Just as an FYI, I'll update the GitHub Action right after merging this one.

adamsilverstein (approved):
Fantastic!

mukeshpanchal27 (approved):

PR #319, opened 2023-11-07
Improve behavior for single-file plugins

### LINKED ISSUE #307: Improve behavior for single-file plugins such as `hello.php`
If I run `wp plugin check hello.php` such as `no_plugin_readme` and `trademarked_term`. The latter because it thinks the slug is `plugins` (from `wp-content/plugins/hello.php`).

mukeshpanchal27:
Fixes #307 

This PR improve the behaviour for the single file plugins. You can test single file plugin such as `Hello Dolly`.

CLI command: `npm run wp-env run cli -- wp plugin check hello.php`


mukeshpanchal27:
In an attempt to test the `Plugin_Request_Utility::is_single_file_plugin()` method, I tried to add a unit test in the plugins folder (`/tests/phpunit/testdata/plugins/foo-bar.php`). However, due to the condition at [this line](https://github.com/WordPress/plugin-check/blob/trunk/includes/Plugin_Context.php#L82) in `Plugin_Context`, it appears that this approach will not work for unit tests. 

The issue might be related to the behaviour of `Plugin_Context::location()`. Whether this is a bug that needs to be addressed or an accepted behaviour would require further investigation.

```
public function test_single_file_plugin_without_error() {
	$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'foo-bar.php' );
	$checks        = new Checks();
	$checks_to_run = array(
		new Plugin_Readme_Check(),
		new Trademarks_Check(),
	);

	add_filter(
		'wp_plugin_check_checks',
		function ( $checks ) {
			return array(
				'plugin_readme' => new Plugin_Readme_Check(),
				'trademarks'    => new Trademarks_Check(),
			);
		}
	);

	$results = $checks->run_checks( $check_context, $checks_to_run );

	$this->assertInstanceOf( Check_Result::class, $results );

	$this->assertEmpty( $results->get_warnings() );
	$this->assertEmpty( $results->get_errors() );
}
```

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -189,11 +188,11 @@ private static function file_get_contents( $file ) {
...
 	 * @since n.e.x.t
 	 *
-	 * @param Check_Context $plugin Context for the plugin to check.
+	 * @param Check_Result $result The Check Result to amend.
 	 * @return array List of absolute file paths.
 	 */
-	private static function get_files( Check_Context $plugin ) {
-		$location = $plugin->location();
+	private static function get_files( Check_Result $result ) {
+		$location = $result->plugin()->location();
```
felixarntz: I don't think these changes are needed for the logic you're introducing here? `Plugin_Request_Utility::is_single_file_plugin()` only looks at `$result->plugin()`, so it would be better to keep this as is and only pass the `Check_Context $plugin` to both this method and `Plugin_Request_Utility::is_single_file_plugin()`.

Methods should have as much access as needed, but as little access as possible.

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -49,7 +48,7 @@ abstract class Abstract_File_Check implements Static_Check {
 	 *                   the check).
 	 */
 	final public function run( Check_Result $result ) {
-		$files = self::get_files( $result->plugin() );
+		$files = self::get_files( $result );
```
felixarntz: See my other comment, I think this can be reverted.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -159,4 +160,20 @@ public static function get_directories_to_ignore() {
...
+	 * @param Check_Result $result The Check Result to amend.
+	 * @return bool true if the single file plugin, otherwise false.
+	 */
+	public static function is_single_file_plugin( Check_Result $result ) {
+		if ( $result->plugin()->path() === $result->plugin()->location() ) {
+			return false;
+		}
+
+		return true;
+	}
```
felixarntz: I'm not sure this class is the most suitable place for this method. Why not add this to the `Plugin_Context` class (and thus also make it available in `Check_Context`? This method only concerns the plugin specifically, so it should be part of the relevant class. Then you can simplify the logic to:
```php
return $this->path() !== $this->location();
```

felixarntz (changes requested):
@mukeshpanchal27 Good catch, and I think you're on the right track with the approach. The solution could be simplified though, some of the changes aren't needed.

includes/Checker/Checks/Abstract_File_Check.php
mukeshpanchal27: Got it. Reverted in https://github.com/WordPress/plugin-check/pull/319/commits/6c6c3dfadd99217bec53e3536d255c0087cc86bf

includes/Utilities/Plugin_Request_Utility.php
mukeshpanchal27: Thanks for feedback. Moved function in https://github.com/WordPress/plugin-check/pull/319/commits/7eea1044a27d047d56c1679d8b55af104a1847ff

swissspidy (approved):
Nice and concise!

includes/Plugin_Context.php
```diff
@@ -85,4 +85,15 @@ public function location() {
 
 		return $path;
 	}
+
+	/**
+	 * Checks if single file plugin.
```
felixarntz:
```suggestion
	 * Checks if the plugin is a single file plugin without a dedicated directory.
	 *
	 * This is the case when the single file is directly placed within `wp-content/plugins`.
```

felixarntz (approved):
Thanks @mukeshpanchal27, looks great!

felixarntz:
@swissspidy I believe the `codecov` report failing here was added as part of the Behat tests PR? Can you advise how to best proceed with this? How would we fix this, and is it even relevant here?

swissspidy:
@felixarntz The failed Behat check was just a flake, can be simply restarted. Or do you mean the `codecov/project` status check? That's purely informative. If we want to make it blocking, we can though.

felixarntz:
@swissspidy What's the threshold for `codecov` to fail? Making it blocking is probably a good idea to make sure new tests are added for new code.

On that note @mukeshpanchal27, could you maybe add 2 unit tests for the extra condition in `Plugin_Readme_Check` and `Trademarks_Check`, using single file plugins?

swissspidy:
Not sure what the default threshold is, I suppose any negative change causes a failed check.

Note that right now coverage right now is only collected through Behat tests, not the unit tests. That needs to be added separately. Happy to whip up a quick PR.

felixarntz:
@swissspidy Oh okay. In that case it would be great if you could open an extra PR for this, or feel free to use this one.

swissspidy:
Working on a new one over here: #321

mukeshpanchal27:
> On that note @mukeshpanchal27, could you maybe add 2 unit tests for the extra condition in `Plugin_Readme_Check` and `Trademarks_Check`, using single file plugins?

@felixarntz I have added two tests, but Codecov is indicating an error. Could we address the coverage issue once this pull request is merged?

felixarntz (approved):
Thank you @mukeshpanchal27 for adding the two tests, those looks great.

Regarding the continuing workflow error, I think we'll just have to wait for #321 to be merged, then we can refresh this PR with `trunk` and then it should hopefully pass.

swissspidy:
Why not just merge this now? 🤔

felixarntz:
@swissspidy I think we should avoid failing tests. There's no rush on getting this particular PR shipped, so I don't think we have to merge this before addressing the failures.

swissspidy:
To be fair it's not a failed _test_, but a failed _status check_ because there is more code being introduced and we don't have code coverage set up.

PR #321, opened 2023-11-08
Set up code coverage for unit tests

swissspidy:
So much for "let me just whip up a quick PR" 😅 

Beware, it's quite messy, but it works.

**Updating wp-env**

This was almost working, but the wp-env version was way too old. There was no xdebug on the phpunit container. PHPUnit would complain with `No code coverage driver available`. In newer versions that container got removed and merged into `tests-cli`.

**Composer dependencies, `platform`, autoloader, and more**

The original version at 53fc7f1a1ca2b0fe008c46eac8365161f1d94804 / 355860bf7c6a31129a0897aaeaa8f76a8d08aed3 was rather straightforward and worked like this: keep dependencies as-is, but on CI update PHPUnit if needed.
This was done because the `platform` config in `composer.json` will cause PHPUnit 6.5x to be installed, but on newer PHP versions we need newer PHPunit

After initial code review this was changed to be a bit more complicated:

* Move PHPunit dependency to `/build-cs/composer.json`
* Remove `/build-cs/composer.lock` and the `platform` config in `build-cs/composer.json`, rely on `composer update` for installing the most suitable PHPUnit version
* Automatically run `composer update` in `build-cs` when running tests to install PHPUnit

The problem there:

* `wp-cli/wp-cli-tests` in the main `/composer.json` also requires PHPUnit, so there will be another version of PHPUnit installed, causing conflicts because of the autoloader
* The autoloader from `/composer.json` is required for unit tests to work
* PHPStan in `build-cs/composer.json` requires PHP 7.2, so we would need to ignore platform requirements when installing dependencies...
* However ignoring platform requirements would then install a PHPUnit version that is too new and does not support older PHP version
* `--prefer-lowest` would also not work because then you get like PHPUnit 4.8

Long story short, the solution for all of that is:

* Move `composer.json` with just PHPUnit into its own `/build-phpunit` directory.
* For unit tests, install `/composer.json` dependencies but add the dev autoloader on top, so there's only one version of PHPUnit being loaded ever
* Install PHPunit with `--ignore-platform-reqs`
* Do not install any `build-cs` dependencies so PHPStan won't cause issues.


swissspidy:
OK this is a weird one.

The job with coverage enabled is failing with `Error: The PHPUnit Polyfills library is a requirement for running the WP test suite.`

The question is: why are the other ones even passing??

IMHO this should use phpunit from the plugin and not wp-env...

patches/@wordpress+env+8.11.0.patch
```diff
@@ -0,0 +1,19 @@
...
+ RUN rm /tmp/composer-setup.php`;
+ 
+-	// Install any Composer packages we might need globally.
+-	// Make sure to do this as the user and ensure the binaries are available in the $PATH.
+-	dockerFileContent += `
+-USER $HOST_UID:$HOST_GID
+-ENV PATH="\${PATH}:/home/$HOST_USERNAME/.composer/vendor/bin"
+-RUN composer global require --dev phpunit/phpunit:"^5.7.21 || ^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0"
+-USER root`;
+-
```
swissspidy, line 16: Workaround for https://github.com/WordPress/gutenberg/issues/51419

phpunit.xml.dist
```diff
@@ -6,15 +6,29 @@
 	convertNoticesToExceptions="true"
 	convertWarningsToExceptions="true"
 	>
+	<coverage processUncoveredFiles="false">
+		<include>
+			<directory suffix=".php">includes/</directory>
+			<file>plugin-check.php</file>
+		</include>
+	</coverage>
```
felixarntz, line 14: Why do we need both this and the similarly looking `<filter>` bit below?

.github/workflows/php-test.yml
```diff
@@ -70,7 +72,39 @@ jobs:
...
+            composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^7.5 --with-dependencies
+          elif [[ $PHP_VERSION == "7.2"  ]]; then
+            echo "Installing PHPUnit 8.5x"
+            composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^8.5 --with-dependencies
+          elif [[ $PHP_VERSION != "7.0"  ]]; then
+            echo "Installing latest version of PHPUnit"
+            composer update --ignore-platform-reqs --no-interaction --no-scripts yoast/phpunit-polyfills --with-dependencies
+          fi
+        env:
+          PHP_VERSION: ${{ matrix.php }}
```
felixarntz: Can't we solve this in a simpler way via only `composer.json` (without hard-coding a "PHP version to PHPUnit version" map)?

felixarntz (commented):
Thanks @swissspidy, also for updating `@wordpress/env` and thus doing some much needed cleanup here!

Mostly looks good to me, just two questions.

phpunit.xml.dist
swissspidy: PHPUnit requires such configuration to be able to collect code coverage. It complains otherwise.

.github/workflows/php-test.yml
swissspidy: I wouldn’t know how. Composer always uses a lower version because of the PHP requirements & the lock file. This map is how I‘ve successfully done it in other projects as well. If you know a better way feel free to make edits!

felixarntz: We could consider updating the `pretest-php` and `pretest-php-multisite` scripts in `package.json` to `composer update` instead of `composer install`. We are only using dependencies for workflow related things, so they don't affect the production code tested, so IMO this is fine.

I have used that approach on other WordPress plugins that needed to work in a wider PHP version range, and it works without any issues, and without manual workaround like this.

swissspidy: That doesn't work. `composer update` will still just install PHPUnit 6.5 because that's the last version supporting PHP 7.0. And if you use `--ignore-platform-reqs` then it will always install PHPunit 9, which won't work on older PHP versions.

swissspidy: If I'm missing something and you have precedent code, please feel free to make edits or submit a PR to this one.

felixarntz: > That doesn't work. `composer update` will still just install PHPUnit 6.5 because that's the last version supporting PHP 7.0.

What is the problem with that? Maybe I'm not understanding why this is needed in the first place. If a CI workflow uses PHP 7.0, why would PHPUnit 6.5 not be correct?

swissspidy: Because PHPUnit 6.5 does not run on PHP 8.

When I run composer update on my machine with PHP 8.2, I still get 6.5. I can‘t run any tests. _That‘s_ the problem.

felixarntz: Why are tests passing then without this? 🤔 
In `trunk` there are no errors as far as I can tell, and it already tests with PHP 8.x as well.

Composer generally handles this kind of stuff. If PHPUnit is for some reason stuck at 6.5.x, then there's probably another dependency "holding us back" by depending on this older version.

felixarntz: Oh wait, now it's clear what the problem is, the `platform` requirement in `composer.json`. Sorry, I got completely confused somehow before now, and was looking at the `performance-lab` `composer.json` 🤦 

Indeed, here we should keep the `platform` requirement in place, due to having a couple of _non-dev_ dependencies as well. I wish it was possible to specify `platform` in a way that it only applies to `require`, but not `require-dev`. Then the latter could still use more recent versions as long as they would remain compatible with any `require` dependencies.

I guess the approach of force installing a specific PHPUnit version is an okay workaround, but a better solution may be to use a development / PHPUnit specific `composer.json` without the `platform` limitation, kind of similar to how we already do in the `build-cs` folder for PHP linting.

swissspidy: Well we could move all dev dependencies like phpunit to build-cs/

edit: actually that has platform 8.0 so that won‘t work either

package.json
```diff
@@ -10,19 +10,22 @@
...
     "lint-js": "wp-scripts lint-js",
     "format-js": "npm run lint-js -- --fix",
-    "prelint-php": "wp-env run composer 'install --no-interaction'",
-    "phpstan": "wp-env run composer run-script phpstan",
-    "lint-php": "wp-env run composer run-script lint",
-    "format-php": "wp-env run composer run-script format",
-    "pretest-php": "wp-env run composer 'install --no-interaction'",
-    "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/$(basename $(pwd))/phpunit.xml.dist --verbose'",
-    "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/$(basename $(pwd))/tests/phpunit/multisite.xml --verbose'"
+    "prelint-php": "wp-env run tests-cli --env-cwd=wp-content/plugins/plugin-check/ composer install --no-interaction",
```
mukeshpanchal27:
```suggestion
    "prelint-php": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename $(pwd))/ composer install --no-interaction",
```

This change will again introduce #181 so we should update it. The change applies to all the commands.

swissspidy: This could just be changed to `composer install`, no?

swissspidy: Actually, we can completely remove this line

felixarntz: Why can it be removed?

.github/workflows/php-test.yml
felixarntz: In that case we should remove the `platform` from `build-cs/composer.json`, I don't think there's a good reason to have it there. Forcing a `platform` in `composer.json` is generally very wrong, the only reason we have to do it here is to ensure compatibility of the production dependencies with PHP 7.0. But for `build-cs/composer.json` that is irrelevant, we should rely on the actual PHP version used.

So I think moving all `require-dev` to that file (and removing the `platform` requirement there) is worth a try.

package.json
swissspidy: Because PHPCS is configured in the build-cs directory, no dependencies need to be installed in the root.

Running `composer lint` automatically installs it in build-cs.

.github/workflows/php-test.yml
swissspidy: And then probably also the lock file there

swissspidy: OK that's even more of a mess now...

swissspidy: @felixarntz so moving PHPUnit to `build-cs` is even more of a mess because there we have things like PHPStan which only support 7.2+, so we can't do `composer update` there because of that.

swissspidy: At this point I would really just recommend going with what I had working yesterday and then a better option can be explored separately.

swissspidy: @felixarntz See PR description for the new approach.

It's extremely messy, but it works.

composer.lock
```diff
@@ -4199,54 +4199,6 @@
...
-            ],
-            "support": {
-                "docs": "https://github.com/wp-phpunit/docs",
-                "issues": "https://github.com/wp-phpunit/issues",
-                "source": "https://github.com/wp-phpunit/wp-phpunit"
-            },
-            "time": "2023-08-09T01:26:57+00:00"
-        },
         {
             "name": "yoast/phpunit-polyfills",
```
felixarntz, line 4203: Why is this still here? Shouldn't it no longer be installed in the main `composer.json`?

phpunit.xml.dist
```diff
@@ -6,15 +6,28 @@
...
 	<testsuites>
 		<testsuite name="default">
 			<directory suffix=".php">./tests/phpunit</directory>
 			<exclude>./tests/phpunit/testdata/plugins/*</exclude>
+			<exclude>./tests/phpunit/bootstrap.php</exclude>
```
felixarntz: Would probably be cleaner to move all actual tests to `tests/phpunit/tests` so that we can just put the single directory here without any exclusions, similar to how it is in core. But we could do that separately.

tests/phpunit/multisite.xml
```diff
@@ -6,15 +6,31 @@
...
+		<include>
+			<directory suffix=".php">../../includes/</directory>
+			<file>../../plugin-check.php</file>
+		</include>
+	</coverage>
 	<testsuites>
 		<testsuite name="default">
 			<directory suffix=".php">./</directory>
 			<exclude>./testdata/plugins/*</exclude>
+			<exclude>./bootstrap.php</exclude>
```
felixarntz: See above.

felixarntz (commented):
@swissspidy Thanks, this approach looks excellent to me!

It's messy but it's also the most elegant solution I can think of. The separate `build-phpunit` directory with its own `composer.json` makes total sense to me.

Two minor things below, most importantly shouldn't the `yoast/phpunit-polyfills` package no longer be installed in the main `vendor` directory?

Other than that, there's a failure in https://github.com/WordPress/plugin-check/pull/321/checks?check_run_id=18672695525, can we address that? It seems extremely nit-picky, so we should probably put a threshold on when codecov fails (maybe 5% less coverage? or at least a very basic 2%?)

phpunit.xml.dist
swissspidy: Yes, agreed. The current folder structure is not great.

composer.lock
swissspidy: Because it's a dependency of `wp-cli/wp-cli-tests`, which is still in the main `/composer.json` file for Behat tests.

felixarntz (approved):
Thanks @swissspidy, looks great!

composer.json
```diff
@@ -67,7 +68,7 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "WordPress\\Plugin_Check\\Test_Data\\": "tests/phpunit/testdata",
+      "WordPress\\Plugin_Check\\Test_Data\\": "tests/phpunit/testdata/Checks",
```
joemcgill, line 71: I assume we do not need to autoload the Filesystem classes in this directory since they get loaded through the `With_Mock_Filesystem` trait, otherwise other classes in that folder would not be automatically loaded. I think that's fine for now 👍🏻

joemcgill (approved):
Yikes. This is indeed a complicated set of hoops to jump through. Thanks for the writeup, @swissspidy and for coming up with this workaround. This looks like it should work and I've been able to confirm that running tests locally still work as expected after these changes when using the npm script defined in the `package.json` file.

All of this kind of makes me curious what benefit we're getting out of keeping a platform config setting in our `composer.json` in the first place? From what I can see, we're not actually installing any dependencies that are bundled into the plugin package. Even the PHPCS rulesets could be moved to dev-dependencies. For tools like code coverage or the WP CLI Tests suite, if they're only meant to work in certain workflows and not in local development environments, then it may be worth removing them also and only installing manually in the workflows where they're being used in order to simplify the setup a bit. I think looking at alternate composer setups can be handled apart from unblocking this effort though—just thinking out loud here.

swissspidy:
> All of this kind of makes me curious what benefit we're getting out of keeping a platform config setting in our `composer.json` in the first place? From what I can see, we're not actually installing any dependencies that are bundled into the plugin package.

Yes we are. Plugin Check requires PHPCS as a prod dependency for many checks.

swissspidy:
> For tools like code coverage or the WP CLI Tests suite, if they're only meant to work in certain workflows and not in local development environments

I do like to run tests locally though :-)

joemcgill:
> Yes we are. Plugin Check requires PHPCS as a prod dependency for many checks.

🤦🏻 of course. Afternoon brain got me again. In that case, I wonder if defining a PHP version via the platform could actually cause issues whenever the plugin is run in a more recent version of PHP? Granted, our test matrix should be able to pick up those issues so it's more of a theoretical problem currently, but something to keep in mind.

PR #326, opened 2023-11-28
Plugin check UX improvement for results

### LINKED ISSUE #325: Plugin check UX improvement for results
We can add an expand and collapse button for displaying results. by default, all tables are expanded. There will be a common button that collapses all tables. and after every file name, there is also a button that separately opens and closes every table. 

image

jakaria-istauk:
Fixes #325


assets/js/plugin-check-admin.js
```diff
@@ -108,6 +108,13 @@
 		for ( let i = 0; i < categoriesList.length; i++ ) {
 			categoriesList[ i ].disabled = false;
 		}
+
+		//Adding Open All/Collapse All button
```
mukeshpanchal27:
```suggestion
		// Adding Open All/Collapse All button.
```

Inline comment should be end with period.

🔢 Apply same to other places also.

assets/js/plugin-check-admin.js
```diff
@@ -108,6 +108,13 @@
...
+		if ( resultsContainer.innerHTML !== '' ) {
+			resultsContainer.innerHTML =
+				'<div><button style="margin-top:1.33em;" class="button button-primary clollaps-all" data-state="open">Clollaps All</button></div>' +
```
mukeshpanchal27: All test should be translatable. You can add string in `wp_localize_script`.

templates/results-table.php
```diff
@@ -1,4 +1,7 @@
-<h4><?php esc_html_e( 'FILE:', 'plugin-check' ); ?> {{ data.file }}</h4>
+<h4 id="plugin-check__results-heading-{{data.index}}" class="plugin-check__results-heading" data-index="{{data.index}}">
+	<?php esc_html_e( 'FILE:', 'plugin-check' ); ?> {{ data.file }} 
+	<button class="collapse-btn" data-state="collapse"><?php esc_html_e( 'Collapse' ); ?></button>
```
mukeshpanchal27: Text domain is missing for `Collapse` string.

mukeshpanchal27 (changes requested):
Thanks @jakaria-istauk for the PR.

- Ran `npm run lint-js -- --fix` to fix the JS lint issue and uploaded the changes.
- Left some feedback that needs to address.

jakaria-istauk:
hi @mukeshpanchal27 , 
Thanks for your feedback. I gave an update according to your feedback.

davidperezgar:
I'd suggest to make an arrow like it works the accordions, so it's minimal interface.
Like this one: `<span class="dashicons dashicons-arrow-down-alt2"></span>`

PR #331, opened 2023-12-08
Add Enqueued Resources check

### LINKED ISSUE #25: Create Enqueued_Resources_Check
## Description

Checks that all plugin scripts and styles are enqueued using core functions and not directly output in HTML. This will use the existing WordPress.WP.EnqueuedResources sniff in the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards/).

## Acceptance Criteria

- Class `Enqueued_Resources_Check` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Resources_Check` should extend the `Abstract_PHP_CodeSniffer_Check` abstract class
    - `'sniffs' => 'WordPress.WP.EnqueuedResources'`
    - `'standard' => 'WordPress'`
- Add a sample `<link>` in the `tests/testdata/plugins/test-plugin-with-errors\test-plugin-with-errors.php` file to include any sample script. Make sure it will come under the `WordPress.WP.EnqueuedResources` error.

## Tests Coverage
- `run()` method adds warnings to the `$check_result` when checking an invalid file - check reference [here](https://github.com/10up/plugin-check/blob/trunk/tests/Checker/Checks/I18n_Usage_Check_Tests.php#L20)
- `run()` method does not add any warnings to the `$check_result` when checking a valid file - check reference [here](https://github.com/10up/plugin-check/blob/trunk/tests/Checker/Checks/I18n_Usage_Check_Tests.php#L51)

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/25

**Note:**

### DIFF
--- a/includes/Checker/Checks/Enqueued_Resources_Check.php
+++ b/includes/Checker/Checks/Enqueued_Resources_Check.php
@@ -29,7 +29,7 @@ class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
 	 * @return array The categories for the check.
 	 */
 	public function get_categories() {
-		return array( Check_Categories::CATEGORY_PERFORMANCE );
+		return array( Check_Categories::CATEGORY_GENERAL );
 	}
 
 	/**
--- a/includes/Checker/Checks/Enqueued_Resources_Check.php
+++ b/includes/Checker/Checks/Enqueued_Resources_Check.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Checks\Enqueued_Resources_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+
+/**
+ * Check for running WordPress enqueued resources sniffs.
+ *
+ * @since n.e.x.t
+ */
+class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
+
+	use Stable_Check;
+
+	/**
+	 * Gets the categories for the check.
+	 *
+	 * Every check must have at least one category.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array The categories for the check.
+	 */
+	public function get_categories() {
+		return array( Check_Categories::CATEGORY_PERFORMANCE );
+	}
+
+	/**
+	 * Returns an associative array of arguments to pass to PHPCS.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An associative array of PHPCS CLI arguments.
+	 */
+	protected function get_args() {
+		return array(
+			'extensions' => 'php',
+			'standard'   => 'WordPress',
+			'sniffs'     => 'WordPress.WP.EnqueuedResources',
+		);
+	}
+}
--- a/includes/Checker/Checks/Enqueued_Resources_Check.php
+++ b/includes/Checker/Checks/Enqueued_Resources_Check.php
@@ -29,7 +29,7 @@ class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
 	 * @return array The categories for the check.
 	 */
 	public function get_categories() {
-		return array( Check_Categories::CATEGORY_GENERAL );
+		return array( Check_Categories::CATEGORY_PLUGIN_REPO );
 	}
 
 	/**
--- a/includes/Checker/Checks/Enqueued_Resources_Check.php
+++ b/includes/Checker/Checks/Enqueued_Resources_Check.php
@@ -13,7 +13,7 @@ use WordPress\Plugin_Check\Traits\Stable_Check;
 /**
  * Check for running WordPress enqueued resources sniffs.
  *
- * @since n.e.x.t
+ * @since 1.0.2
  */
 class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
 
@@ -24,7 +24,7 @@ class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
 	 *
 	 * Every check must have at least one category.
 	 *
-	 * @since n.e.x.t
+	 * @since 1.0.2
 	 *
 	 * @return array The categories for the check.
 	 */
@@ -35,7 +35,7 @@ class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
 	/**
 	 * Returns an associative array of arguments to pass to PHPCS.
 	 *
-	 * @since n.e.x.t
+	 * @since 1.0.2
 	 *
 	 * @return array An associative array of PHPCS CLI arguments.
 	 */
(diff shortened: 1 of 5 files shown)


swissspidy (approved):

tests/phpunit/testdata/plugins/test-plugin-enqueued-resources-with-errors/load.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+?>
+<link rel="stylesheet" href="http://someurl/somefile.css">
+<script src="http://someurl/somefile.js"></script>
+
+<?php
+
+?><link rel="stylesheet" href="https://someurl/somefile.css"><?php
+?><script src="https://someurl/somefile.js"></script><?php
```
mukeshpanchal27: Single JS and CSS is enough no need to increase the error.

```suggestion
```

tests/phpunit/tests/Checker/Checks/Enqueued_Resources_Check_Tests.php
```diff
@@ -0,0 +1,40 @@
...
+		$errors = $check_result->get_errors();
+
+		$this->assertNotEmpty( $errors );
+		$this->assertArrayHasKey( 'load.php', $errors );
+		$this->assertEquals( 4, $check_result->get_error_count() );
+	}
+
+	public function test_run_without_errors() {
+		$check         = new Enqueued_Resources_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-enqueued-resources-without-errors/load.php' );
```
mukeshpanchal27, line 30: The file lacks any code to verify the `Enqueued_Resources_Check`. To rectify this, it is advisable to enqueue both style and CSS within the file using the appropriate function. This action ensures that the test successfully passes.

mukeshpanchal27 (dismissed):
Thank you for addressing the checks. I have provided some feedback for your consideration.

Regarding this matter, we are currently in the process of evaluating the acceptance criteria for this check. I don't believe it's fully prepared for engineering work yet. @felixarntz, could you please share your thoughts on this?

includes/Checker/Checks/Enqueued_Resources_Check.php
```diff
@@ -0,0 +1,49 @@
...
+	 * Gets the categories for the check.
+	 *
+	 * Every check must have at least one category.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array The categories for the check.
+	 */
+	public function get_categories() {
+		return array( Check_Categories::CATEGORY_PERFORMANCE );
```
felixarntz: Not sure about this. Whether or not you hard-code assets doesn't really do anything to performance. I think this is rather a general best practice or plugin requirement.

felixarntz (dismissed):
@ernilambar @mukeshpanchal27 I think it might be okay to merge this as it's a simple check using WPCS and it's certainly a best practice. I would however prefer to check with the plugin review team whether they're okay with it.

I think in any case it could be merged as a warning, but right now it is an error and I would like to validate whether that's okay with the review team.

The other thing is that the category of this check should IMO be a different one.

includes/Checker/Checks/Enqueued_Resources_Check.php
ernilambar: Which category is preferred here? I will update the PR when we get confirmation about warning/error from the review team.

mukeshpanchal27:
> I think in any case it could be merged as a warning, but right now it is an error and I would like to validate whether that's okay with the review team.

The WPCS itself return error https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/EnqueuedResourcesSniff.php#L65 so we should return error for this check. Please correct me if i miss anything. Thanks!

includes/Checker/Checks/Enqueued_Resources_Check.php
mukeshpanchal27: I go with `General` category.

ernilambar:
In the Acceptance Criteria in https://github.com/WordPress/plugin-check/issues/25 it is mentioned that warnings should be displayed.

mukeshpanchal27:
It seems the AC are not currently up to date and haven't been approved by any team member. Perhaps this was mistakenly labeled as a notice when, in fact, the sniff itself returns an error, indicating that we should display an error message.

includes/Checker/Checks/Enqueued_Resources_Check.php
ernilambar: Used `General` category.

ernilambar:
PR has been rebased. `GENERAL` category is assigned for this check. Message is kept as `Error`.

davidperezgar:
Hello, 
This check is part of Plugins Team, so the category should be Plugin repo. In our reviews, we actually enforce users to enqueue all assets, so it should be an error.

includes/Checker/Checks/Enqueued_Resources_Check.php
davidperezgar: Hello, It has to be in repo category. It's a requirement for the plugins that we review.

ernilambar: @davidperezgar This has been already updated.

davidperezgar: Ok

ernilambar:
PR has been updated and ready for review.

davidperezgar:
Hello, I have compared Internal Scanner and this check and is not giving same results. 

For example is not detecting this code:
```
// Add the inline footer JS into an external JS file
function inline_footer_js() {
    // Create a nonce
    $nonce = wp_create_nonce('currency_converter_nonce');
    ?>
    <script type="text/javascript">
       var ajaxurl = '<?php echo esc_url(admin_url('admin-ajax.php')); ?>';
       var currencyConverterNonce = '<?php echo esc_js($nonce); ?>';
    </script>
    <?php
}
add_action('wp_footer', 'inline_footer_js');
```

What we are doing in our scanner is using regex. @frantorres could help as well on this. I give you the regex that we are using:
script_src: `/<script\s[^>]*?src\s*=\s*["|\']?([^>"\']*)["|\']?[^>]*?>/i`

script_type: `/<script\s[^>]*?type\s*=\s*["|\']?([^>"\']*(script)[^>"\']*)["|\']?[^>]*?>/i`

script_only: `/<script\s*>/i`

link: `/<link\s[^>]*?rel\s*=\s*["|\']?([^>"\']*(stylesheet)[^>"\']*)["|\']?[^>]*?>/i`

link-type: `/<link\s[^>]*?type\s*=\s*["|\']?([^>"\']*(text\\/css)[^>"\']*)["|\']?[^>]*?>/i`

style: `/<style(\s*|\s[^>]*)>/i`

They were tested and working correctly. I don't know the best approach, but WPCS seems not to be the only one.

davidperezgar:
I've reviewed with Fran, and We could suggest to WPCS to have better detection of enqueues.

swissspidy:
Yes, contributing to WPCS would be best here 👍 

Right now it only has 2 regexes for external scripts/styles, none for inline scripts etc.

Code in question: https://github.com/WordPress/WordPress-Coding-Standards/blob/d77b0204f49e4b4a2bbac791e3bbfbfe603edf06/WordPress/Sniffs/WP/EnqueuedResourcesSniff.php#L51-L81

davidperezgar (approved):
All seems ok. We're going to make an issue to WPCS in order to have better detection of enqueues.

Issue #74, opened 2023-02-08
Enqueued_Scripts_Size_Check includes script dependencies


jjgrainger:
The `Enqueued_Scripts_Size_Check` only accounts for scripts registered by the plugin that include the plugins URL in the script. There may be cases where a script requires a dependency that is sources outside of plugin directory, such as jQuery.

The `Enqueued_Scripts_Size_Check` should be updated to account for those external dependencies as part of the total script size for a page load.

### Acceptance Criteria
- Update the `Enqueued_Scripts_Size_Check` class to account for external dependencies.
- Dependencies will be determined for a script using the `WP_Scripts` class.
- The dependency sizes will be stored separately from the plugin scripts sizes to help with reporting.
- Check error message will have detailed reporting highlighting.
    - The total script size enqueued for the page, including plugin scripts and inline scripts
    - The total size of script dependencies.
    - For example - "The total size of scripts enqueued for the page are X with dependencies adding X"

swissspidy:
This will require some logic similar to `load_script_textdomain` to get the file path from the script URL.

Might be easiest to first converge this class with `Enqueued_Styles_Size_Check` (e.g. by adding an abstract class or trait) so this logic can be reused.

PR #333, opened 2023-12-11
Add Enqueued_Styles_Size_Check

### LINKED ISSUE #19: Create Enqueued_Styles_Size_Check
## Description

This check should work the same as [Enqueued_Scripts_Size_Check](https://github.com/10up/plugin-check/issues/17), adapted for styles instead of scripts.

## Acceptance Criteria

- Class `Enqueued_Styles_Size_Check` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Styles_Size_Check` should extend the `Runtime_Check` abstract class which extends `Check` interface
- Method `run` should be implemented from the `Runtime_Check` abstract class if needed

## Tests Coverage
- `Enqueued_Styles_Size_Check` class is expected to extend `Runtime_Check` class
- `Enqueued_Styles_Size_Check` class is expected to be instance of `Runtime_Check` class

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/19

**Note:**


davidperezgar:
I've changed the tag so the Performance Team can review it.

includes/Checker/Checks/Enqueued_Styles_Size_Check.php
```diff
@@ -0,0 +1,265 @@
...
+		if ( $plugin_style_size > $this->threshold_size ) {
+			foreach ( $plugin_styles as $plugin_style ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						'This style has a size of %1$s which in combination with the other styles enqueued on %2$s exceeds the style size threshold of %3$s.',
+						size_format( $plugin_style['size'] ),
+						$url,
+						size_format( $this->threshold_size )
+					),
```
swissspidy: I just realized `Enqueued_Scripts_Size_Check` has the same flaw, so would be great to make the string translatable there as well.

```suggestion
					sprintf(
						/* translators: 1: style file size. 2: tested URL. 3: threshold file size. */
						__( 'This style has a size of %1$s which in combination with the other styles enqueued on %2$s exceeds the style size threshold of %3$s.', 'plugin-check' )
						size_format( $plugin_style['size'] ),
						$url,
						size_format( $this->threshold_size )
					),
```

swissspidy (approved):
IDK why this wasn't picked up earlier, apologies!

Given that this adapts `Enqueued_Scripts_Size_Check` for styles, it looks pretty straightforward.

Just one suggestion regarding i18n.

swissspidy:
Related:

* #74

Maybe we could add this here as well from the beginning?

includes/Checker/Checks/Enqueued_Styles_Size_Check.php
```diff
@@ -0,0 +1,266 @@
...
+use WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation;
+use WordPress\Plugin_Check\Checker\With_Shared_Preparations;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+use WordPress\Plugin_Check\Traits\URL_Aware;
+
+/**
+ * Check for enqueued style sizes.
+ *
+ * @since n.e.x.t
```
mukeshpanchal27: Add specific version number as we don't have shell script that change `n.e.x.t` to upcoming version when releasing.

May be `1.0.2` is next version

includes/Checker/Checks/Enqueued_Styles_Size_Check.php
```diff
@@ -0,0 +1,266 @@
...
+			$plugin_style_size += $style_size;
+		}
+
+		if ( $plugin_style_size > $this->threshold_size ) {
+			foreach ( $plugin_styles as $plugin_style ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: 1: style file size. 2: tested URL. 3: threshold file size. */
+						__( 'This style has a size of %1$s which in combination with the other styles enqueued on %2$s exceeds the style size threshold of %3$s.', 'plugin-check' )
```
mukeshpanchal27:
```suggestion
						__( 'This style has a size of %1$s which in combination with the other styles enqueued on %2$s exceeds the style size threshold of %3$s.', 'plugin-check' ),
```

It shows error.

mukeshpanchal27 (changes requested):

includes/Checker/Checks/Enqueued_Styles_Size_Check.php
swissspidy: Whoops my bad :)

swissspidy:
In a separate PR we can look into creating an abstract class or so to reduce code repetition between this class and `Enqueued_Scripts_Size_Check`

westonruter (approved):

mukeshpanchal27 (approved):

PR #334, opened 2023-12-11
Prevent rewrite flushes for runtime checks

### LINKED ISSUE #330: Plugin Check plugin will remove code in .htaccess
When I run the Plugin Check plugin on any plugin, the entire part in the `.htaccess` file between `#BEGIN WordPress` and `#END WordPress` is removed.
I mean the code from `<IfModule>` to `</IfModule>`.
I need to open the Settings > Permalinks page, which regenerates the missing rows.

Edit: This happen when the permalink structure is set to "Post name"

swissspidy:
Fixes #330

To test:

1. Activate a plugin
2. Set up pretty permalinks
3. Run performance checks for the activated plugin
4. See that `.htaccess` file won't be empty.

Cannot be tested with unit tests as there is no Apache with mod_rewrite in a testing environment, so it's not possible to verify `.htaccess` modifications in any way.

### DIFF
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -27,11 +27,27 @@ final class Runtime_Environment_Setup {
 		// Get the existing active plugins.
 		$active_plugins = get_option( 'active_plugins' );
 
+		// Get the existing permalink structure.
+		$permalink_structure = get_option( 'permalink_structure' );
+
 		// Set the new prefix.
 		$old_prefix = $wpdb->set_prefix( $table_prefix . 'pc_' );
 
 		// Create and populate the test database tables if they do not exist.
 		if ( $wpdb->posts !== $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $wpdb->posts ) ) ) {
+			/*
+			 * Set the same permalink structure *before* install finishes,
+			 * so that wp_install_maybe_enable_pretty_permalinks() does not flush rewrite rules.
+			 *
+			 * See https://github.com/WordPress/plugin-check/issues/330
+			 */
+			add_action(
+				'populate_options',
+				static function() use( $permalink_structure ) {
+					add_option( 'permalink_structure', $permalink_structure );
+				}
+			);
+
 			wp_install(
 				'Plugin Check',
 				'plugincheck',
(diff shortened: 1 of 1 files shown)


includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -27,11 +27,27 @@ public function set_up() {
...
 		if ( $wpdb->posts !== $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $wpdb->posts ) ) ) {
+			/*
+			 * Set the same permalink structure *before* install finishes,
+			 * so that wp_install_maybe_enable_pretty_permalinks() does not flush rewrite rules.
+			 *
+			 * See https://github.com/WordPress/plugin-check/issues/330
+			 */
+			add_action(
+				'populate_options',
+				static function () use( $permalink_structure ) {
```
westonruter: Micro nit:

```suggestion
				static function () use ( $permalink_structure ) {
```

There is always a space after `use` in the core codebase.

swissspidy: Hmm this kinda stuff should be handled by WPCS :/

felixarntz: Also required by WPCS 3+ I think.

felixarntz (approved):
Thanks @swissspidy, great catch!

includes/Checker/Runtime_Environment_Setup.php
swissspidy: Aren‘t we on WPCS 3 already?

westonruter (approved):

includes/Checker/Runtime_Environment_Setup.php
westonruter: Yes:

https://github.com/WordPress/plugin-check/blob/35eee0bbbceaeb5bbd7281e185d7b65d3f03c7ef/composer.json#L10

westonruter: I've reported this upstream: https://github.com/WordPress/WordPress-Coding-Standards/issues/764#issuecomment-1851018615

westonruter: Already PR opened to fix! https://github.com/WordPress/WordPress-Coding-Standards/pull/2415

mukeshpanchal27:
Thanks @swissspidy. Just miss to approve the PR last night. Great work.

codestylist:
Thanks to everyone who fixed this small but important issue ;-). Good job!

PR #336, opened 2023-12-12
Add Direct_DB_Queries_Check

### LINKED ISSUE #29: Create Direct_DB_Queries_Check
## Description

Checks for direct database queries within the plugin codebase. These should be avoided as these queries can negatively impact performance and bypass inbuilt caches. This sniff will evaluate $wpdb usage and warn the developer about its use. This will use the existing `WordPress.DB.DirectDatabaseQuery` sniff in the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards/).

## Acceptance Criteria

- Class `Direct_DB_Queries_Check` should be created and exists at `includes/Checker/Checks`
- Class `Direct_DB_Queries_Check` should extend the `Abstract_PHP_CodeSniffer_Check` abstract class
    - `'sniffs' => 'WordPress.DB.DirectDatabaseQuery'`
    - `'standard' => 'WordPress'`
- Add a sample code in the `tests/testdata/plugins/test-plugin-with-errors\test-plugin-with-errors.php` file, which will come under the `WordPress.DB.DirectDatabaseQuery` error - we can use DB query by using `$wpdb` global variable.

## Tests Coverage
- `run()` method adds warnings to the `$check_result` when checking an invalid file - check reference [here](https://github.com/10up/plugin-check/blob/trunk/tests/Checker/Checks/I18n_Usage_Check_Tests.php#L20)
- `run()` method does not add any warnings to the `$check_result` when checking a valid file - check reference [here](https://github.com/10up/plugin-check/blob/trunk/tests/Checker/Checks/I18n_Usage_Check_Tests.php#L51)

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/29

**Note:**

### DIFF
--- a/tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
+++ b/tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * Tests for the Direct_DB_Queries_Check class.
+ *
+ * @package plugin-check
+ */
+
+use WordPress\Plugin_Check\Checker\Check_Context;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Direct_DB_Queries_Check;
+
+class Direct_DB_Queries_Check_Tests extends WP_UnitTestCase {
+
+	public function test_run_with_errors() {
+		$check         = new Direct_DB_Queries_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-direct-db-queries-with-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$warnings = $check_result->get_warnings();
+
+		$this->assertNotEmpty( $warnings );
+		$this->assertArrayHasKey( 'load.php', $warnings );
+		$this->assertEquals( 6, $check_result->get_warning_count() );
+	}
+
+	public function test_run_without_errors() {
+		$check         = new Direct_DB_Queries_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-direct-db-queries-without-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$warnings = $check_result->get_warnings();
+
+		$this->assertEmpty( $warnings );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
+	}
+}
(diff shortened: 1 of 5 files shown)


ernilambar:

felixarntz (approved):
@ernilambar I think this looks good. Since that is clearly a WordPress best practice, and since it's a simple check that only triggers warnings, I think it could be merged as is.

tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
```diff
@@ -0,0 +1,40 @@
...
+		$check         = new Direct_DB_Queries_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-direct-db-queries-with-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$warnings = $check_result->get_warnings();
+
+		$this->assertNotEmpty( $warnings );
+		$this->assertArrayHasKey( 'load.php', $warnings );
```
felixarntz, line 25: Maybe worth also asserting here that there are no errors (i.e. that there are only warnings, but no errors)?

tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
```diff
@@ -0,0 +1,40 @@
...
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-direct-db-queries-without-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$warnings = $check_result->get_warnings();
+
+		$this->assertEmpty( $warnings );
+		$this->assertEquals( 0, $check_result->get_warning_count() );
+	}
```
mukeshpanchal27: This test doesn't serve any purpose as the plugin `load.php` is empty. We can consider either removing this test or adding code to the file that will enable it to pass this particular sniff.

tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
```diff
@@ -0,0 +1,40 @@
...
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-direct-db-queries-with-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$warnings = $check_result->get_warnings();
+
+		$this->assertNotEmpty( $warnings );
+		$this->assertArrayHasKey( 'load.php', $warnings );
+		$this->assertEquals( 6, $check_result->get_warning_count() );
```
mukeshpanchal27, line 26: Certainly, it would be beneficial to add an assertion for error messages or specific error codes to ensure comprehensive test coverage.

mukeshpanchal27 (commented):
Thanks @ernilambar for the PR. I've provided some feedback, but aside from those points, everything appears to be in good shape from my perspective.

tests/phpunit/tests/Checker/Checks/Direct_DB_Queries_Check_Tests.php
ernilambar: Passing code is added.

ernilambar:
PR has been updated and ready to review.

mukeshpanchal27 (approved):

PR #339, opened 2023-12-14
Add Categories and Checks CLI commands

### LINKED ISSUE #303: CLI: add command for listing all possible checks and categories
Makes it easier to see which values can be passed to `---checks` and `--categories`

Would also be nice to see which checks are experimental. Maybe even a flag to not list experimental checks.

Could easily support different output formats like csv or table.

ernilambar:
Fixes #303

Command:
`wp plugin list-check-categories`

Output:
Screen Shot 2023-12-14 at 4 01 54 PM

Command:
`wp plugin list-checks`

Output:
Screen Shot 2023-12-14 at 4 01 39 PM

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -125,7 +125,6 @@ final class Plugin_Check_Command {
 				'ignore-warnings'      => false,
 				'ignore-errors'        => false,
 				'include-experimental' => false,
-
 			)
 		);
 
@@ -236,7 +235,7 @@ final class Plugin_Check_Command {
 	}
 
 	/**
-	 * Runs plugin list-checks.
+	 * Lists the available checks for plugins.
 	 *
 	 * ## OPTIONS
 	 *
@@ -260,16 +259,14 @@ final class Plugin_Check_Command {
 	 * @param array $args       List of the positional arguments.
 	 * @param array $assoc_args List of the associative arguments.
 	 *
-	 * @throws WP_CLI\ExitException Show error if not valid runner.
+	 * @throws WP_CLI\ExitException Show error if invalid format argument.
 	 */
 	public function list_checks( $args, $assoc_args ) {
 		$check_repo = new Default_Check_Repository();
 
 		$all_checks = array();
 
-		$checks_items = $check_repo->get_checks()->to_map();
-
-		foreach ( $checks_items as $key => $check ) {
+		foreach ( $check_repo->get_checks() as $key => $check ) {
 			$item = array();
 
 			$item['slug']      = $key;
@@ -297,7 +294,7 @@ final class Plugin_Check_Command {
 	}
 
 	/**
-	 * Runs plugin list-check-categories.
+	 * Lists the available check categories for plugins.
 	 *
 	 * ## OPTIONS
 	 *
@@ -325,6 +322,8 @@ final class Plugin_Check_Command {
 	 *
 	 * @param array $args       List of the positional arguments.
 	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws WP_CLI\ExitException Show error if invalid format argument.
 	 */
 	public function list_check_categories( $args, $assoc_args ) {
 		// Get options based on the CLI arguments.
@@ -354,12 +353,10 @@ final class Plugin_Check_Command {
 	 * @return array List of the check categories.
 	 */
 	private function get_check_categories() {
-		$categories = array();
-
 		$check_categories = new Check_Categories();
 		$categories_slugs = $check_categories->get_categories();
 
-		$categories = array_map(
+		return array_map(
 			function ( $slug ) {
 				return array(
 					'slug' => $slug,
@@ -368,8 +365,6 @@ final class Plugin_Check_Command {
 			},
 			$categories_slugs
 		);
-
-		return $categories;
 	}
 
 	/**
@@ -381,7 +376,7 @@ final class Plugin_Check_Command {
 	 * @param array $defaults   List of the default arguments.
 	 * @return array List of the associative arguments.
 	 *
-	 * @throws WP_CLI\ExitException Show error if plugin not found.
+	 * @throws WP_CLI\ExitException Show error if invalid format argument.
 	 */
 	private function get_options( $assoc_args, $defaults ) {
 		$options = wp_parse_args( $assoc_args, $defaults );
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -19,8 +19,6 @@ use WP_CLI;
 
 /**
  * Plugin check command.
- *
- * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
  */
 final class Plugin_Check_Command {
 
@@ -42,30 +40,6 @@ final class Plugin_Check_Command {
 		'table',
 		'csv',
 		'json',
-	);
-
-	/**
-	 * Default fields.
-	 *
-	 * @since n.e.x.t
-	 * @var array
-	 */
-	protected $default_fields = array(
-		'check'                 => array(
-			'line',
-			'column',
-			'code',
-			'message',
-		),
-		'list-checks'           => array(
-			'slug',
-			'category',
-			'stability',
-		),
-		'list-check-categories' => array(
-			'name',
-			'slug',
-		),
 	);
 
 	/**
@@ -142,20 +116,18 @@ final class Plugin_Check_Command {
 	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
 	 */
 	public function check( $args, $assoc_args ) {
-		/*
-		 * Bail early if the Plugin Checker is not activated.
-		 *
-		 * If the Plugin Checker is not activated, it will throw an error and
-		 * CLI commands won't be usable.
-		 */
-		if ( is_plugin_inactive( $this->plugin_context->basename() ) ) {
-			WP_CLI::error(
-				__( 'Plugin Checker is not active.', 'plugin-check' )
-			);
-		}
-
 		// Get options based on the CLI arguments.
-		$options = $this->get_check_options( $assoc_args );
+		$options = $this->get_options(
+			$assoc_args,
+			array(
+				'checks'               => '',
+				'format'               => 'table',
+				'ignore-warnings'      => false,
+				'ignore-errors'        => false,
+				'include-experimental' => false,
+
+			)
+		);
 
 		// Create the plugin and checks array from CLI arguments.
 		$plugin = isset( $args[0] ) ? $args[0] : '';
@@ -238,8 +210,11 @@ final class Plugin_Check_Command {
 			$warnings = $result->get_warnings();
 		}
 
+		// Default fields.
+		$default_fields = $this->get_check_default_fields( $assoc_args );
+
 		// Get formatter.
-		$formatter = $this->get_formatter( $assoc_args, 'check' );
+		$formatter = $this->get_formatter( $assoc_args, $default_fields );
 
 		// Print the formatted results.
 		// Go over all files with errors first and print them, combined with any warnings in the same file.
@@ -286,9 +261,6 @@ final class Plugin_Check_Command {
 	 * @param array $assoc_args List of the associative arguments.
 	 *
 	 * @throws WP_CLI\ExitException Show error if not valid runner.
-	 *
-	 * @SuppressWarnings(PHPMD.NPathComplexity)
-	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
 	 */
 	public function list_checks( $args, $assoc_args ) {
 		$check_repo = new Default_Check_Repository();
@@ -308,10 +280,17 @@ final class Plugin_Check_Command {
 		}
 
 		// Get options based on the CLI arguments.
-		$options = $this->get_list_checks_options( $assoc_args );
+		$options = $this->get_options( $assoc_args, array( 'format' => 'table' ) );
 
 		// Get formatter.
-		$formatter = $this->get_formatter( $options, 'list-checks' );
+		$formatter = $this->get_formatter(
+			$options,
+			array(
+				'slug',
+				'category',
+				'stability',
+			)
+		);
 
 		// Display results.
 		$formatter->display_items( $all_checks );
@@ -346,19 +325,22 @@ final class Plugin_Check_Command {
 	 *
 	 * @param array $args       List of the positional arguments.
 	 * @param array $assoc_args List of the associative arguments.
-	 *
-	 * @SuppressWarnings(PHPMD.NPathComplexity)
-	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
 	 */
 	public function list_check_categories( $args, $assoc_args ) {
 		// Get options based on the CLI arguments.
-		$options = $this->get_list_check_categories_options( $assoc_args );
+		$options = $this->get_options( $assoc_args, array( 'format' => 'table' ) );
 
 		// Get check categories details.
 		$categories = $this->get_check_categories();
 
 		// Get formatter.
-		$formatter = $this->get_formatter( $options, 'list-check-categories' );
+		$formatter = $this->get_formatter(
+			$options,
+			array(
+				'name',
+				'slug',
+			)
+		);
 
 		// Display results.
 		$formatter->display_items( $categories );
@@ -391,84 +373,17 @@ final class Plugin_Check_Command {
 	}
 
 	/**
-	 * Validates the associative arguments for 'check' command.
+	 * Validates the associative arguments.
 	 *
 	 * @since n.e.x.t
 	 *
 	 * @param array $assoc_args List of the associative arguments.
+	 * @param array $defaults   List of the default arguments.
 	 * @return array List of the associative arguments.
 	 *
 	 * @throws WP_CLI\ExitException Show error if plugin not found.
 	 */
-	private function get_check_options( $assoc_args ) {
-		$defaults = array(
-			'checks'               => '',
-			'format'               => 'table',
-			'ignore-warnings'      => false,
-			'ignore-errors'        => false,
-			'include-experimental' => false,
-		);
-
-		$options = wp_parse_args( $assoc_args, $defaults );
-
-		if ( ! in_array( $options['format'], $this->output_formats, true ) ) {
-			WP_CLI::error(
-				sprintf(
-					// translators: 1. Output formats.
-					__( 'Invalid format argument, valid value will be one of [%1$s]', 'plugin-check' ),
-					implode( ', ', $this->output_formats )
-				)
-			);
-		}
-
-		return $options;
-	}
-
-	/**
-	 * Validates the associative arguments 'list-check-categories' command.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @param array $assoc_args List of the associative arguments.
-	 * @return array List of the associative arguments.
-	 *
-	 * @throws WP_CLI\ExitException Show error if format is invalid.
-	 */
-	private function get_list_check_categories_options( $assoc_args ) {
-		$defaults = array(
-			'format' => 'table',
-		);
-
-		$options = wp_parse_args( $assoc_args, $defaults );
-
-		if ( ! in_array( $options['format'], $this->output_formats, true ) ) {
-			WP_CLI::error(
-				sprintf(
-					// translators: 1. Output formats.
-					__( 'Invalid format argument, valid value will be one of [%1$s]', 'plugin-check' ),
-					implode( ', ', $this->output_formats )
-				)
-			);
-		}
-
-		return $options;
-	}
-
-	/**
-	 * Validates the associative arguments 'list-checks' command.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @param array $assoc_args List of the associative arguments.
-	 * @return array List of the associative arguments.
-	 *
-	 * @throws WP_CLI\ExitException Show error if format is invalid.
-	 */
-	private function get_list_checks_options( $assoc_args ) {
-		$defaults = array(
-			'format' => 'table',
-		);
-
+	private function get_options( $assoc_args, $defaults ) {
 		$options = wp_parse_args( $assoc_args, $defaults );
 
 		if ( ! in_array( $options['format'], $this->output_formats, true ) ) {
@@ -489,35 +404,49 @@ final class Plugin_Check_Command {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array  $assoc_args Associative arguments.
-	 * @param string $type       Command type.
+	 * @param array $assoc_args     Associative arguments.
+	 * @param array $default_fields Default fields.
 	 * @return WP_CLI\Formatter The formatter instance.
 	 */
-	private function get_formatter( $assoc_args, $type ) {
-		$default_fields = $this->default_fields[ $type ];
-
+	private function get_formatter( $assoc_args, $default_fields ) {
 		if ( isset( $assoc_args['fields'] ) ) {
 			$default_fields = wp_parse_args( $assoc_args['fields'], $default_fields );
-		}
-
-		// This applies only to 'wp plugin check' command.
-		if ( 'check' === $type ) {
-			// If both errors and warnings are included, display the type of each result too.
-			if ( empty( $assoc_args['ignore_errors'] ) && empty( $assoc_args['ignore_warnings'] ) ) {
-				$default_fields = array(
-					'line',
-					'column',
-					'type',
-					'code',
-					'message',
-				);
-			}
 		}
 
 		return new WP_CLI\Formatter(
 			$assoc_args,
 			$default_fields
 		);
+	}
+
+	/**
+	 * Returns check default fields.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $assoc_args Associative arguments.
+	 * @return array Default fields.
+	 */
+	private function get_check_default_fields( $assoc_args ) {
+		$default_fields = array(
+			'line',
+			'column',
+			'code',
+			'message',
+		);
+
+		// If both errors and warnings are included, display the type of each result too.
+		if ( empty( $assoc_args['ignore_errors'] ) && empty( $assoc_args['ignore_warnings'] ) ) {
+			$default_fields = array(
+				'line',
+				'column',
+				'type',
+				'code',
+				'message',
+			);
+		}
+
+		return $default_fields;
 	}
 
 	/**
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -8,7 +8,9 @@
 namespace WordPress\Plugin_Check\CLI;
 
 use Exception;
+use WordPress\Plugin_Check\Checker\Check_Categories;
 use WordPress\Plugin_Check\Checker\CLI_Runner;
+use WordPress\Plugin_Check\Checker\Default_Check_Repository;
 use WordPress\Plugin_Check\Checker\Runtime_Check;
 use WordPress\Plugin_Check\Checker\Runtime_Environment_Setup;
 use WordPress\Plugin_Check\Plugin_Context;
@@ -17,6 +19,8 @@ use WP_CLI;
 
 /**
  * Plugin check command.
+ *
+ * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
  */
 final class Plugin_Check_Command {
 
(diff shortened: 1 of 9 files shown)


includes/CLI/Plugin_Check_Checks_Command.php
```diff
@@ -0,0 +1,177 @@
...
+		if ( is_null( $runner ) ) {
+			$runner = new CLI_Checks_Runner();
+		}
+
+		// Make sure we are using the correct runner instance.
+		if ( ! ( $runner instanceof CLI_Checks_Runner ) ) {
+			WP_CLI::error(
+				__( 'CLI Checks Runner was not initialized correctly.', 'plugin-check' )
+			);
+		}
```
swissspidy: Hmm I really don't like that this simple command requires a new `*_Runner` just because the check registration happens in `Abstract_Check_Runner::register_checks()`

Especially since the current `CLI_Runner` class is mostly a hack anyway

I think we need some refactoring to move the registration out of the runner class, maybe to a factory method on `Default_Check_Repository` or so.

Listing all registered checks should be as simple as getting a `Default_Check_Repository` and listing what's in there.

cc @felixarntz for thoughts

includes/CLI/Plugin_Check_Categories_Command.php
```diff
@@ -0,0 +1,191 @@
...
+		 * Bail early if the Plugin Checker is not activated.
+		 *
+		 * If the Plugin Checker is not activated, it will throw an error and
+		 * CLI commands won't be usable.
+		 */
+		if ( is_plugin_inactive( $this->plugin_context->basename() ) ) {
+			WP_CLI::error(
+				__( 'Plugin Checker is not active.', 'plugin-check' )
+			);
+		}
```
swissspidy: This code can be removed as it's dead code.

This class can only ever be loaded when the plugin is active, so it does not make sense to check whether the plugin is active.

includes/CLI/Plugin_Check_Categories_Command.php
```diff
@@ -0,0 +1,191 @@
...
+namespace WordPress\Plugin_Check\CLI;
+
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Plugin_Context;
+use WP_CLI;
+
+/**
+ * Plugin check categories command.
+ */
+final class Plugin_Check_Categories_Command {
```
swissspidy: Would it be possible to add some functional tests for this command in `tests/behat/features/list-categories.feature`?

includes/CLI/Plugin_Check_Checks_Command.php
```diff
@@ -0,0 +1,177 @@
...
+
+use WordPress\Plugin_Check\Checker\CLI_Checks_Runner;
+use WordPress\Plugin_Check\Plugin_Context;
+use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
+use WP_CLI;
+
+/**
+ * Plugin check checks command.
+ */
+final class Plugin_Check_Checks_Command {
```
swissspidy: Would it be possible to add some functional tests for this command in `tests/behat/features/list-checks.feature`?

cli.php
```diff
@@ -35,7 +37,10 @@
 // Create the CLI command instance and add to WP CLI.
 $plugin_command = new Plugin_Check_Command( $context );
 WP_CLI::add_command( 'plugin', $plugin_command );
-
+$plugin_categories_command = new Plugin_Check_Categories_Command( $context );
+WP_CLI::add_command( 'plugin', $plugin_categories_command );
+$plugin_checks_command = new Plugin_Check_Checks_Command( $context );
+WP_CLI::add_command( 'plugin', $plugin_checks_command );
```
swissspidy: When we add these new commands, we definitely need to go over the naming here.

WP-CLI commands typically follow the `wp <noun> <verb>` category, e.g. `wp post list`, `wp plugin activate`, `wp plugin check`, etc.

With new commands, I think it makes sense to establish our own group:

* `wp plugin-check list-checks`
	* Alternatively `wp plugin-check checks list` but I don't envision any other possible commands under this group
* `wp plugin-check list-categories`
	* Alternatively `wp plugin-check categories list` but I don't envision any other possible commands under this group
* `wp plugin-check check`
* `wp plugin check` -> can become an alias for `wp plugin-check check`

felixarntz: +1 for these proposed names. My only suggestion would be to stick to `plugin` over `plugin-check` as a base group name:
* It's more intuitive to use. `wp plugin-check check` or `wp plugin-check list-checks` feels more cumbersome to me.
* All our commands are related to plugins, so they fit well under the `plugin` group.
* While it can be a good practice to group under a plugin specific group, this plugin is a WordPress project, so it makes sense to go for convenience here, as we'll certainly not abuse the usage of this regular CLI group.

includes/CLI/Plugin_Check_Categories_Command.php
```diff
@@ -0,0 +1,191 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $args       List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
+	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+	 */
+	public function check_categories( $args, $assoc_args ) {
```
felixarntz: Let's rename these methods according to @swissspidy's suggested name changes as they should start with a verb, i.e. `list_check_categories()` for example.

includes/CLI/Plugin_Check_Checks_Command.php
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @param array $args       List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws WP_CLI\ExitException Show error if not valid runner.
+	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
+	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+	 */
+	public function check_checks( $args, $assoc_args ) {
```
felixarntz: See above, I'd suggest we should use `list_checks` as the name.

includes/CLI/Plugin_Check_Checks_Command.php
```diff
@@ -0,0 +1,177 @@
...
+		// Get the Runner.
+		$runner = Plugin_Request_Utility::get_runner();
+
+		// Create the runner if not already initialized.
+		if ( is_null( $runner ) ) {
+			$runner = new CLI_Checks_Runner();
```
felixarntz: It's not great that we have to use a runner class here just to get the available checks. I realize there is no other way given the current implementation, but I think that would be relatively straightforward to improve: We should decouple registering the default checks from the runner class.

Since we already have the `Default_Check_Repository` class, which literally carries the term "default" in its name, we may want to incorporate registration of the default checks right into that class? It would make sense to me given the name, and it wouldn't hurt the abstraction, as there's still the `Check_Repository` interface as an agnostic foundation.

I would propose something like this:
* Add a private method `Default_Check_Repository::register_default_checks()`.
* Move all code from `Abstract_Check_Runner::register_checks()`, except the first line, into this new method to register the checks on the same instance.
* Call the new `register_default_checks()` in the constructor of the class.
* Remove `Abstract_Check_Runner::register_checks()`, since it's now only a single line that could just be moved to where the method is called (since it's only used in that one place anyway).

Once we have implemented these changes, the CLI command here could simply use `Default_Check_Repository` and wouldn't need any of the other classes as a workaround.

felixarntz (changes requested):
Thank you for the PR @ernilambar! Looks great so far. I've added a few points of feedback below. Let me know if you have any questions.

cli.php
swissspidy: > My only suggestion would be to stick to `plugin` over `plugin-check` as a base group name:

Then I don't think we can keep `wp plugin check <pluginslug>` working, as it becomes a group instead (but I'd need to verify first)

So are you suggesting a setup like this:

* `wp plugin check list-checks`
* `wp plugin check list-categories`
* `wp plugin check check-plugin`

?

felixarntz: No, what I mean is:
* `wp plugin check`
* `wp plugin list-checks`
* `wp plugin list-check-categories`

swissspidy: Hmm then the subcommands are a bit disconnected, it‘s not clear that the last 2 belong to Plugin Check.

felixarntz: Why would the last 2 be different? They all would be in the `plugin` group and include the word "check" to clarify that they relate to plugin _check_ functionality.

swissspidy: Just because they include the word "check" doesn't make it clear enough that all these three commands belong together IMHO. That's why WP-CLI commands are typically grouped together as I mentioned earlier.

I guess none of the possible options are super pleasing so let's just start with that, and then when we refactor things in #153 we can revisit.

ernilambar:
Code is updated as per the comments. New commands will be:

* `wp plugin list-checks`
* `wp plugin list-check-categories`

Behat tests implementation for new commands is remaining.

@swissspidy Is it possible to run behat tests in local environment? Or is it implemented now in GitHub actions only?

swissspidy:
Use `composer behat` to run the Behat tests.

You may need to run `composer prepare-behat-tests` once first to create the test database

includes/CLI/Plugin_Check_Checks_Command.php
ernilambar: @felixarntz I have implemented as per the comment. But I am confused about the tests for `Default_Check_Repository`. In the tests of this class, it is assumed that no checks are registered by default. But now we have registered checks in the constructor. So all checks are failing. How do you suggest we should proceed?

ernilambar:
PR ready for review.

- CLI commands updated as per the comments.
- Behat tests added
- Updated PHP Unit tests to reflect new changes.

includes/CLI/Plugin_Check_Checks_Command.php
ernilambar: I have updated the tests to reflect new changes.

cli.php
```diff
@@ -35,7 +37,10 @@
 // Create the CLI command instance and add to WP CLI.
 $plugin_command = new Plugin_Check_Command( $context );
 WP_CLI::add_command( 'plugin', $plugin_command );
-
+$plugin_list_check_categories_command = new Plugin_List_Check_Categories_Command( $context );
+WP_CLI::add_command( 'plugin', $plugin_list_check_categories_command );
+$plugin_list_checks_command = new Plugin_List_Checks_Command( $context );
+WP_CLI::add_command( 'plugin', $plugin_list_checks_command );
```
felixarntz: No need for separate classes, it makes the code more complicated than it needs to be. When adding a class as WP-CLI command, all public methods become commands on the same namespace.

Especially since the classes share some commonalities, we should add the new commands logic to the existing `Plugin_Check_Command` class instead. This will avoid some duplicate code.

includes/CLI/Plugin_List_Check_Categories_Command.php
```diff
@@ -0,0 +1,179 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Plugin_Context $plugin_context Plugin context.
+	 */
+	public function __construct( Plugin_Context $plugin_context ) {
+		$this->plugin_context = $plugin_context;
+	}
```
felixarntz: Related to my above comment: All of this for example is already present in `Plugin_Check_Command`, so we wouldn't need to duplicate it if we used the same class for all commands.

Most methods of this class could be moved over to that class as is.

includes/CLI/Plugin_List_Check_Categories_Command.php
```diff
@@ -0,0 +1,179 @@
...
+	 * Validates the associative arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $assoc_args List of the associative arguments.
+	 * @return array List of the associative arguments.
+	 *
+	 * @throws WP_CLI\ExitException Show error if plugin not found.
+	 */
+	private function get_options( $assoc_args ) {
```
felixarntz: This method could still be moved over to `Plugin_Check_Command` by renaming it to be more specific, e.g. `get_list_check_categories_options()`.

We should probably rename `Plugin_Check_Command::get_options()` to `Plugin_Check_Command::get_check_options()` accordingly so that all 3 methods are scoped to the command they are for.

includes/CLI/Plugin_List_Check_Categories_Command.php
```diff
@@ -0,0 +1,179 @@
...
+
+	/**
+	 * Gets the formatter instance to format check results.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $assoc_args Associative arguments.
+	 * @return WP_CLI\Formatter The formatter instance.
+	 */
+	private function get_formatter( $assoc_args ) {
```
felixarntz: By combining the classes, this method could be removed since `Plugin_Check_Command` already has it. The only difference in the three commands is the `$default_fields`, which we can easily solve by making it a new parameter on the private method. That way all 3 command methods can use this one single method and pass their own needed `$default_fields`.

includes/CLI/Plugin_List_Checks_Command.php
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param array $args       List of the positional arguments.
+	 * @param array $assoc_args List of the associative arguments.
+	 *
+	 * @throws WP_CLI\ExitException Show error if not valid runner.
+	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
+	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+	 */
+	public function list_checks( $args, $assoc_args ) {
```
felixarntz: See above, this and other methods could be moved to `Plugin_Check_Command` as needed.

includes/CLI/Plugin_List_Checks_Command.php
```diff
@@ -0,0 +1,163 @@
...
+	 * Validates the associative arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $assoc_args List of the associative arguments.
+	 * @return array List of the associative arguments.
+	 *
+	 * @throws WP_CLI\ExitException Show error if plugin not found.
+	 */
+	private function get_options( $assoc_args ) {
```
felixarntz: This method could still be moved over to `Plugin_Check_Command` by renaming it to be more specific, e.g. `get_list_checks_options()`.

tests/phpunit/tests/Checker/Default_Check_Collection_Tests.php
```diff
@@ -32,22 +32,24 @@ public function set_up() {
...
 
 	public function test_to_array() {
-		$this->assertSame(
-			array_values( $this->checks ),
-			$this->collection->to_array()
-		);
+		$collection_array = $this->collection->to_array();
+
+		$this->assertContains( $this->checks['static_check'], $collection_array );
+		$this->assertContains( $this->checks['runtime_check'], $collection_array );
```
felixarntz: All the changes in these tests are fine, but not really great, as they make the tests more complicated. I realize this is needed to accommodate the change that `Default_Check_Repository` now includes all the default checks, but there's a cleaner way.

I think eventually we should decouple the default logic for registering checks from the actual list of default checks. Maybe we can introduce a `Empty_Check_Repository` class that will take over everything from `Default_Check_Repository`, except the default checks registration. Then `Default_Check_Repository` can be changed to extend that class with the only difference being the check registration.

The tests here could then all be kept as is, simply by using `Empty_Check_Repository` instead of `Default_Check_Repository`. That would still cover all the necessary logic.

tests/phpunit/tests/Checker/Default_Check_Repository_Tests.php
```diff
@@ -28,14 +28,20 @@ public function test_register_static_check() {
 		$check = new Static_Check();
 		$this->repository->register_check( 'static_check', $check );
 
-		$this->assertSame( array( 'static_check' => $check ), $this->repository->get_checks()->to_map() );
+		$checks_map = $this->repository->get_checks()->to_map();
+
+		$this->assertArrayHasKey( 'static_check', $checks_map );
+		$this->assertInstanceOf( Static_Check::class, $checks_map['static_check'] );
```
felixarntz: See my above comment. If we broke `Default_Check_Repository` into separate `Empty_Check_Repository` and `Default_Check_Repository extends Empty_Check_Repository`, we could keep the code in here as is, using `Empty_Check_Repository` instead. That would still cover all logic, in a simpler and safer way.

This test class could then be renamed to `Empty_Check_Repository_Tests`.

includes/CLI/Plugin_List_Checks_Command.php
```diff
@@ -0,0 +1,163 @@
...
+			foreach ( $checks_items as $key => $check ) {
+				$item = array();
+
+				$item['slug']      = $key;
+				$item['stability'] = strtolower( $check->get_stability() );
+				$item['category']  = join( ', ', $check->get_categories() );
+
+				$all_checks[] = $item;
+			}
+		}
```
felixarntz: No need for the if clause, without it the code would work just the same.
```suggestion
		foreach ( $checks_items as $key => $check ) {
			$item = array();

			$item['slug']      = $key;
			$item['stability'] = strtolower( $check->get_stability() );
			$item['category']  = join( ', ', $check->get_categories() );

			$all_checks[] = $item;
		}
```

includes/CLI/Plugin_List_Check_Categories_Command.php
```diff
@@ -0,0 +1,179 @@
...
+				function ( $slug ) {
+					return array(
+						'slug' => $slug,
+						'name' => ucfirst( str_replace( '_', ' ', $slug ) ),
+					);
+				},
+				$categories_slugs
+			);
+
+		}
```
felixarntz: See my other comment, the if clause is unnecessary.
```suggestion
		$categories = array_map(
			function ( $slug ) {
				return array(
					'slug' => $slug,
					'name' => ucfirst( str_replace( '_', ' ', $slug ) ),
				);
			},
			$categories_slugs
		);
```

felixarntz (commented):
@ernilambar The logic itself looks great now. I think there are a couple places where we can drastically simplify the code and avoid duplicate code, most importantly by using the existing `Plugin_Check_Command` class for the 2 new CLI commands.

ernilambar:
PR has been update as per the review comments.

Note: 
- Since PHPMD was generating error, I have added ignore `@SuppressWarnings(PHPMD.ExcessiveClassComplexity)` in `Plugin_Check_Command`.
- `get_list_check_categories_options()` and `get_list_checks_options()` methods are similar now. But there is some room for enhancements for those listing commands. When we implement those, methods would be different.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -266,35 +424,94 @@ private function get_options( $assoc_args ) {
...
 	 */
-	private function get_formatter( $assoc_args ) {
-		$default_fields = array(
-			'line',
-			'column',
-			'code',
-			'message',
-		);
+	private function get_formatter( $assoc_args, $type ) {
+		$default_fields = $this->default_fields[ $type ];
```
felixarntz: Why do we need a new class property for this? This method is only called once in each check, so I think it's simpler and more intuitive and efficient to pass the actual default fields to this method rather than passing the type and then looking up the default fields.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -242,7 +400,7 @@ static function ( $dirs ) use ( $excluded_directories ) {
 	 *
 	 * @throws WP_CLI\ExitException Show error if plugin not found.
 	 */
-	private function get_options( $assoc_args ) {
+	private function get_check_options( $assoc_args ) {
```
felixarntz: Now that I see how similar the three `get_*_options()` methods are (like you already pointed out in https://github.com/WordPress/plugin-check/pull/339#issuecomment-1875228573), I think we should consolidate them to avoid duplicate logic.

Two options I can think of:
* Either we move the `$options['format']` check that is in each of the 3 methods into a separate private method and call it from all places. Benefit: Keeps the code decoupled and allows to make individual changes to each of the three checks.
* Or we remove the 3 methods in favor of only keeping a single `get_options()` method. Then that method would need a new `$defaults` parameter (since that is the only difference between the 3 methods). Benefit: Even less duplicate code.

Let me know what you think.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -211,7 +239,7 @@ static function ( $dirs ) use ( $excluded_directories ) {
 		}
 
 		// Get formatter.
-		$formatter = $this->get_formatter( $assoc_args );
+		$formatter = $this->get_formatter( $assoc_args, 'check' );
```
felixarntz: See my other comment, I think it would be cleaner to simply pass the actual default fields here.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -17,6 +19,8 @@
 
 /**
  * Plugin check command.
+ *
+ * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
```
felixarntz: I don't think we should ignore this warning here. Either we decide we don't care about this, in which case we should ignore it in the PHPMD config file. Or we should address the problem that this class has. I would prefer to try the latter.

Can you try removing this and see whether you can find a solution? Potentially my other feedback about avoiding the duplicate logic between the 3 `get_*_options()` methods will already help with this.

felixarntz (commented):
@ernilambar I think this is almost good to go now. A few follow up feedback points.

includes/Checker/Default_Check_Repository.php
```diff
@@ -7,105 +7,59 @@
...
-	protected $runtime_checks = array();
-
-	/**
-	 * Array map holding all static checks.
-	 *
-	 * @since n.e.x.t
-	 * @var array
-	 */
-	protected $static_checks = array();
+class Default_Check_Repository extends Empty_Check_Repository implements Check_Repository {
```
felixarntz: The `implements` here is redundant as `Empty_Check_Repository` already implements this interface anyway.
```suggestion
class Default_Check_Repository extends Empty_Check_Repository {
```

includes/Checker/Empty_Check_Repository.php
```diff
@@ -0,0 +1,119 @@
...
+	protected $static_checks = array();
+
+	/**
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function __construct() {
+	}
+
```
felixarntz: No need to have an empty constructor.
```suggestion
```

tests/phpunit/tests/Checker/Check_Categories_Tests.php
```diff
@@ -53,7 +53,7 @@ public function test_filter_checks_by_categories( array $categories, array $all_
 		$check_categories = new Check_Categories();
 		$filtered_checks  = $check_categories->filter_checks_by_categories( $checks, $categories );
 
-		$this->assertEquals( $expected_filtered_checks, $filtered_checks->to_map() );
+		$this->assertEquals( array_keys( $expected_filtered_checks ), array_values( array_intersect( array_keys( $filtered_checks->to_map() ), array_keys( $expected_filtered_checks ) ) ) );
```
felixarntz: I'm not sure I understand the change in the assertion, it also looks overly complicated to me. As far as I can tell you didn't change anything in the `Check_Categories` class in this PR. Can you clarify why this change is needed?

includes/CLI/Plugin_Check_Command.php
ernilambar: I have used second approach.

includes/CLI/Plugin_Check_Command.php
ernilambar: Private property removed.

tests/phpunit/tests/Checker/Check_Categories_Tests.php
ernilambar: `Check_Categories_Tests` was using `Default_Check_Repository` class which we modified in this PR. I have now restored original assert statement and used `Empty_Check_Repository` instead of `Default_Check_Repository`.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -114,20 +116,18 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
 	 */
 	public function check( $args, $assoc_args ) {
-		/*
-		 * Bail early if the Plugin Checker is not activated.
-		 *
-		 * If the Plugin Checker is not activated, it will throw an error and
-		 * CLI commands won't be usable.
-		 */
-		if ( is_plugin_inactive( $this->plugin_context->basename() ) ) {
```
ernilambar, line 123: @felixarntz I took liberty to remove this dead code because we were getting `Cyclomatic complexity` issue. Removing this helped reducing that complexity.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -271,21 +404,37 @@ private function get_options( $assoc_args ) {
...
+
+	/**
+	 * Returns check default fields.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $assoc_args Associative arguments.
+	 * @return array Default fields.
+	 */
+	private function get_check_default_fields( $assoc_args ) {
```
ernilambar, line 425: @felixarntz I have created this separate private method to reduce `Cyclomatic complexity` and to make `check()` method leaner.

includes/CLI/Plugin_Check_Command.php
felixarntz: Not sure why this was here in the first place. Probably safe to remove indeed 👍

includes/CLI/Plugin_Check_Command.php
```diff
@@ -114,20 +116,18 @@ public function __construct( Plugin_Context $plugin_context ) {
...
-		$options = $this->get_options( $assoc_args );
+		$options = $this->get_options(
+			$assoc_args,
+			array(
+				'checks'               => '',
+				'format'               => 'table',
+				'ignore-warnings'      => false,
+				'ignore-errors'        => false,
+				'include-experimental' => false,
+
```
```suggestion
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -232,25 +235,155 @@ static function ( $dirs ) use ( $excluded_directories ) {
 		}
 	}
 
+	/**
+	 * Runs plugin list-checks.
```
```suggestion
	 * Lists the available checks for plugins.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -232,25 +235,155 @@ static function ( $dirs ) use ( $excluded_directories ) {
...
+				'stability',
+			)
+		);
+
+		// Display results.
+		$formatter->display_items( $all_checks );
+	}
+
+	/**
+	 * Runs plugin list-check-categories.
```
```suggestion
	 * Lists the available check categories for plugins.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -232,25 +235,155 @@ static function ( $dirs ) use ( $excluded_directories ) {
...
+	 * @throws WP_CLI\ExitException Show error if not valid runner.
+	 */
+	public function list_checks( $args, $assoc_args ) {
+		$check_repo = new Default_Check_Repository();
+
+		$all_checks = array();
+
+		$checks_items = $check_repo->get_checks()->to_map();
+
+		foreach ( $checks_items as $key => $check ) {
```
felixarntz: Could be simplified to:
```suggestion
		foreach ( $check_repo->get_checks() as $key => $check ) {
```
You don't need to call `to_map()`, since the object can be iterated as it implements `IteratorAggregate`.

felixarntz (approved):
This looks great @ernilambar!

Just a few last nit-picks, but no real blockers. Thank you for all your work on this! 🙌

includes/CLI/Plugin_Check_Command.php
```diff
@@ -232,25 +235,155 @@ static function ( $dirs ) use ( $excluded_directories ) {
...
+				return array(
+					'slug' => $slug,
+					'name' => ucfirst( str_replace( '_', ' ', $slug ) ),
+				);
+			},
+			$categories_slugs
+		);
+
+		return $categories;
+	}
```
```suggestion
	private function get_check_categories() {
		$check_categories = new Check_Categories();
		$categories_slugs = $check_categories->get_categories();

		return array_map(
			function ( $slug ) {
				return array(
					'slug' => $slug,
					'name' => ucfirst( str_replace( '_', ' ', $slug ) ),
				);
			},
			$categories_slugs
		);
	}
```

mukeshpanchal27:
In the follow-up PR, it's essential to update the documentation to reflect the changes made by this pull request. Specifically, we removed the `Abstract_Check_Runner::register_checks()` method. cc. @felixarntz 

The affected documentation files that need updating are:

- [README.md line 14](https://github.com/WordPress/plugin-check/blob/trunk/README.md?plain=1#L14)
- [creating-a-static-check.md line 192](https://github.com/WordPress/plugin-check/blob/trunk/docs/creating-a-static-check.md?plain=1#L192)

ernilambar:
PR updated as per the comments. Fixed spacing issues and improved function docs.

swissspidy (approved):

PR #341, opened 2023-12-17
Improve check message

### LINKED ISSUE #200: Message should show at top of screen.
### Describe the bug

When a plugin has a check that fail, there is a message that is shown after all the checks. This should be at the top of screen. 

### Steps to Reproduce

1. Run the plugin checker. 
2. Scroll to the bottom of the screen. See message. 

Screenshot 2023-06-21 at 15 26 09

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

ernilambar:
Tasks:

* Show check message above the result table (Fixes #200)

### DIFF
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -236,6 +236,8 @@
 	 */
 	async function runChecks( data ) {
 		let isSuccessMessage = true;
+		const allResults = [];
+
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			try {
 				const results = await runCheck( data.plugin, data.checks[ i ] );
@@ -247,13 +249,17 @@
 				) {
 					isSuccessMessage = false;
 				}
-				renderResults( results );
+				allResults.push( results );
 			} catch ( e ) {
 				// Ignore for now.
 			}
 		}
 
 		renderResultsMessage( isSuccessMessage );
+
+		for ( let i = 0; i < allResults.length; i++ ) {
+			renderResults( allResults[ i ] );
+		}
 	}
 
 	/**
--- a/assets/js/plugin-check-admin.js
+++ b/assets/js/plugin-check-admin.js
@@ -236,8 +236,6 @@
 	 */
 	async function runChecks( data ) {
 		let isSuccessMessage = true;
-		const allResults = [];
-
 		for ( let i = 0; i < data.checks.length; i++ ) {
 			try {
 				const results = await runCheck( data.plugin, data.checks[ i ] );
@@ -249,17 +247,13 @@
 				) {
 					isSuccessMessage = false;
 				}
-				allResults.push( results );
+				renderResults( results );
 			} catch ( e ) {
 				// Ignore for now.
 			}
 		}
 
 		renderResultsMessage( isSuccessMessage );
-
-		for ( let i = 0; i < allResults.length; i++ ) {
-			renderResults( allResults[ i ] );
-		}
 	}
 
 	/**
@@ -275,13 +269,11 @@
 			? pluginCheck.successMessage
 			: pluginCheck.errorMessage;
 
-		resultsContainer.innerHTML += renderTemplate(
-			'plugin-check-results-complete',
-			{
+		resultsContainer.innerHTML =
+			renderTemplate( 'plugin-check-results-complete', {
 				type: messageType,
 				message: messageText,
-			}
-		);
+			} ) + resultsContainer.innerHTML;
 	}
 
 	/**
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -272,26 +272,22 @@ final class Admin_Page {
 				'type' => 'text/template',
 			)
 		);
-
-		// Print check's page styles.
-		echo '<style>';
-		echo '
-		#plugin-check__results .notice {
-			margin-top: 20px;
-		}
-		#plugin-check__results .notice ~ h4:first-of-type {
-			margin-top: 20px;
-		}
-		#plugin-check__results > h4:first-of-type {
-			margin-top: 80.5px;
-		}
-		@media ( max-width: 782px ) {
-			#plugin-check__results > h4:first-of-type {
-				margin-top: 88.5px;
+		?>
+		<style>
+			#plugin-check__results .notice,
+			#plugin-check__results .notice + h4 {
+				margin-top: 20px;
 			}
-		}
-		';
-		echo '</style>';
+			#plugin-check__results h4:first-child {
+				margin-top: 80.5px;
+			}
+			@media ( max-width: 782px ) {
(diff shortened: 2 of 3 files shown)


ernilambar:
**Before:**
before-pr

**After:**
after-pr

swissspidy (approved):

assets/js/plugin-check-admin.js
```diff
@@ -247,13 +249,17 @@
...
 			} catch ( e ) {
 				// Ignore for now.
 			}
 		}
 
 		renderResultsMessage( isSuccessMessage );
+
+		for ( let i = 0; i < allResults.length; i++ ) {
+			renderResults( allResults[ i ] );
+		}
```
mukeshpanchal27: This currently loops through similar data, requiring users to wait for all checks to complete before receiving results. For instance, in a large plugin with numerous Errors/Notices, users only see the results after all checks finish, displaying errors or successes. In my view, this negatively impacts user experience as users suddenly encounter a barrage of errors after a brief wait. 

To observe this issue, you can compare the user experience of the WooCommerce plugin before and after the PR changes.

mukeshpanchal27 (changes requested):
Thank you for the PR. I have a concern regarding the approach taken in this PR. It might be beneficial for us to explore alternative options or discuss the approach within the issue before proceeding further.

assets/js/plugin-check-admin.js
ernilambar: Yah, I noticed that in the bigger plugins. I have reverted that loop and kept as it is. To fix this, I have prepended the final output message instead of appending. If this approach is also acceptable then we can re-explore another approaches. :-)

templates/admin-page.php
```diff
@@ -63,6 +63,8 @@
 		<?php } ?>
 	</div>
 
+	<br />
```
felixarntz: While we don't use any custom CSS yet on this screen, that doesn't mean we shouldn't. I don't think a `<br>` tag is correct here, since it carries semantic meaning and is typically an inline element.

I think we should rather put an inline `style` attribute on one of the two `div`s, either with `margin-bottom` on the top one or with `margin-top` on the bottom one.

Alternatively, we could print a `<style>` tag in the head of the page, but that may be overkill for just this one style rule.

ernilambar: I have removed br tag and create separate PR to add inline styling. https://github.com/WordPress/plugin-check/pull/368 Since this PR needs more discussion, we could merge that PR separately.

mukeshpanchal27 (changes requested):
Thank you for updating the approach.

The primary drawback is that the entire output shifts downward when an error message is appended.

During the checks, if someone starts reviewing the notices, once all checks are completed, the position of that initial error moves to the bottom due to new errors appearing at the top. Consequently, since there are no error messages at the bottom, if someone is inspecting the WC (WooCommerce) plugin and reaches the bottom, they won't find any notice. To address this, we should consider adding error messages at the bottom as well, allowing users to view them without having to scroll back to the top. What do you think?

mukeshpanchal27:
Recoding for small plugin checks: 

https://github.com/WordPress/plugin-check/assets/10103365/5dd24770-871e-4794-b832-81f658ccb607

ernilambar:
> To address this, we should consider adding error messages at the bottom as well, allowing users to view them without having to scroll back to the top. What do you think?

But also this wont fix the issue of shifting of output.

templates/admin-page.php
felixarntz: Not sure a separate PR is needed here. Using a `<br>` isn't quite right here semantically, and since there's now the more severe issue https://github.com/WordPress/plugin-check/pull/341#pullrequestreview-1800002667 outlined by @mukeshpanchal27, which requires CSS to solve anyway, I think we should address this here.

felixarntz:
@ernilambar @mukeshpanchal27 
> > To address this, we should consider adding error messages at the bottom as well, allowing users to view them without having to scroll back to the top. What do you think?
> 
> But also this wont fix the issue of shifting of output.

Both of this can be addressed. We can keep the notice only at the top, but we need to "reserve" the appropriate space for the notice. We can print a `<style>` tag on the page to make that happen (and also address https://github.com/WordPress/plugin-check/pull/341#discussion_r1434267845). Something like:
```html
<style>
.plugin-check__results .notice:first-child {
	margin-top: # Whatever margin looks good
}

.plugin-check__results h4:first-child {
	margin-top: # Margin that is the same as the above margin plus the actual height of the notice
}
</style>
```

By doing that, the contents shouldn't shift any longer from the notice being added.

ernilambar:
I have updated the PR with CSS. Currently I have added CSS to existing `admin_footer()` method because when I added new method to hook to `admin_head`, GitHub Action generate error like `Too many public methods.`

includes/Admin/Admin_Page.php
```diff
@@ -272,6 +272,26 @@ public function admin_footer() {
 				'type' => 'text/template',
 			)
 		);
+
+		// Print check's page styles.
+		echo '<style>';
+		echo '
```
felixarntz: Nit-pick, but better to directly output rather than using `echo`. For instance, it gives you syntax highlighting in most IDEs.
```suggestion
		// Print admin page styles.
		?>
		<style>
```

includes/Admin/Admin_Page.php
```diff
@@ -272,6 +272,26 @@ public function admin_footer() {
...
+		#plugin-check__results > h4:first-of-type {
+			margin-top: 80.5px;
+		}
+		@media ( max-width: 782px ) {
+			#plugin-check__results > h4:first-of-type {
+				margin-top: 88.5px;
+			}
+		}
+		';
+		echo '</style>';
```
felixarntz: See above:
```suggestion
		</style>
		<?php
```

includes/Admin/Admin_Page.php
```diff
@@ -272,6 +272,26 @@ public function admin_footer() {
...
+			margin-top: 20px;
+		}
+		#plugin-check__results > h4:first-of-type {
+			margin-top: 80.5px;
+		}
+		@media ( max-width: 782px ) {
+			#plugin-check__results > h4:first-of-type {
+				margin-top: 88.5px;
+			}
+		}
```
felixarntz: A few suggestions to simplify this CSS (there's no need to use `first-of-type` or the `~` operator):
```suggestion
		#plugin-check__results .notice,
		#plugin-check__results .notice + h4 {
			margin-top: 20px;
		}
		#plugin-check__results h4:first-child {
			margin-top: 80.5px;
		}
		@media ( max-width: 782px ) {
			#plugin-check__results h4:first-child {
				margin-top: 88.5px;
			}
		}
```

felixarntz (approved):
@ernilambar LGTM, just two small things below which IMO would be good to improve.

mukeshpanchal27 (approved):
LGTM!

PR #344, opened 2023-12-18
Remove default select option from dropdown

### LINKED ISSUE #343: Remove default select option from plugins dropdown
Proposing to remove default select option:

`<option value=""><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>`

Ref: `/templates/admin-page.php` L26

* With this number of clicks will be reduced when there is only one option in the dropdown.
* Since this is a required field, making one selected would be better UI I believe.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/343

### DIFF
--- a/templates/admin-page.php
+++ b/templates/admin-page.php
@@ -23,7 +23,9 @@
 				</h2>
 
 				<select id="plugin-check__plugins-dropdown" name="plugin_check_plugins">
-					<option value=""><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
+					<?php if ( 1 !== count( $available_plugins ) ) { ?>
+						<option value=""><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
+					<?php } ?>
 					<?php foreach ( $available_plugins as $plugin_basename => $available_plugin ) { ?>
 						<option value="<?php echo esc_attr( $plugin_basename ); ?>"<?php selected( $selected_plugin_basename, $plugin_basename ); ?>>
 							<?php echo esc_html( $available_plugin['Name'] ); ?>


mukeshpanchal27:
Thanks @ernilambar for the PR. I like the idea to remove default empty selection. If we there is no plugin then it will show `No plugins available.` message.

I value the input from @felixarntz and @joemcgill on this matter.

felixarntz:
Hmm, I'm not sure about this. The only benefit I could see here is if a site only has a single plugin and that's what you want to test. But otherwise I would find it an arbitrary choice to have the alphabetically first plugin pre-selected, so from that perspective I prefer the neutral default we have now.

How about this?
* If there's one plugin, we only put that plugin there and have it pre-selected.
* Otherwise, things remain as they are now.

I think that would be a good compromise that would address both concerns.

I don't feel strongly about it though, curious to hear other perspectives.

mukeshpanchal27:
> How about this?
> 
> * If there's one plugin, we only put that plugin there and have it pre-selected.
> * Otherwise, things remain as they are now.

I personally like these suggestion.

ernilambar:
PR updated as per the comment.

templates/admin-page.php
```diff
@@ -23,7 +23,13 @@
 				</h2>
 
 				<select id="plugin-check__plugins-dropdown" name="plugin_check_plugins">
-					<option value=""><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
+					<?php
+					if ( 1 !== count( $available_plugins ) ) {
+					?>
```
mukeshpanchal27:
```suggestion
					<?php if ( 1 !== count( $available_plugins ) ) { ?>
```

templates/admin-page.php
```diff
@@ -23,7 +23,13 @@
...
+						<option value=""><?php esc_html_e( 'Select Plugin', 'plugin-check' ); ?></option>
+					<?php
+					}
+					?>
```
mukeshpanchal27:
```suggestion
					<?php } ?>
```

mukeshpanchal27 (commented):
Thanks for the update @ernilambar. Left nit-pick other then look good to me.

swissspidy (approved):

felixarntz (approved):
Thanks @ernilambar!

PR #345, opened 2023-12-18
Add documentation for CLI commands

### LINKED ISSUE #38: Create documentation for CLI command
Create documentation for the `wp plugin check` command. This documentation will live in this repo.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/38


docs/wp-plugin-check.md
```diff
@@ -0,0 +1,34 @@
+[Back to overview](./README.md)
+
+# wp plugin check
```
swissspidy: Can we try using `wp scaffold package-readme` for this so the documentation is automated?

See https://github.com/wp-cli/scaffold-package-command#wp-scaffold-package-readme

ernilambar: Sure. I will do research.

ernilambar: `wp scaffold package-readme` requires valid package directory as parameter with valid `composer.json`. So as of now we would not be able to use that command in our repo. When we separate our CLI and make it as a package then only we would be able to use this.

swissspidy: What updates to composer.json do we need to make?

ernilambar: I dug deeper into the scaffold command but faced multiple issues. To generate doc content it used `wp cli cli-dump` commands output but there all commands are there but our `plugin check` is not there. Also location of `README.md` and also file name is not configurable. I am testing in this PR https://github.com/WordPress/plugin-check/pull/356

swissspidy: Ah, I see, right, it looks at all the registered commands, but the ones in the plugin are of course only registered when WP is active. Too bad!

Then this should be OK for now.

ernilambar:
I developed simple helper CLI command to generate doc for the CLI commands. https://github.com/ernilambar/pc-cli-docs So doc generation is now semi-automatic instead of manual.
It would be great if we could find a way incorporate this command in the trunk and exclude it when we deploy to SVN in the plugin directory.

Preview: https://github.com/ernilambar/plugin-check/blob/38-add-cli-documentation/docs/CLI.md

There is no `EXAMPLES` section for `wp plugin list-checks` because it is missing in the command itself. I have created separate PR to add example. https://github.com/WordPress/plugin-check/pull/396

swissspidy:
It would be better to contribute to https://github.com/wp-cli/scaffold-package-command to make it easier to generate documentation for commands within a plugin. Kudos for building something custom. However, long term it is not viable to fork and maintain a custom tool that just works with Plugin Check. Especially since these commands don't change frequently.

swissspidy (approved):

docs/CLI.md
```diff
@@ -0,0 +1,105 @@
...
+By default, `.git`, `vendor` and `node_modules` directories are excluded.
+```
+## EXAMPLES
+```
+wp plugin check akismet
+wp plugin check akismet --checks=late_escaping
+wp plugin check akismet --format=json
+```
+
+# wp plugin list-checks
```
mukeshpanchal27, line 55: Add `EXAMPLES` that added in #396

docs/CLI.md
```diff
@@ -0,0 +1,105 @@
...
+  - csv
+  - json
+---
+```
+## EXAMPLES
+```
+wp plugin list-check-categories
+wp plugin list-check-categories --format=json
+```
+
```
mukeshpanchal27:
```suggestion
```

mukeshpanchal27 (commented):
Thanks for the PR! I've left some minor feedback for your consideration.

docs/CLI.md
ernilambar: Done

mukeshpanchal27 (approved):

PR #352, opened 2023-12-20
Remove HTML tag from messages in Trademarks Check

### LINKED ISSUE #312: HTML tags in error messages unexpectedly escaped
Screenshot 2023-10-30 at 21 35 35

The `<code>` tags shouldn't be escaped here.

ernilambar:
Closes https://github.com/WordPress/plugin-check/issues/312

* Remove `<code>` HTML tag from the message string and used `"` instead.
* Affected checks - `Trademarks_Check`


felixarntz (approved):
Thanks @ernilambar! Regardless of whether we'll come up with something more elaborate in #312, this change makes sense to me.

This is about a plugin name, which makes more sense to be wrapped in quotes rather than `code` (it's not code).

swissspidy (approved):

PR #353, opened 2023-12-20
Require Node v20

swissspidy:
See https://make.wordpress.org/core/2023/12/20/updating-wordpress-to-use-more-modern-versions-of-node-js-npm-2/


ernilambar:
I checked this PR in my local setup. `npm install` and our other node scripts are also working as expected.

mukeshpanchal27 (commented):
Thanks @swissspidy for the PR.

Got some error when run `npm run phpstan`. Are you facing same?

```
-- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
     Error                                                                                                                                                                                                 
 -- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
     Ignored error pattern /^Function str_contains not found.$/ in paths: /var/www/html/wp-content/plugins/plugin-check/includes/Checker/Checks/Abstract_File_Check.php,                                   
     /var/www/html/wp-content/plugins/plugin-check/includes/Traits/Find_Readme.php, /var/www/html/wp-content/plugins/plugin-check/includes/Traits/File_Editor_URL.php was not matched in reported errors.  
 -- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
```

ernilambar:
@mukeshpanchal27 This error seems to appear in `trunk` also. But `composer run-script phpstan` is working as expected.

swissspidy:
PHPStan has nothing to do with Node.

This particular error is ignored in PHPStan FWIW, https://github.com/WordPress/plugin-check/blob/72f3eb321cf1cd05c234367443da1ccdfd90bcd5/phpstan.neon.dist#L17-L23

So you should not encounter it. I don't get any errors on `trunk` at least.

But again, not related to do this PR.

.nvmrc
```diff
@@ -1 +1 @@
-16
+20
```
mukeshpanchal27:
```suggestion
20

```

Not blocker: Missing end line

mukeshpanchal27 (approved):

felixarntz (approved):

PR #354, opened 2023-12-21
Fix plugin name from readme in `Trademarks` check

ernilambar:
When checking plugin name from readme, we are using this `/===(.*)===/i` regex. So its matches will also include whitespace around the actual plugin name. This PR will omit that whitespace characters.

* Updated `Trademarks_Check`
* Updated `Trademarks_Check_Tests` to reflect this change

Before

before-this-pr

After

after-this-pr


swissspidy (approved):

mukeshpanchal27 (approved):
Nice catch.

PR #357, opened 2023-12-26
Add admin page link in plugin actions

ernilambar:
This PR adds admin page link in plugin action links. This will be helpful accessing the page in the admin panel.

Screen Shot 2023-12-26 at 10 53 19 AM


ernilambar:

felixarntz:
I personally don't see the value in this change. All other plugins have a more helpful link to check for a specific plugin, and the screen can be reached in the Tools menu. What's the benefit of a general link to the screen here? I find it more confusing than helpful (next to the other plugins' check action links).

davidperezgar:
I do think that is helpful. Maybe the problem is in the name. Let's make a change to "Check Plugin"

PR #358, opened 2023-12-26
Increase Behat test coverage

ernilambar:


ernilambar:

tests/behat/features/plugin-check.feature
```diff
@@ -29,3 +29,73 @@ Feature: Test that the WP-CLI command works.
...
+      """
+
+  Scenario: Check Akismet
+    Given a WP install with the Plugin Check plugin
+
+    When I try the WP-CLI command `plugin check akismet`
+    Then STDOUT should contain:
+      """
+      FILE: views/config.php
+      """
```
swissspidy: Worth noting that this is a bit error prone as it could break whenever Akismet releases a new version. But should be ok for now I guess

tests/behat/features/plugin-check.feature
```diff
@@ -29,3 +29,73 @@ Feature: Test that the WP-CLI command works.
...
+
+    When I try the WP-CLI command `plugin check akismet --exclude-directories=views`
```
swissspidy: If you don't expect the command to fail, use `When I run` instead of `When I try`. Same goes for above.

ernilambar: `try` replaced with `run`.

swissspidy (approved):

tests/behat/features/plugin-check.feature
```diff
@@ -29,3 +29,73 @@ Feature: Test that the WP-CLI command works.
       """
       All output should be run through an escaping function
       """
+
+    When I run `wp plugin check hello.php --format=csv`
```
mukeshpanchal27:
```suggestion
    When I run the WP-CLI command `wp plugin check hello.php --format=csv`
```

Should mention `the WP-CLI command` as we did for other cases. What do you think @swissspidy?

🔢 Apply to other places

mukeshpanchal27 (commented):
Thanks for the PR. Left one minor suggestion.

tests/behat/features/plugin-check.feature
swissspidy: Yes indeed, good point! That's required for actually being able to collect the code coverage.

ernilambar: PR updated.

ernilambar: Earlier `wp` is omitted when `When I run WP-CLI command` was used like given below. I have also followed similar. Hope that is ok. `When I try the WP-CLI command 'plugin check foo-bar'`

mukeshpanchal27 (approved):

PR #366, opened 2023-12-29
Update Code Obfuscation Check to display multiple errors

### LINKED ISSUE #363: Code Obfuscation Check does not show multiple obfuscated files
Issue similar to https://github.com/WordPress/plugin-check/issues/362

When there are multiple files with obfuscated code, only one error is shown.

I tested with following code in multiple files.

    <?php @Zend;
    3272;
    /* Ò!This is not a text file!ﬁÌ  */

When we create new method like `files_preg_match()` or similar in `Abstract_File_Check` to find multiple matched instances, we could use that here also.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/363

* Added new static methods in `Abstract_File_Check`
  - `files_preg_match()` - Returns multiple files unlike `file_preg_match()`
  - Similarly `files_str_contains`
* Updated methods in `Code_Obfuscation_Check` to handle array of files instead of a single file
* `files_preg_match()` returns array of file and match. This data could be useful in the future enhancements.
Eg:
Screen Shot 2023-12-29 at 1 41 55 PM

CLI output:

Screen Shot 2023-12-29 at 1 32 26 PM


swissspidy (approved):

ernilambar:
May be this PR should be rewritten following the approach used in [Detect localhost/127.0.0.1 code in multiple files ](https://github.com/WordPress/plugin-check/pull/399)

ernilambar:
Closing in favor of https://github.com/WordPress/plugin-check/pull/532

PR #367, opened 2023-12-29
Change logic for finding plugin slug in CLI

### LINKED ISSUE #364: Improve parsing of associative arguments to accept arguments regardless of position
Command 1:

    wp plugin check hello.php --fields=line,column,type,code

Command 2:

    wp plugin check --fields=line,column,type,code hello.php

These both commands are expected to give same output. But it seems our CLI breaks when assoc args are passed before the `plugin` slug parameter.

PHP fatal error:

    Fatal error: Uncaught Exception: Invalid plugin slug: Plugin with slug --fields=line,column,type,code is not installed. in /var/www/html/wp-content/plugins/plugin-check/includes/Utilities/Plugin_Request_Utility.php:77

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/364

### DIFF
--- a/includes/Checker/CLI_Runner.php
+++ b/includes/Checker/CLI_Runner.php
@@ -57,8 +57,19 @@ class CLI_Runner extends Abstract_Check_Runner {
 	 * @throws Exception Thrown if the plugin parameter is empty.
 	 */
 	protected function get_plugin_param() {
-		// Get the plugin parameter from the command line arguments.
-		$plugin = isset( $_SERVER['argv'][3] ) ? $_SERVER['argv'][3] : '';
+		// Exclude first three reserved elements.
+		$params = array_slice( $_SERVER['argv'], 3 );
+
+		// Remove associative arguments.
+		$params = array_filter(
+			$params,
+			function ( $val ) {
+				return ! str_starts_with( $val, '--' );
+			}
+		);
+
+		// Use only first element. We dont support checking multiple plugins at once yet!
+		$plugin = count( $params ) > 0 ? reset( $params ) : '';
 
 		if ( empty( $plugin ) ) {
 			throw new Exception(
--- a/includes/Checker/CLI_Runner.php
+++ b/includes/Checker/CLI_Runner.php
@@ -63,7 +63,7 @@ class CLI_Runner extends Abstract_Check_Runner {
 		// Remove associative arguments.
 		$params = array_filter(
 			$params,
-			function ( $val ) {
+			static function ( $val ) {
 				return ! str_starts_with( $val, '--' );
 			}
 		);
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -24,5 +24,4 @@ parameters:
         - includes/Traits/File_Editor_URL.php
     -
       message: '/^Function str_starts_with not found.$/'
-      paths:
-        - includes/Checker/CLI_Runner.php
+      path: includes/Checker/CLI_Runner.php
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -22,3 +22,7 @@ parameters:
         - includes/Checker/Checks/Plugin_Readme_Check.php
         - includes/Traits/Find_Readme.php
         - includes/Traits/File_Editor_URL.php
+    -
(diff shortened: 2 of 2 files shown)


swissspidy (approved):

phpstan.neon.dist
```diff
@@ -21,3 +21,7 @@ parameters:
         - includes/Checker/Checks/Abstract_File_Check.php
         - includes/Traits/Find_Readme.php
         - includes/Traits/File_Editor_URL.php
+    -
+      message: '/^Function str_starts_with not found.$/'
+      paths:
+        - includes/Checker/CLI_Runner.php
```
mukeshpanchal27:
```suggestion
      path: includes/Checker/CLI_Runner.php
```

For single file let's use `path`

includes/Checker/CLI_Runner.php
```diff
@@ -57,8 +57,19 @@ public static function is_plugin_check() {
...
 	protected function get_plugin_param() {
-		// Get the plugin parameter from the command line arguments.
-		$plugin = isset( $_SERVER['argv'][3] ) ? $_SERVER['argv'][3] : '';
+		// Exclude first three reserved elements.
+		$params = array_slice( $_SERVER['argv'], 3 );
+
+		// Remove associative arguments.
+		$params = array_filter(
+			$params,
+			function ( $val ) {
```
mukeshpanchal27:
```suggestion
			static function ( $val ) {
```

Per [729](https://github.com/WordPress/performance/pull/729), Use static closures for minor performance improvement whenever instance access is not needed.

mukeshpanchal27 (commented):
Thank you, @ernilambar, for the PR! I've provided some feedbacks. Aside from that, everything appears to be in good from my perspective.

ernilambar:
PR updated as per the review comments.

ernilambar:
@mukeshpanchal27 After last commit, Behat test (PHP 8.0 WP Latest) is failing. Any idea?

swissspidy:
Does it pass locally for you? There might be some error/notice happening

ernilambar:
> Does it pass locally for you? There might be some error/notice happening

I could not make it run Behat tests locally. Some database error occurs and tests fails even before starting. So I usually check those in GitHub Actions.

mukeshpanchal27:
> @mukeshpanchal27 After last commit, Behat test (PHP 8.0 WP Latest) is failing. Any idea?

The [nikic/php-parser](https://github.com/nikic/PHP-Parser/releases/tag/v5.0.0) get updated from 4.18.0 to 5.0.0 in past few hours that may be issue as they update things related to PHP versions. cc. @swissspidy

swissspidy:
> > @mukeshpanchal27 After last commit, Behat test (PHP 8.0 WP Latest) is failing. Any idea?
> 
> The [nikic/php-parser](https://github.com/nikic/PHP-Parser/releases/tag/v5.0.0) get updated from 4.18.0 to 5.0.0 in past few hours that may be issue as they update things related to PHP versions. cc. @swissspidy

This project dos not use PHP-Parser, nor do any of the dependencies.

> I could not make it run Behat tests locally. Some database error occurs and tests fails even before starting. So I usually check those in GitHub Actions.

Have you tried running `composer prepare-behat-tests` first to set up the database? Or run `WP_CLI_TEST_DBTYPE=sqlite composer behat` to test without MySQL.

-----

From what I can see [in the logs](https://github.com/WordPress/plugin-check/actions/workflows/behat-test.yml), this failure started already at https://github.com/WordPress/plugin-check/actions/runs/7443210203 (though it worked fine on the PR branch itself)

So it's not isolated to this PR.

Unfortunately I'm unable to reproduce this locally right now, so let's deal with this in a separate PR to debug in more detail. No blocker for this PR.

mukeshpanchal27:
> Have you tried running `composer prepare-behat-tests` first to set up the database? Or run `WP_CLI_TEST_DBTYPE=sqlite composer behat` to test without MySQL.

I've tested it locally, and it seems to be working fine for me as well. Not entirely sure why the issue persists. I've opened a dedicated issue to track and investigate that. Thank you!

mukeshpanchal27 (approved):

PR #375, opened 2024-01-05
Introduce check category labels

### LINKED ISSUE #373: Introduce proper user-facing names for check categories
The available check categories are currently only defined as a slug, via class constants on `Check_Categories`. While those slugs are the key to the functionality, the categories are also displayed to end users. Right now this is accomplished via `ucfirst( str_replace( '_', ' ', $category ) )`, which really isn't right. We need to use a proper i18n string for each category.

In other words, each available category should also have a label. Here's what I'd suggest:
* Introduce a new method `Check_Categories::get_category_labels()`.
* The method should return an associative array of the category slugs and their respective labels.
* The slugs should be the ones from the class constants, and the labels should be new i18n friendly strings.
* The `Admin_Page` class should then call the new method to have both the category slugs and labels, and the `admin-page.php` template should be adjusted accordingly.
* In case the PR #339 has already been merged, there's another reference to the categories in there that would need to be updated in the same way as well.

Another related observation is that the `Check_Categories::get_categories()` method is unnecessarily complex, using `ReflectionClass` to dynamically get the class constants. I think that's a bit "too smart" and more importantly inefficient, so I think it would be better to simply return a "hard-coded" array including the class constants in there. I think it would be worth making that change here too, while we're enhancing that class anyway.

cc @mukeshpanchal27 @swissspidy @ernilambar

ernilambar:
Closes https://github.com/WordPress/plugin-check/issues/373

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -195,11 +195,10 @@ final class Admin_Page {
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
 		$category_labels = Check_Categories::get_category_labels();
-		$categories      = array_keys( $category_labels );
 
 		// Get user settings for category preferences and set a default value to check all categories by default.
 		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
-		$user_enabled_categories = 'all_categories' === $user_enabled_categories ? $categories : explode( '__', $user_enabled_categories );
+		$user_enabled_categories = 'all_categories' === $user_enabled_categories ? array_keys( $category_labels ) : explode( '__', $user_enabled_categories );
 
 		require WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'templates/admin-page.php';
 	}
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -194,7 +194,8 @@ final class Admin_Page {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$categories = Check_Categories::get_categories();
+		$categories      = Check_Categories::get_categories();
+		$category_labels = Check_Categories::get_category_labels();
 
 		// Get user settings for category preferences and set a default value to check all categories by default.
 		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'all_categories' );
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -194,8 +194,8 @@ final class Admin_Page {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$categories      = Check_Categories::get_categories();
 		$category_labels = Check_Categories::get_category_labels();
+		$categories      = array_keys( $category_labels );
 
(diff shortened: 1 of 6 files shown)


swissspidy:
Now that #339 is merged we can use these in `includes/CLI/Plugin_Check_Command.php`.

ernilambar:
> Now that #339 is merged we can use these in `includes/CLI/Plugin_Check_Command.php`.

PR updated.

swissspidy (approved):

includes/Checker/Check_Categories.php
```diff
@@ -29,27 +29,30 @@ class Check_Categories {
...
+	 *
+	 * @return array An array of category labels.
+	 */
+	public static function get_category_labels() {
+		return array(
+			self::CATEGORY_GENERAL       => esc_html__( 'General', 'plugin-check' ),
+			self::CATEGORY_PLUGIN_REPO   => esc_html__( 'Plugin Repo', 'plugin-check' ),
+			self::CATEGORY_SECURITY      => esc_html__( 'Security', 'plugin-check' ),
+			self::CATEGORY_PERFORMANCE   => esc_html__( 'Performance', 'plugin-check' ),
+			self::CATEGORY_ACCESSIBILITY => esc_html__( 'Accessibility', 'plugin-check' ),
```
felixarntz: I don't think these should be escaped here. Escaping should only happen on output.

includes/Admin/Admin_Page.php
```diff
@@ -194,7 +194,8 @@ public function render_page() {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$categories = Check_Categories::get_categories();
+		$categories      = Check_Categories::get_categories();
+		$category_labels = Check_Categories::get_category_labels();
```
felixarntz: Nit-pick, but this could be simplified to only use `$category_labels`. It already includes the categories as the keys.

We could then also simplify the `admin-page.php` template, to iterate through `$category_labels` for access to both the slugs and the labels, instead of iterating through `$categories` and accessing the labels from the other variable.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -355,12 +355,13 @@ public function list_check_categories( $args, $assoc_args ) {
...
 		return array_map(
-			function ( $slug ) {
+			function ( $slug ) use ( $category_labels ) {
 				return array(
 					'slug' => $slug,
-					'name' => ucfirst( str_replace( '_', ' ', $slug ) ),
+					'name' => $category_labels[ $slug ],
 				);
 			},
 			$categories_slugs
```
felixarntz: Nit-pick: We could use only `$category_labels` here as well as it contains both the slugs and the labels.

felixarntz (commented):
@ernilambar Thanks for the PR, this looks great!

Two non-blocking nit-picks, my only crucial feedback is to not escape the strings too early.

ernilambar:
PR updated.

includes/Admin/Admin_Page.php
```diff
@@ -194,7 +194,8 @@ public function render_page() {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$categories = Check_Categories::get_categories();
+		$category_labels = Check_Categories::get_category_labels();
+		$categories      = array_keys( $category_labels );
```
felixarntz: Not a blocker, but since `$categories` isn't used in the template anymore, we don't need to have it as a variable. You could move the `array_keys( $category_labels )` to the assignment in line 202 instead to avoid the variable.

felixarntz (approved):
Thanks @ernilambar, looks great!

PR #376, opened 2024-01-08
Add check name in results

ernilambar:
In this PR:

* Every `Check` is assumed to have a name. So `get_name()` has been added to `interface Check`.
* `check` has been added to `final class Check_Result`.
* `includes/Traits/Amend_Check_Result.php` - Methods are updated to accomodate newly added `check` field.
* `Late_Escaping_Check.php`: `get_name()` method is implemented. Eg:

```
public function get_name() {
   return 'late_escaping';
}
```
Output:
Screenshot 2024-01-08 at 11 50 03 AM


ernilambar:
Requesting for feedback whether this approach is good or not.

felixarntz:
@ernilambar I like the idea of optionally supporting the inclusion of the check identifier. That said, I think we'll need to carefully think about how to implement that change. Preferably, this should be handled in a central place, without having to define the check identifiers as both array keys and in each class.

ernilambar:
@felixarntz May be this approach is better. Please review it. https://github.com/WordPress/plugin-check/pull/380

PR #378, opened 2024-01-08
Implement field names validation in check command

### LINKED ISSUE #360: Validate field name values passed in `--fields`
When I run following command:

    wp plugin check hello-dolly --fields=nonexistent_name

Output:

    FILE: hello.php
    Error: Invalid field: nonexistent_name.

This validation should be done in the initial.

Expected output:

    Error: Invalid field: nonexistent_name.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/360

* `get_formatter()` takes third optional parameter `$valid_fields`. Validation will be done only if third argument is passed.
* Implemented only in `plugin check` command. In other commands WP-CLI will handle itself.

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -213,7 +213,17 @@ final class Plugin_Check_Command {
 		$default_fields = $this->get_check_default_fields( $assoc_args );
 
 		// Get formatter.
-		$formatter = $this->get_formatter( $assoc_args, $default_fields );
+		$formatter = $this->get_formatter(
+			$assoc_args,
+			$default_fields,
+			array(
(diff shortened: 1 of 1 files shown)


ernilambar:

swissspidy:
Hmm not really a fan of duplicating logic the formatter already gives us for free and adding more complexity. The current output is maybe not super optimal because we're using it in a loop with some heading before it, but I don't think it's a big deal.

felixarntz:
@ernilambar +1 to what @swissspidy says. I personally am unsure this is important to validate. If we want to add validation for arguments, there would probably be a more flexible and useful way than only checking for whether any unsupported parameters are passed. I think it works well enough at this point.

PR #379, opened 2024-01-09
Enhance list-checks command

ernilambar:
Refactor list-checks command with improvements.

* Fix: `get_checks()` of `Default_Check_Repository` by default does not return experimental checks. This PR fix that issue for returing all checks with parameter `Default_Check_Repository::TYPE_ALL | Default_Check_Repository::INCLUDE_EXPERIMENTAL`.
* Addition: `--categories` argument is added to filter checks by categories.
* Addition: `--stability` argument is added to filter checks by stability. Takes either `stable` or `experimental`.

Output:

`wp plugin list-checks --stability=stable` - Currently all checks will be displayed as we dont have experimental checks yet.

stability

`wp plugin list-checks --categories=performance`

single-category

`wp plugin list-checks --categories=performance,general`

multiple-category

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -9,6 +9,7 @@ namespace WordPress\Plugin_Check\CLI;
 
 use Exception;
 use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Checker\Check_Repository;
 use WordPress\Plugin_Check\Checker\CLI_Runner;
 use WordPress\Plugin_Check\Checker\Default_Check_Repository;
 use WordPress\Plugin_Check\Checker\Runtime_Check;
@@ -255,13 +256,8 @@ final class Plugin_Check_Command {
 	 * [--categories]
 	 * : Limit displayed results to include only specific categories.
 	 *
-	 * [--stability=<stability>]
-	 * : Check stability.
-	 * ---
-	 * options:
-	 *   - stable
-	 *   - experimental
-	 * ---
+	 * [--include-experimental]
+	 * : Include experimental checks.
 	 *
 	 * @subcommand list-checks
 	 *
@@ -275,27 +271,42 @@ final class Plugin_Check_Command {
 	public function list_checks( $args, $assoc_args ) {
 		$check_repo = new Default_Check_Repository();
 
-		$all_checks = array();
-
-		foreach ( $check_repo->get_checks( Default_Check_Repository::TYPE_ALL | Default_Check_Repository::INCLUDE_EXPERIMENTAL ) as $key => $check ) {
-			$item = array();
-
-			$item['slug']      = $key;
-			$item['stability'] = strtolower( $check->get_stability() );
-			$item['category']  = $check->get_categories();
-
-			$all_checks[] = $item;
-		}
-
 		// Get options based on the CLI arguments.
 		$options = $this->get_options(
 			$assoc_args,
 			array(
-				'format'     => 'table',
-				'categories' => '',
-				'stability'  => '',
+				'format'               => 'table',
+				'categories'           => '',
+				'include-experimental' => false,
 			)
 		);
+
+		$check_flags = Check_Repository::TYPE_ALL;
+
+		// Check whether to include experimental checks.
+		if ( $options['include-experimental'] ) {
+			$check_flags = $check_flags | Check_Repository::INCLUDE_EXPERIMENTAL;
+		}
+
+		$collection = $check_repo->get_checks( $check_flags );
+
+		// Filters the checks by specific categories.
+		if ( ! empty( $options['categories'] ) ) {
+			$categories = array_map( 'trim', explode( ',', $options['categories'] ) );
+			$collection = Check_Categories::filter_checks_by_categories( $collection, $categories );
+		}
+
+		$all_checks = array();
+
+		foreach ( $collection as $key => $check ) {
+			$item = array();
+
+			$item['slug']      = $key;
+			$item['stability'] = strtolower( $check->get_stability() );
+			$item['category']  = join( ', ', $check->get_categories() );
+
+			$all_checks[] = $item;
+		}
 
 		// Get formatter.
 		$formatter = $this->get_formatter(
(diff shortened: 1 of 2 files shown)


ernilambar:

swissspidy (approved):

includes/CLI/Plugin_Check_Command.php
```diff
@@ -266,18 +277,25 @@ public function list_checks( $args, $assoc_args ) {
 
 		$all_checks = array();
 
-		foreach ( $check_repo->get_checks() as $key => $check ) {
+		foreach ( $check_repo->get_checks( Default_Check_Repository::TYPE_ALL | Default_Check_Repository::INCLUDE_EXPERIMENTAL ) as $key => $check ) {
```
swissspidy: These constants or on the `Check_Repository` class, not `Default_Check_Repository`.

Also, can we do the filtering here instead? The checks collection returned by `get_checks()` has a `filter` method for this purpose.

```suggestion
		foreach ( $check_repo->get_checks( Check_Repository::TYPE_ALL | Check_Repository::INCLUDE_EXPERIMENTAL )->filter(
			static function ( Check $check ) {
				// Filter by $check->get_categories() and $check->get_stability()
				// return bool
			}
		) as $key => $check ) {
```

tests/behat/features/plugin-list-checks.feature
```diff
@@ -14,3 +14,32 @@ Feature: Test that the WP-CLI plugin list checks command works.
...
+      """
+
+    When I run the WP-CLI command `plugin list-checks --format=json --stability=experimental`
+    Then STDOUT should be:
+      """
+      []
+      """
+
+    When I run the WP-CLI command `plugin list-checks --format=csv --categories=general`
+    Then STDOUT should be:
```
mukeshpanchal27:
```suggestion
    Then STDOUT should contain:
```

Given the current situation, this test will pass. However, when new checks are added, it is anticipated to fail. Please modify it to pass only if it contains certain criteria.

tests/behat/features/plugin-list-checks.feature
```diff
@@ -14,3 +14,32 @@ Feature: Test that the WP-CLI plugin list checks command works.
...
+      """
+      slug,category,stability
+      i18n_usage,general,stable
+      """
+
+    When I run the WP-CLI command `plugin list-checks --format=csv --categories="general, security"`
+    Then STDOUT should be:
```
mukeshpanchal27: Check similar feedback comment.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -289,6 +307,35 @@ public function list_checks( $args, $assoc_args ) {
...
+			);
+		}
+
+		// Display categories in better way.
+		array_walk(
+			$all_checks,
+			static function ( &$item ) {
+				$item['category'] = join( ', ', $item['category'] );
+			}
+		);
```
mukeshpanchal27: Could you please explain why the changes needed?

mukeshpanchal27 (changes requested):

includes/CLI/Plugin_Check_Command.php
ernilambar: WP-CLI displays array value like this `["General", "Performance"]`. So we are converting to display like `general, performance`.

includes/CLI/Plugin_Check_Command.php
felixarntz: I think retrieving all checks here is not the right approach. IMO it should behave the same way as `wp plugin check` does:
* By default, only stable checks are returned.
* Instead of a `--stability` option, we should use a boolean flag option `--include-experimental`.
* Only if that flag is set, we should get both stable and experimental checks.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -289,6 +307,35 @@ public function list_checks( $args, $assoc_args ) {
...
 
+		// Filter by stability.
+		if ( ! empty( $options['stability'] ) ) {
+			$all_checks = array_filter(
+				$all_checks,
+				static function ( $item ) use ( $options ) {
+					return $options['stability'] === $item['stability'];
+				}
+			);
+		}
```
felixarntz: See above: This shouldn't be needed as we should instead only retrieve experimental checks in the first place if the flag is set.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -289,6 +307,35 @@ public function list_checks( $args, $assoc_args ) {
...
+
+		// Filter by categories.
+		if ( ! empty( $options['categories'] ) ) {
+			$all_checks = array_filter(
+				$all_checks,
+				static function ( $item ) use ( $options ) {
+					$categories = array_map( 'trim', explode( ',', $options['categories'] ) );
+					return array_intersect( $categories, $item['category'] );
+				}
+			);
```
felixarntz: This should be simplified by reusing `Check_Categories::filter_checks_by_categories()` which is already implemented for this purpose.

includes/CLI/Plugin_Check_Command.php
felixarntz: I don't think this is needed. Before you had the `join()` call when creating the `$all_checks` list, and I think it can be kept that way.

The only thing you'll have to make sure of is that you do all the check filtering using the standard methods on the original check collection, _before_ creating the `$all_checks` variable for CLI output.

felixarntz (changes requested):
@ernilambar I think the ideas here make sense, however the implementation can be simplified.

ernilambar:
Updates:

* `get_checks()` will return stable checks only by default. If `--include-experimental` is given, it will include experimental checks along with stable checks.
* `--categories` - Existing `Check_Categories::filter_checks_by_categories()` is used for filtering checks based on categories.
* `--stability` is removed and implemented `--include-experimental` flag.

felixarntz (commented):
@ernilambar The production code changes look great to me now. However it looks like there are some Behat test failures, see https://github.com/WordPress/plugin-check/actions/runs/7470848675/job/20330167511?pr=379.

cc @swissspidy

swissspidy:
Those Behat failures are unrelated and fixed in #381

felixarntz (approved):
Let's merge #381 and then refresh this branch to make sure the failures are gone :)

Thank you for the PR!

mukeshpanchal27 (approved):

PR #380, opened 2024-01-09
Show check name in results

ernilambar:
Related to https://github.com/WordPress/plugin-check/issues/329

```
wp plugin check pc-sample --fields=line,column,type,code,message,check
```

Sample output:

Screenshot 2024-01-09 at 5 18 46 PM


ernilambar:
Looks like PHPUnit tests also need to be adjusted with new `check` value. I will update those after review.

ernilambar:

swissspidy:
This is an alternative to #376, correct?

ernilambar:
> This is an alternative to #376, correct?

Yes.

includes/Traits/Check_Name.php
```diff
@@ -0,0 +1,33 @@
...
+use ReflectionClass;
+
+/**
+ * Trait for checks name.
+ *
+ * @since n.e.x.t
+ */
+trait Check_Name {
+	/**
+	 * Returns the checks name.
```
mukeshpanchal27:
```suggestion
	 * Returns the check name.
```

includes/Traits/Check_Name.php
```diff
@@ -0,0 +1,33 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string Check name.
+	 */
+	public function get_name() {
```
mukeshpanchal27: I think we should update method name to `get_check_name`. WDYT?

includes/Traits/Check_Name.php
```diff
@@ -0,0 +1,33 @@
...
+		$reflection = new ReflectionClass( $this );
+
+		$class_name = strtolower( $reflection->getShortName() );
+		$class_name = str_replace( '_check', '', $class_name );
+
+		return $class_name;
```
mukeshpanchal27:
```suggestion

		return str_replace( '_check', '', $class_name );
```

includes/Traits/Amend_Check_Result.php
```diff
@@ -41,6 +42,7 @@ protected function add_result_message_for_file( Check_Result $result, $error, $m
 				'line'   => $line,
 				'column' => $column,
 				'link'   => $this->get_file_editor_url( $result, $file, $line ),
+				'check'  => $this->get_name(),
```
mukeshpanchal27: Does it optional? What about admin screen?

includes/Traits/Check_Name.php
```diff
@@ -0,0 +1,33 @@
...
+namespace WordPress\Plugin_Check\Traits;
+
+use ReflectionClass;
+
+/**
+ * Trait for checks name.
+ *
+ * @since n.e.x.t
+ */
+trait Check_Name {
```
mukeshpanchal27: Given that we are currently using this trait in a single place and the code is minimal, it might be more straightforward to incorporate the changes directly into the function using `get_class`. This could lead to a cleaner and more concise implementation.

mukeshpanchal27 (commented):
Thanks @ernilambar for the PR! Left some feedback and questions for consideration.

includes/Traits/Amend_Check_Result.php
ernilambar: This method is supposed to be called by Check only. So I believe we can assume it won't be empty.

includes/Traits/Check_Name.php
ernilambar: Trait is removed.

ernilambar:
PR updated.

swissspidy:
If we display the check name/slug in the results like this, we need to be certain it‘s correct and matches the slug used in Default_Check_Repository. This way users can easily exclude the check next time around.

But what if someone uses the wp_plugin_check_checks filter to add their custom check where this might not be the case (i.e class name differe heavily from the array key)?

IMHO the current solution is not robust enough for this case.

ernilambar:
What if validate slug and class name in `function register_check( $slug, Check $check )`. If the slug and class name does not follow the standard, exception is thrown right away while registering.

Another alternative:
In the filter `wp_plugin_check_checks` rather than accepting `slug=>Class_Name`, we accept class name only. And our code will generate slug accordingly. This way there will be no mismatched slug and class name.

ernilambar:
Closing. We need to explore more robust approach.

PR #382, opened 2024-01-10
Fix exclude checks

### LINKED ISSUE #359: Fatal error when in checks in `--checks` and `--exclude-checks` overlaps
Eg:
`wp plugin check hello-dolly --checks=plugin_review_phpcs,late_escaping --exclude-checks=late_escaping`

Our CLI arguments:

    [--exclude-checks=<checks>]
    Exclude checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping. Applies after evaluating `--checks.`

Since we are calculating excludes after checks, overlapping of checks in the arguments should not generate fatal error.
In the above example scenario, only `plugin_review_phpcs` should be in the final checks list.

**OR**
If we do not expect both arguments in the same command, we should validate and give `WP_CLI::error()`

Output: 
Screenshot 2024-01-10 at 12 17 28 PM

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/359

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -413,6 +413,10 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 			->include( $check_slugs ) // Ensures only the checks with the given slugs are included.
 			->exclude( $excluded_checks ); // Exclude provided checks from list.
 
+		if ( ! empty( $excluded_checks ) ) {
+			$check_slugs = array_diff( $check_slugs, $excluded_checks );
+		}
+
 		// Filters the checks by specific categories.
 		$categories = $this->get_categories();
 		if ( $categories ) {


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -413,6 +413,10 @@ final public function get_checks_to_run() {
 			->include( $check_slugs ) // Ensures only the checks with the given slugs are included.
 			->exclude( $excluded_checks ); // Exclude provided checks from list.
 
+		if ( ! empty( $excluded_checks ) ) {
```
mukeshpanchal27, line 416: Why we check additionally for exclude checks? We already checked above 🤔

mukeshpanchal27 (changes requested):

includes/Checker/Abstract_Check_Runner.php
ernilambar: In $check_slugs, there are all given slugs. After excluding, `$collection` will have list of checks (after excluding). But in next step we are again validating the checks with `$collection->require( $check_slugs )`. So error was triggered from there.

mukeshpanchal27: I've raised PR #387, utilizing a different approach. Please review it when you have a moment.

ernilambar:
Closing in favor of https://github.com/WordPress/plugin-check/pull/387

PR #383, opened 2024-01-10
Implement `try` in checks preparation

ernilambar:


ernilambar:
Adding `try` in `final public function prepare()` in `abstract class Abstract_Check_Runner` hides that fatal error. But I doubt keeping try there with empty `catch` block is good approach.

ernilambar:
Closing this. We need to research for more robust approach.

PR #384, opened 2024-01-11
Run checks in the context of the minimum WP version supported by a plugin

dd32:
Some PHPCS rules are WordPress version dependent, only being triggered for older (or newer) versions of WordPress.

One such rule is this, which is triggered unless the minimum wp is set to 6.2:
```
LINE  84: ERROR Unsupported placeholder used in $wpdb->prepare(). Found: "%i".
                (WordPress.DB.PreparedSQLPlaceholders.UnsupportedPlaceholder)
>>  84:  ········$count·=·$wpdb->get_var($wpdb->prepare("SELECT·COUNT(*)·FROM·%i",·$table_name));
```

WPCS allows setting the minimum required WP version, which this PR implements
https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#setting-minimum-supported-wp-version-from-the-command-line-wordpresscs-0140

I've not added a unit test for this, as I wasn't immediately able to figure out how to do so, but this PR works in my initial testing.

### DIFF
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -80,25 +80,7 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 		$orig_cmd_args = $_SERVER['argv'];
 
 		// Create the default arguments for PHPCS.
-		$defaults = array(
-			'',
-			$result->plugin()->location(),
-			'--report=Json',
-			'--report-width=9999',
-		);
-
-		$directories_to_ignore = Plugin_Request_Utility::get_directories_to_ignore();
-		if ( ! empty( $directories_to_ignore ) ) {
-			$defaults[] = '--ignore=*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
-		}
-
-		// Set the Minimum WP version supported for the plugin.
-		if ( $result->plugin()->minimum_supported_wp() ) {
-			// Due to the syntax of runtime-set, these must be passed as individual args.
-			$defaults[] = '--runtime-set';
-			$defaults[] = 'minimum_wp_version';
-			$defaults[] = $result->plugin()->minimum_supported_wp();
-		}
+		$defaults = $this->get_argv_defaults( $result );
 
 		// Set the check arguments for PHPCS.
 		$_SERVER['argv'] = $this->parse_argv( $this->get_args(), $defaults );
@@ -162,6 +144,38 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 		// Format check arguments for PHPCS.
 		foreach ( $check_args as $key => $value ) {
 			$defaults[] = "--{$key}=$value";
+		}
+
+		return $defaults;
+	}
+
+	/**
+	 * Gets the default command arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Check_Result $result The check result to amend, including the plugin context to check.
+	 * @return array An indexed array of PHPCS CLI arguments.
+	 */
+	private function get_argv_defaults( Check_Result $result ): array {
+		$defaults = array(
+			'',
+			$result->plugin()->location(),
+			'--report=Json',
+			'--report-width=9999',
+		);
+
+		$directories_to_ignore = Plugin_Request_Utility::get_directories_to_ignore();
+		if ( ! empty( $directories_to_ignore ) ) {
+			$defaults[] = '--ignore=*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
+		}
+
+		// Set the Minimum WP version supported for the plugin.
+		if ( $result->plugin()->minimum_supported_wp() ) {
+			// Due to the syntax of runtime-set, these must be passed as individual args.
+			$defaults[] = '--runtime-set';
+			$defaults[] = 'minimum_wp_version';
+			$defaults[] = $result->plugin()->minimum_supported_wp();
 		}
 
 		return $defaults;
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -92,6 +92,14 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 			$defaults[] = '--ignore=*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
 		}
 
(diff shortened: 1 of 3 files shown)


includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -92,6 +92,14 @@ final public function run( Check_Result $result ) {
 			$defaults[] = '--ignore=*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
 		}
 
+		// Set the Minimum WP version supported for the plugin.
+		if ( $result->plugin()->minimum_supported_wp() ) {
+			// Due to the syntax of runtime-set, these must be passed as individual args.
+			$defaults[] = '--runtime-set';
+			$defaults[] = 'minimum_wp_version';
+			$defaults[] = $result->plugin()->minimum_supported_wp();
```
dd32: I guess this should probably be wrapped in escapeshellarg() as it's technically a command line? I'm not sure this is actually required though, nor if it'll work..
```suggestion
			$defaults[] = escapeshellarg( $result->plugin()->minimum_supported_wp() );
```

For context of the syntax requiring individual args, this is handled here:
https://github.com/squizlabs/PHP_CodeSniffer/blob/7763e2e1f773cb0615ed8afa133189fc804f583d/src/Config.php#L828-L847 which doesn't have a handler for `--runtime-set=...` in the `default:`: https://github.com/squizlabs/PHP_CodeSniffer/blob/7763e2e1f773cb0615ed8afa133189fc804f583d/src/Config.php#L848-L849

includes/Traits/Find_Readme.php
```diff
@@ -25,7 +25,7 @@ trait Find_Readme {
 	 */
 	protected function filter_files_for_readme( array $files, $plugin_relative_path ) {
 		// Find the readme file.
-		$readme_list = self::filter_files_by_regex( $files, '/readme\.(txt|md)$/i' );
+		$readme_list = preg_grep( '/readme\.(txt|md)$/i', $files );
```
dd32, line 28: This change was added as `filter_files_by_regex()` is provided by [Abstract_File_Check](https://github.com/WordPress/plugin-check/blob/a24d647a3e48e6b8f5daa87fd9002f730c2c1e1f/includes/Checker/Checks/Abstract_File_Check.php#L114-L116) and moving the method from there to somewhere else didn't feel appropriate.

As the method is nothing but a wrapper for `preg_grep()` it's removal seems highly appropriate.

joemcgill:
Thanks @dd32. This does seem like something the Plugin Check plugin should account for. It would be great to get some tests added for this functionality if you have time.

Additionally, it seems this issue currently effects the current version of the plugin published to the .org repo, which is currently in the `legacy-plugin` branch of the repo, so applying a similar change there may be needed until https://github.com/WordPress/plugin-check/issues/283 is resolved and version 1.0.0 can be released.

includes/Traits/Find_Readme.php
felixarntz: Good catch on decoupling it. That was an oversight indeed (originally this method used to be in a class that extended `Abstract_File_Check`).

I'm not sure there's value in removing the method entirely since the class provides abstractions related to filtering files (`filter_files_by_regex()` is a more intuitive name for what the method is for in context of the class compared to `preg_grep()`). Anyway, that would be unrelated to this PR :)

includes/Plugin_Context.php
```diff
@@ -98,4 +102,34 @@ public function location() {
...
+
+	/**
+	 * Determine the minimum supported WordPress version of the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool|string The minimum version supported, otherwise false.
+	 */
+	public function minimum_supported_wp() {
+		$headers = get_plugin_data( $this->main_file );
```
felixarntz: I think the logic in this method should be memoized as it is quite expensive to parse this. We shouldn't do that more than once per plugin per request, even if the method is called more than once.

Can you add a protected or private class property that stores this upon first call, and then returns the value from the property if already set on subsequent calls?

includes/Plugin_Context.php
```diff
@@ -98,4 +102,34 @@ public function location() {
 	public function is_single_file_plugin() {
 		return $this->path() !== $this->location();
 	}
+
+	/**
+	 * Determine the minimum supported WordPress version of the plugin.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool|string The minimum version supported, otherwise false.
```
felixarntz: Nit-pick: A cleaner implementation would be to always return a `string`. Better to avoid multiple return types if possible.
```suggestion
	 * @return string The minimum version supported, or empty string if unknown.
```

includes/Plugin_Context.php
```diff
@@ -98,4 +102,34 @@ public function location() {
...
+			if ( $readme_file ) {
+				$parser = new Parser( $readme_file );
+
+				if ( ! empty( $parser->requires ) ) {
+					return $parser->requires;
+				}
+			}
+		}
+
+		return false;
```
felixarntz: See above.
```suggestion
		return '';
```

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
felixarntz: Not sure whether it's needed 🤔 

PHPCS isn't run in command line context in this branch, so I don't think there's any risk for executing other command line commands here. I believe the only risk from malicious data here would be to cause a failure in PHPCS.

I think it's okay to leave it out for now, as I'm unsure whether there are any side effects from adding it.

felixarntz (commented):
Thanks @dd32, great catch! I left some feedback below.

includes/Plugin_Context.php
dd32: Yeah I agree with this, and was considering adding a static var, but I wasn't sure how this actually operates in practice.

Feel free to push whatever changes to the PR that's needed for a merge

dd32:
> It would be great to get some tests added for this functionality if you have time.

Feel free to add them, as I said, I couldn't figure out how to in the overcomplicated abstractions.

> Additionally, it seems this issue currently effects the current version of the plugin published to the .org repo, which is currently in the `legacy-plugin` branch of the repo, so applying a similar change there may be needed until https://github.com/WordPress/plugin-check/issues/283 is resolved and version 1.0.0 can be released.

Probably, but I don't think that's a priority, since AFAIK the legacy version should be abandoned in its entirety based on how things have been moving

felixarntz (approved):
Thanks @dd32!

swissspidy (approved):

PR #387, opened 2024-01-11
Fix: Fatal error when in checks in `--checks` and `--exclude-checks` overlaps

### LINKED ISSUE #359: Fatal error when in checks in `--checks` and `--exclude-checks` overlaps

mukeshpanchal27:
Closes #359

### Checklist:

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -419,7 +419,9 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 			$collection = Check_Categories::filter_checks_by_categories( $collection, $categories );
 		}
 
-		return $collection->to_map();
+		return $collection
+			->require( array_keys( $collection->to_map() ) ) // Ensures all of the given slugs are valid.
+			->to_map();
 	}
 
 	/**
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -410,6 +410,7 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 		$excluded_checks = $this->get_check_exclude_slugs();
 
 		$collection = $this->check_repository->get_checks( $check_flags )
+			->require( $check_slugs ) // Ensures all of the given slugs are valid.
 			->include( $check_slugs ) // Ensures only the checks with the given slugs are included.
 			->exclude( $excluded_checks ); // Exclude provided checks from list.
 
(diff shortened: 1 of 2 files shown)


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -419,9 +419,7 @@ final public function get_checks_to_run() {
 			$collection = Check_Categories::filter_checks_by_categories( $collection, $categories );
 		}
 
-		return $collection
-			->require( $check_slugs ) // Ensures all of the given slugs are valid.
```
ernilambar, line 423: You have removed `require( $check_slugs )`. Are we sure we want to remove validation altogether?

mukeshpanchal27: Missed that completely nice catch. Reverted in https://github.com/WordPress/plugin-check/pull/387/commits/9fb14284cea537776679592dfd7435965e149b97

ernilambar (approved):

swissspidy (approved):

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -420,7 +420,7 @@ final public function get_checks_to_run() {
 		}
 
 		return $collection
-			->require( $check_slugs ) // Ensures all of the given slugs are valid.
+			->require( array_keys( $collection->to_map() ) ) // Ensures all of the given slugs are valid.
```
felixarntz: This line is now useless as far as I can tell. The `require()` should be called to avoid people passing check slugs that aren't available.

I think a proper fix would be to instead leave the `require()` as is, but move it between lines 412 and 413, i.e. before the exclusions are applied.

This way the requirement is maintained without causing the error as far as I can tell.

felixarntz (changes requested):
I think the fix here currently breaks another _intentional_ functionality.

includes/Checker/Abstract_Check_Runner.php
mukeshpanchal27: Thanks for the feedback. Initially i think do that but thanks for pointing. Addressed in https://github.com/WordPress/plugin-check/pull/387/commits/5fc1a149806a3125f4c101ef02603c16a4e9ee3a

felixarntz (approved):
Thanks @mukeshpanchal27!

PR #389, opened 2024-01-12
Validate check categories

### LINKED ISSUE #348: Display warning if non existent parameters are passed in `--checks` and `--categories`
Proper warning message should be displayed rather than fatal error when non existent check is passed to `--checks`

`wp plugin check hello-universe --checks=nonexistent_check`

Currently it generates PHP fatal error. 

    Fatal error: Uncaught Exception: Check with the slug "nonexistent_check" does not exist. in /var/www/html/wp-content/plugins/plugin-check/includes/Checker/Default_Check_Collection.php:158

May be we should display warning message like

`WARNING: Invalid check slug 'nonexistent_check'`

We could implement similar for categories also.

`wp plugin check hello-universe --categories=nonexistent_category`

ernilambar:
This PR partially fixes https://github.com/WordPress/plugin-check/issues/348

* Fetching of checks and check categories are different in our setup. So it is not viable to use same approach for the validation.
* Validation is implemented in `plugin check` and `plugin list-checks` commands.
* If invalid categories are passed, there will be ERROR rather than warning.


ernilambar:
Blocked by https://github.com/WordPress/plugin-check/issues/390

ernilambar:
Superseded by https://github.com/WordPress/plugin-check/issues/153
Closing issue.

PR #391, opened 2024-01-15
Refactor CLI class

### LINKED ISSUE #390: Refactor `Plugin_Check_Command` class to reduce `ExcessiveClassComplexity`
Would it make sense to add new abstract `Abstract_Plugin_Check_Command` class so that we can move few common methods from the current `Plugin_Check_Command` class?

Current Cyclomatic Complexity of this class around the higher limit (50). So any new addition triggers error in PHPMD.

Common and utility methods that could be moved to another class:

* `get_options()`
* `get_formatter()`
* `flatten_file_results()`
* `display_results()`
* `has_runtime_check()`

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/390

With this PR, `CyclomaticComplexity` for the classes are as below:

* `Abstract_Plugin_Check_Command` - 21
* `Plugin_Check_Command` - 28

### DIFF
--- a/includes/CLI/Abstract_Plugin_Check_Command.php
+++ b/includes/CLI/Abstract_Plugin_Check_Command.php
@@ -1,0 +1,202 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\CLI\Abstract_Plugin_Check_Command
+ *
+ * @package plugin-check
+ */
+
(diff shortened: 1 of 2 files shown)


ernilambar:
Closing in favor of https://github.com/WordPress/plugin-check/issues/153

PR #392, opened 2024-01-15
Check upgrade notice in plugin readme

ernilambar:
If readme file has upgrade notice is longer than 300 characters then `WARNING` will be shown.

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -320,7 +320,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	/**
 	 * Checks the readme file upgrade notice.
 	 *
-	 * @since n.e.x.t
+	 * @since 1.0.0
 	 *
 	 * @param Check_Result $result      The Check Result to amend.
 	 * @param string       $readme_file Readme file.
@@ -339,11 +339,11 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		foreach ( $notices as $version => $notice ) {
 			if ( strlen( $notice ) > $maximum_characters ) {
 				if ( empty( $version ) ) {
-					/* translators: Maximum characters. */
-					$message = sprintf( __( 'The upgrade notice exceeds the limit of %d characters.', 'plugin-check' ), $maximum_characters );
+					/* translators: %s: maximum limit. */
+					$message = sprintf( _n( 'The upgrade notice exceeds the limit of %d character.', 'The upgrade notice exceeds the limit of %d characters.', $maximum_characters, 'plugin-check' ), $maximum_characters );
 				} else {
-					/* translators: 1: Version, 2: Maximum characters. */
-					$message = sprintf( __( 'The upgrade notice for \'%1$s\' exceeds the limit of %2$d characters.', 'plugin-check' ), $version, $maximum_characters );
+					/* translators: 1: version, 2: maximum limit. */
+					$message = sprintf( _n( 'The upgrade notice for "%1$s" exceeds the limit of %2$d character.', 'The upgrade notice for "%1$s" exceeds the limit of %2$d characters.', $maximum_characters, 'plugin-check' ), $version, $maximum_characters );
 				}
 
 				$this->add_result_warning_for_file( $result, $message, 'upgrade_notice_limit', $readme_file );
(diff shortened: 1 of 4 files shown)


ernilambar:

mukeshpanchal27:
Is this related to plugin requirements? If so, could you please provide the URL? I couldn't find it in the `legacy-plugin` repository.

ernilambar:
I took reference from official readme file. https://wordpress.org/plugins/readme.txt & https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

```
== Upgrade Notice ==

= 1.0 =
Upgrade notices describe the reason a user should upgrade.  No more than 300 characters.
```

mukeshpanchal27:
Per [WP directory plugin](https://wpdirectory.net/search/01HM8W02J29NK8SEFB38C7XGDS) list there is more then 30k+ plugins that have `Upgrade Notice` section. Note entirely sure but it's not official requirement for the plugin. I would like to know @felixarntz and @swissspidy though on this. Thanks!

The current approach pass for this upgrade notice 🤔 

Plugin: [Redirection](https://wordpress.org/plugins/redirection/)

```
== Upgrade Notice ==

= 5.4 =
* You may need to configure the IP header option if using a proxy

= 3.0 =
* Upgrades the database to support IPv6. Please backup your data and visit the Redirection settings to perform the upgrade
* Switches to the WordPress REST API
* Permissions changed from 'administrator' role to 'manage_options' capability

= 3.6.1 =
* Note Redirection will not work with PHP < 5.4 after 3.6 - please upgrade your PHP

= 3.7 =
* Requires minimum PHP 5.4. Do not upgrade if you are still using PHP < 5.4

= 4.0 =
* Alters database to support case insensitivity, trailing slashes, and query params. Please backup your data

= 4.7 =
* Requires minimum PHP 5.6+. Do not upgrade if you are still using PHP < 5.6

= 4.9 =
* Alters database to support enhanced logging. Please backup your data
```

swissspidy:
> Per [WP directory plugin](https://wpdirectory.net/search/01HM8W02J29NK8SEFB38C7XGDS) list there is more then 30k+ plugins that have `Upgrade Notice` section. Note entirely sure but it's not official requirement for the plugin. I would like to know @felixarntz and @swissspidy though on this. Thanks!

Correct, having a section itself is not a requirement, but if there is one provided, the example readme suggests using max 300 characters.

This is probably more a question for the plugin review team, though upgrade notices are not really a thing on initial submission, so it won't really apply there.

The question also is whether this 300 character limit is still correct and enforced.

> The current approach pass for this upgrade notice

@mukeshpanchal27 I _think_ the 300 characters limit is per upgrade notice _per version_, not for the whole section.

From what I can see this change here only looks at the first notice, not each of them individually. So that would need to be improved in the PR to make it more robust.

ernilambar:
PR updated.

ernilambar:
@foosantos Will this be worth adding to the check? It is not useful for the initial plugin submission though. But could be helpful when the plugin is updated.

felipevelzani:
Hi @ernilambar — I'm going to check with the team on this one. Thanks for working on it.

felipevelzani:
cc @bordoni

bordoni:
Tests are failing on 8.1, lets get that resolved. But yest, I like that change.

davidperezgar:
Since it's a `WARNING`, I don't see a problem and suggest developer to make best practices. 
I have checked the interface and is working correctly. For me it's OK!

davidperezgar (approved):
Everything seems OK!

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -314,6 +317,40 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
 		}
 	}
 
+	/**
+	 * Checks the readme file upgrade notice.
+	 *
+	 * @since 1.0.0
```
swissspidy:
```suggestion
	 * @since 1.0.2
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -314,6 +317,40 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+
+		// Bail if no upgrade notices.
+		if ( 0 === count( $notices ) ) {
+			return;
+		}
+
+		foreach ( $notices as $version => $notice ) {
+			if ( strlen( $notice ) > $maximum_characters ) {
+				if ( empty( $version ) ) {
+					/* translators: %s: maximum limit. */
```
swissspidy:
```suggestion
					/* translators: %d: maximum limit. */
```

swissspidy (approved):

PR #395, opened 2024-01-19
Make parser warnings more clear

### LINKED ISSUE #394: Make more clear about readme parser warnings
If there are warnings returned from the Parser, we display message like this:

```
The following readme parser warnings were detected: tested_header_ignored, requires_php_header_ignored
```

Message here is not quite clear. May be we could display more clearer messages for each warnings.

Existing warnings in the Parser:

* contributor_ignored
* requires_php_header_ignored
* tested_header_ignored
* requires_header_ignored

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/394

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -286,13 +286,22 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	 * @return string Stable WordPress version.
 	 */
 	private function get_wordpress_stable_version() {
-		$version = get_bloginfo( 'version' );
+		$version = get_transient( 'wp_plugin_check_latest_wp_version' );
 
-		// Strip off any -alpha, -RC, -beta suffixes.
-		list( $version, ) = explode( '-', $version );
+		if ( false === $version ) {
+			$response = wp_remote_get( 'https://api.wordpress.org/core/version-check/1.7/' );
 
-		if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
-			$version = $matches[0];
+			if ( ! is_wp_error( $response ) ) {
+				$body = json_decode( wp_remote_retrieve_body( $response ), true );
+
+				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
+					$latest_release = reset( $body['offers'] );
+
+					$version = $latest_release['new_bundled'];
+
+					set_transient( 'wp_plugin_check_latest_wp_version', $version, DAY_IN_SECONDS );
+				}
+			}
 		}
 
 		return $version;
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -212,29 +212,29 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		$messages = array(
 			'contributor_ignored'          => sprintf(
 				/* translators: %s: plugin header tag */
-				__( 'One or more contributors listed were ignored. The %s field should only contain WordPress.org usernames. Remember that usernames are case-sensitive.', 'plugin-check' ),
-				"'Contributors'"
+				__( 'One or more contributors listed were ignored. The "%s" field should only contain WordPress.org usernames. Remember that usernames are case-sensitive.', 'plugin-check' ),
+				'Contributors'
 			),
 			'requires_php_header_ignored'  => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.2.4. 3: Example version 7.0. */
-				__( 'The %1$s field was ignored. This field should only contain a PHP version such as %2$s or %3$s.', 'plugin-check' ),
-				"'Requires PHP'",
-				"'5.2.4'",
-				"'7.0'"
+				__( 'The "%1$s" field was ignored. This field should only contain a PHP version such as "%2$s" or "%3$s".', 'plugin-check' ),
+				'Requires PHP',
+				'5.2.4',
+				'7.0'
 			),
 			'tested_header_ignored'        => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 5.1. */
-				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
-				"'Tested up to'",
-				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
-				"'" . number_format( $latest_wordpress_version + 0.1, 1 ) . "'"
+				__( 'The "%1$s" field was ignored. This field should only contain a valid WordPress version such as "%2$s" or "%3$s".', 'plugin-check' ),
+				'Tested up to',
+				number_format( $latest_wordpress_version, 1 ),
+				number_format( $latest_wordpress_version + 0.1, 1 )
 			),
 			'requires_header_ignored'      => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 4.9. */
-				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
-				"'Requires at least'",
-				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
-				"'" . number_format( $latest_wordpress_version - 0.1, 1 ) . "'"
+				__( 'The "%1$s" field was ignored. This field should only contain a valid WordPress version such as "%2$s" or "%3$s".', 'plugin-check' ),
+				'Requires at least',
+				number_format( $latest_wordpress_version, 1 ),
+				number_format( $latest_wordpress_version - 0.1, 1 )
 			),
 			'too_many_tags'                => sprintf(
 				/* translators: %d: maximum tags limit */
@@ -244,17 +244,17 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 			'ignored_tags'                 => sprintf(
 				/* translators: %s: list of tags not supported */
 				__( 'One or more tags were ignored. The following tags are not permitted: %s', 'plugin-check' ),
-				"'" . implode( "', '", $parser->ignore_tags ) . "'"
+				'"' . implode( '", "', $parser->ignore_tags ) . '"'
 			),
 			'no_short_description_present' => sprintf(
 				/* translators: %s: section title */
-				__( 'The %s section is missing. An excerpt was generated from your main plugin description.', 'plugin-check' ),
-				"'Short Description'"
+				__( 'The "%s" section is missing. An excerpt was generated from your main plugin description.', 'plugin-check' ),
+				'Short Description'
 			),
 			'trimmed_short_description'    => sprintf(
 				/* translators: 1: section title; 2: maximum limit */
-				_n( 'The %1$s section is too long and was truncated. A maximum of %2$d character is supported.', 'The %1$s section is too long and was truncated. A maximum of %2$d characters is supported.', 150, 'plugin-check' ),
-				"'Short Description'",
+				_n( 'The "%1$s" section is too long and was truncated. A maximum of %2$d character is supported.', 'The "%1$s" section is too long and was truncated. A maximum of %2$d characters is supported.', 150, 'plugin-check' ),
+				'Short Description',
 				150
 			),
 		);
@@ -291,7 +291,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		if ( false === $version ) {
 			$response = wp_remote_get( 'https://api.wordpress.org/core/version-check/1.7/' );
 
-			if ( ! is_wp_error( $response ) ) {
+			if ( ! is_wp_error( $response ) && 200 === wp_remote_retrieve_response_code( $response ) ) {
 				$body = json_decode( wp_remote_retrieve_body( $response ), true );
 
 				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -207,31 +207,52 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		);
 
 		$messages = array(
-			'contributor_ignored'         => sprintf(
+			'contributor_ignored'          => sprintf(
 				/* translators: %s: plugin header tag */
 				__( 'One or more contributors listed were ignored. The %s field should only contain WordPress.org usernames. Remember that usernames are case-sensitive.', 'plugin-check' ),
 				"'Contributors'"
 			),
-			'requires_php_header_ignored' => sprintf(
+			'requires_php_header_ignored'  => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.2.4. 3: Example version 7.0. */
 				__( 'The %1$s field was ignored. This field should only contain a PHP version such as %2$s or %3$s.', 'plugin-check' ),
 				"'Requires PHP'",
 				"'5.2.4'",
 				"'7.0'"
 			),
-			'tested_header_ignored'       => sprintf(
+			'tested_header_ignored'        => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 5.1. */
 				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
 				"'Tested up to'",
 				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
 				"'" . number_format( $latest_wordpress_version + 0.1, 1 ) . "'"
 			),
-			'requires_header_ignored'     => sprintf(
+			'requires_header_ignored'      => sprintf(
 				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 4.9. */
 				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
 				"'Requires at least'",
 				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
 				"'" . number_format( $latest_wordpress_version - 0.1, 1 ) . "'"
+			),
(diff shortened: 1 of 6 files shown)


ernilambar:
https://meta.trac.wordpress.org/ticket/7412 This will add more parser warnings. It'd be better if we incorporate those changes here also.

ernilambar:
PR is updated and new warnings are also incorporated.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,69 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
 	private function check_for_warnings( Check_Result $result, string $readme_file, Parser $parser ) {
 		$warnings = $parser->warnings ? $parser->warnings : array();
 
+		// This should be ERROR rather than WARNING. So ignoring here to handle separately.
+		if ( isset( $warnings['invalid_plugin_name_header'] ) ) {
+			unset( $warnings['invalid_plugin_name_header'] );
+		}
```
swissspidy: Unset does not require an `isset()` check.

```suggestion
		unset( $warnings['invalid_plugin_name_header'] );
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -217,16 +274,29 @@ private function check_for_warnings( Check_Result $result, string $readme_file,
...
+
+		// Strip off any -alpha, -RC, -beta suffixes.
+		list( $version, ) = explode( '-', $version );
+
+		if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
+			$version = $matches[0];
 		}
+
+		return $version;
 	}
```
swissspidy: This assumes that the site the plugin runs on is using the latest WordPress version, which is not necessarily the case.

To truly get the latest available release (which seems helpful for the error message) we should probably do something like grab the first result from https://api.wordpress.org/core/version-check/1.7/ and store that in a transient for 24 hours or so.

What do you think?

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,69 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+				__( 'The %1$s field was ignored. This field should only contain a PHP version such as %2$s or %3$s.', 'plugin-check' ),
+				"'Requires PHP'",
+				"'5.2.4'",
+				"'7.0'"
+			),
+			'tested_header_ignored'        => sprintf(
+				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 5.1. */
+				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
+				"'Tested up to'",
+				"'" . number_format( floatval( $latest_wordpress_version ), 1 ) . "'",
```
swissspidy: Nit: here and elsewhere I would use `(float)` instead of `floatval()`.

Even better, just cast it to a float once directly inside `get_wordpress_stable_version()` before returning the version. Then you don't need the 4 type casts here.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,69 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+				"'" . implode( "', '", $parser->ignore_tags ) . "'"
+			),
+			'no_short_description_present' => sprintf(
+				/* translators: %s: section title */
+				__( 'The %s section is missing. An excerpt was generated from your main plugin description.', 'plugin-check' ),
+				"'Short Description'"
+			),
+			'trimmed_short_description'    => sprintf(
+				/* translators: 1: section title; 2 maximum limit */
+				__( 'The %1$s section is too long and was truncated. A maximum of %2$d characters is supported.', 'plugin-check' ),
```
swissspidy: Mhm I think we'd want to use `_n()` here because the number of characters can be singular or plural.

```suggestion
				_n( 'The %1$s section is too long and was truncated. A maximum of %2$d character is supported.', 'The %1$s section is too long and was truncated. A maximum of %2$d characters is supported.', 150, 'plugin-check' ),
```

includes/Checker/Checks/Plugin_Readme_Check.php
ernilambar: Transient implemented.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -217,16 +272,38 @@ private function check_for_warnings( Check_Result $result, string $readme_file,
...
+					$latest_release = reset( $body['offers'] );
+
+					$version = $latest_release['new_bundled'];
+
+					set_transient( 'wp_plugin_check_latest_wp_version', $version, DAY_IN_SECONDS );
+				}
+			}
 		}
+
+		return $version;
```
swissspidy, line 319: If the HTTP request failed, this will still be false. Maybe we should keep the bloginfo logic as a fallback?

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
 	private function check_for_warnings( Check_Result $result, string $readme_file, Parser $parser ) {
 		$warnings = $parser->warnings ? $parser->warnings : array();
 
+		// This should be ERROR rather than WARNING. So ignoring here to handle separately.
```
mukeshpanchal27, line 201:
```suggestion
		// TODO: This should be ERROR rather than WARNING. So ignoring here to handle separately.
```

Mark it as "TODO" so that others are aware that this should be implemented sooner rather than later.

ernilambar: I have already made PR for this. https://github.com/WordPress/plugin-check/pull/398

swissspidy (approved):

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+
 		$ignored_warnings = array(
 			'contributor_ignored',
 		);
 
+		$messages = array(
+			'contributor_ignored'          => sprintf(
+				/* translators: %s: plugin header tag */
+				__( 'One or more contributors listed were ignored. The %s field should only contain WordPress.org usernames. Remember that usernames are case-sensitive.', 'plugin-check' ),
+				"'Contributors'"
```
mukeshpanchal27:
```suggestion
				__( 'One or more contributors listed were ignored. The "%s" field should only contain WordPress.org usernames. Remember that usernames are case-sensitive.', 'plugin-check' ),
				'Contributors'
```

Not blocker but for constancy let's use double quote.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'requires_php_header_ignored'  => sprintf(
+				/* translators: 1: plugin header tag; 2: Example version 5.2.4. 3: Example version 7.0. */
+				__( 'The %1$s field was ignored. This field should only contain a PHP version such as %2$s or %3$s.', 'plugin-check' ),
+				"'Requires PHP'",
+				"'5.2.4'",
+				"'7.0'"
```
mukeshpanchal27:
```suggestion
				__( 'The "%1$s" field was ignored. This field should only contain a PHP version such as "%2$s" or "%3$s".', 'plugin-check' ),
				'Requires PHP',
				'5.2.4',
				'7.0'
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'tested_header_ignored'        => sprintf(
+				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 5.1. */
+				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
+				"'Tested up to'",
+				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
+				"'" . number_format( $latest_wordpress_version + 0.1, 1 ) . "'"
```
mukeshpanchal27:
```suggestion
				__( 'The "%1$s" field was ignored. This field should only contain a valid WordPress version such as "%2$s" or "%3$s".', 'plugin-check' ),
				'Tested up to',
				number_format( $latest_wordpress_version, 1 ),
				number_format( $latest_wordpress_version + 0.1, 1 )
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'requires_header_ignored'      => sprintf(
+				/* translators: 1: plugin header tag; 2: Example version 5.0. 3: Example version 4.9. */
+				__( 'The %1$s field was ignored. This field should only contain a valid WordPress version such as %2$s or %3$s.', 'plugin-check' ),
+				"'Requires at least'",
+				"'" . number_format( $latest_wordpress_version, 1 ) . "'",
+				"'" . number_format( $latest_wordpress_version - 0.1, 1 ) . "'"
```
mukeshpanchal27:
```suggestion
				__( 'The "%1$s" field was ignored. This field should only contain a valid WordPress version such as "%2$s" or "%3$s".', 'plugin-check' ),
				'Requires at least',
				number_format( $latest_wordpress_version, 1 ),
				number_format( $latest_wordpress_version - 0.1, 1 )
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'too_many_tags'                => sprintf(
+				/* translators: %d: maximum tags limit */
+				__( 'One or more tags were ignored. Please limit your plugin to %d tags.', 'plugin-check' ),
+				5
+			),
+			'ignored_tags'                 => sprintf(
+				/* translators: %s: list of tags not supported */
+				__( 'One or more tags were ignored. The following tags are not permitted: %s', 'plugin-check' ),
+				"'" . implode( "', '", $parser->ignore_tags ) . "'"
```
mukeshpanchal27:
```suggestion
				__( 'One or more tags were ignored. The following tags are not permitted: "%s"', 'plugin-check' ),
				implode( "', '", $parser->ignore_tags )
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'no_short_description_present' => sprintf(
+				/* translators: %s: section title */
+				__( 'The %s section is missing. An excerpt was generated from your main plugin description.', 'plugin-check' ),
+				"'Short Description'"
```
mukeshpanchal27:
```suggestion
				__( 'The "%s" section is missing. An excerpt was generated from your main plugin description.', 'plugin-check' ),
				'Short Description'
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -198,12 +198,67 @@ private function check_stable_tag( Check_Result $result, string $readme_file, Pa
...
+			),
+			'trimmed_short_description'    => sprintf(
+				/* translators: 1: section title; 2: maximum limit */
+				_n( 'The %1$s section is too long and was truncated. A maximum of %2$d character is supported.', 'The %1$s section is too long and was truncated. A maximum of %2$d characters is supported.', 150, 'plugin-check' ),
+				"'Short Description'",
```
mukeshpanchal27:
```suggestion
				_n( 'The "%1$s" section is too long and was truncated. A maximum of "%2$d" character is supported.', 'The "%1$s" section is too long and was truncated. A maximum of "%2$d" characters is supported.', 150, 'plugin-check' ),
				'Short Description',
```

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -217,16 +272,50 @@ private function check_for_warnings( Check_Result $result, string $readme_file,
...
+
+	/**
+	 * Returns current major WordPress version.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string Stable WordPress version.
+	 */
+	private function get_wordpress_stable_version() {
+		$version = get_transient( 'wp_plugin_check_latest_wp_version' );
```
mukeshpanchal27, line 289: Could you add tests that verify the behaviour of RC or Beta versions? Set transient value for WP and check if it return proper values.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -217,16 +272,50 @@ private function check_for_warnings( Check_Result $result, string $readme_file,
...
+
+		if ( false === $version ) {
+			$response = wp_remote_get( 'https://api.wordpress.org/core/version-check/1.7/' );
+
+			if ( ! is_wp_error( $response ) ) {
```
mukeshpanchal27:
```suggestion
			if ( ! is_wp_error( $response ) && 200 === wp_remote_retrieve_response_code( $response ) ) {
```
For safe guard.

mukeshpanchal27 (changes requested):
Thanks @ernilambar for the PR. Great start. Left some feedback for your consideration.

Additionally.
- Add unit tests that demonstrate all warnings that you listed.
- Add unit test that ignore all these warnings. I think we can archive it using `add_filter( 'wp_plugin_check_ignored_readme_warnings', '__return_empty_array' );`.

includes/Checker/Checks/Plugin_Readme_Check.php
swissspidy: I don't think the version-check API returns beta/RC versions.

Do you mean short-circuiting the HTTP request so we can verify the behavior of the `get_bloginfo` fallback?

mukeshpanchal27: Yes

ernilambar:
> * Add unit test that ignore all these warnings. I think we can archive it using `add_filter( 'wp_plugin_check_ignored_readme_warnings', '__return_empty_array' );`.

Its the opposite. Sending empty array in that filter will disable ignoring and all warnings are displayed. 
We already have `test_filter_wp_plugin_check_ignored_readme_warnings_will_return_no_error` to test same functionality. Do we need another similar one?

includes/Checker/Checks/Plugin_Readme_Check.php
ernilambar: How can we disable HTTP request inside unit test? Suggestion please.

ernilambar: Test added.

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -242,6 +242,98 @@ public function test_run_with_errors_parser_warnings() {
...
+
+		remove_filter(
+			'pre_http_request',
+			static function () {
+				return new WP_Error( 'http_request_blocked', __( 'This request is not allowed', 'plugin-check' ) );
+			}
+		);
+	}
+
+	public function test_run_with_errors_multiple_parser_warnings_and_empty_ignored_array() {
```
mukeshpanchal27, line 299: Could you please add comment why filter return one extra warning? As `test_run_with_errors_multiple_parser_warnings` return 6 warnings. Make sense?

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -242,6 +242,103 @@ public function test_run_with_errors_parser_warnings() {
...
+		$this->assertArrayHasKey( 0, $warnings['readme.txt'] );
+		$this->assertArrayHasKey( 0, $warnings['readme.txt'][0] );
+
+		for ( $i = 0; $i < 6; ++$i ) {
+			$this->assertArrayHasKey( 'code', $warnings['readme.txt'][0][0][ $i ] );
+			$this->assertEquals( 'readme_parser_warnings', $warnings['readme.txt'][0][0][ $i ]['code'] );
+		}
+	}
+
+	public function test_run_with_errors_parser_warnings_and_http_disabled() {
```
mukeshpanchal27: The `get_wordpress_stable_version` code needs adjustments if we want to test the RCs and Beta versions. We can handle that in follow-up enhancement work, as it's not a big deal.

Instead, let's remove this test, as it doesn't do anything different. In its place, we can add the following test.

```
public function test_run_with_errors_parser_warnings_with_custom_set_transient_version() {

	$version = '5.0';
	set_transient( 'wp_plugin_check_latest_wp_version', $version );

	$readme_check  = new Plugin_Readme_Check();
	$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-plugin-readme-parser-warnings/load.php' );
	$check_result  = new Check_Result( $check_context );

	$readme_check->run( $check_result );

	$errors   = $check_result->get_errors();
	$warnings = $check_result->get_warnings();

	$this->assertNotEmpty( $warnings );
	$this->assertEquals( 0, $check_result->get_error_count() );
	$this->assertEquals( 1, $check_result->get_warning_count() );

	$this->assertArrayHasKey( 0, $warnings['readme.txt'] );
	$this->assertArrayHasKey( 0, $warnings['readme.txt'][0] );
	$this->assertArrayHasKey( 'code', $warnings['readme.txt'][0][0][0] );
	$this->assertEquals( 'readme_parser_warnings', $warnings['readme.txt'][0][0][0]['code'] );
	$this->assertStringContainsString( 'The "Tested up to" field was ignored. This field should only contain a valid WordPress version such as "' . $version . '"', $warnings['readme.txt'][0][0][0]['message'] );

	delete_transient( 'wp_plugin_check_latest_wp_version' );
}
```

This test checks the custom version set up by the transient.

mukeshpanchal27 (approved):

PR #399, opened 2024-01-23
Detect localhost/127.0.0.1 code in multiple files

### LINKED ISSUE #362: Localhost check does not show errors in multiple files
I tested with keeping `http://localhost:8888` in two PHP files. And in the check, error is shown for only file.

`/includes/Checker/Checks/Localhost_Check.php`

Line 48:  

    $file = self::file_preg_match( '#https?://(localhost|127.0.0.1)#', $php_files );

This seems to be happening because `self::file_preg_match()` only returns the first match.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/362

* Adds method `files_preg_match` in `Abstract_File_Check` to return multiple results matched with given regex.
* Update `Localhost_Check` to handle array of results
* Update `Localhost_Check_Tests` accordingly

### DIFF
--- a/includes/Checker/Checks/Abstract_File_Check.php
+++ b/includes/Checker/Checks/Abstract_File_Check.php
@@ -144,7 +144,7 @@ abstract class Abstract_File_Check implements Static_Check {
 	/**
 	 * Returns matched files performing a regular expression match on the file contents of the given list of files.
 	 *
-	 * @since 1.0.2
+	 * @since 1.1.0
 	 *
 	 * @param string $pattern The pattern to search for.
 	 * @param array  $files   List of absolute file paths.
@@ -169,7 +169,7 @@ abstract class Abstract_File_Check implements Static_Check {
 	/**
 	 * Returns matched files performing a regular expression match on the file contents of the given list of files with line and column information.
 	 *
-	 * @since 1.0.2
+	 * @since 1.1.0
 	 *
 	 * @param string $pattern The pattern to search for.
 	 * @param array  $files   List of absolute file paths.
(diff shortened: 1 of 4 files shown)


ernilambar:

swissspidy:
Let's discuss the right approach on the issue first. This applies to other checks as well.

frantorres:
Thanks for this change, I think that for this check it would be really useful to also give users the line + column where that happens (as well as showing more than one case for file). I mentioned this on #362 .

Also, I can suggest to add other common "localhost" cases like: 
- `https?:\/\/(?:[^\s\/]*\.)++local(?:host)?` (this covers domains ending in .local and .localhost, for example: docker.local)

ernilambar:
PR has been updated based on feedback.

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -141,6 +141,69 @@ final protected static function file_preg_match( $pattern, array $files, array &
 		return false;
 	}
 
+	/**
+	 * Returns matched files performing a regular expression match on the file contents of the given list of files.
+	 *
+	 * @since 1.0.2
```
swissspidy:
```suggestion
	 * @since 1.1.0
```

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -141,6 +141,69 @@ final protected static function file_preg_match( $pattern, array $files, array &
...
+			}
+		}
+
+		return count( $matched_files ) > 0 ? $matched_files : false;
+	}
+
+	/**
+	 * Returns matched files performing a regular expression match on the file contents of the given list of files with line and column information.
+	 *
+	 * @since 1.0.2
```
swissspidy:
```suggestion
	 * @since 1.1.0
```

swissspidy (approved):

davidperezgar (approved):

PR #401, opened 2024-01-24
Make plugin_repo checked by default in categories

### LINKED ISSUE #400: Make only Plugin Repo as selected by default
Follow up from https://github.com/WordPress/plugin-check/issues/283#issuecomment-1904286396

> With that my only ask is that the "Plugin Repo" should be the only one that comes checked by default to avoid confusion from new plugin developers.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/400

* `plugin_repo` is checked by default in categories in admin page

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -196,9 +196,9 @@ final class Admin_Page {
 
 		$category_labels = Check_Categories::get_category_labels();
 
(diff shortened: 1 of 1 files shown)


mukeshpanchal27 (approved):

swissspidy (approved):

bordoni:
Thank you all so much!

PR #403, opened 2024-01-25
Add check for missing Stable Tag in readme

ernilambar:
* Update `function check_stable_tag()` to test if `Stable Tag` is missing
* Raise ERROR for missing stable tag
* Update tests to reflect new changes

Since the value of `Stable Tag` is very important in our directory, I have kept this check as ERROR. We can keep it as WARNING if otherwise. This check is currently not there in legacy version.

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -172,6 +172,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 				'no_stable_tag',
 				$readme_file
 			);
+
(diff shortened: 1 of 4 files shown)


swissspidy:
@foosantos @bordoni Is such a check for missing Stable Tag useful for the plugin review team?

felipevelzani:
Yes, that is indeed one of the things that is checked manually, so surely relevant. Thank you, @ernilambar and @swissspidy!

swissspidy (approved):

mukeshpanchal27 (approved):

PR #406, opened 2024-01-29
Add Plugin_Header_Requires_Check

### LINKED ISSUE #405: Requires related plugin header check in readme
Related: https://github.com/WordPress/plugin-check/issues/198

For following two headers, we accept fields either in readme file or main plugin file.

- `Requires at least`
- `Requires PHP`

We could raise **ERROR** or **WARNING** if those fields are missing from both files.

It seems preference is given to plugin file header. May be we could create separate Check instead of keeping inside `plugin_readme` check.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/405

* Add `Plugin_Header_Requires_Check`
* Tests added for `Plugin_Header_Requires_Check`
* For single file plugin or plugin without readme file, headers of main plugin is checked. ERROR is raised if no headers found
* For multi file plugin, ERROR is raised if headers are missing from both main plugin file or readme file


ernilambar:
cc: @swissspidy @mukeshpanchal27 @felixarntz

swissspidy:
Let's discuss on the ticket first.

PS Felix is on leave right now, best not to ping him :)

ernilambar:
Closing this for now. We can reopen in the future if needed.

PR #408, opened 2024-01-30
Feature to exclude files in the check

ernilambar:
Fix partially https://github.com/WordPress/plugin-check/issues/388

* `--exclude-files` argument added in `plugin check` command
* Filter `wp_plugin_check_ignore_files` added for ignoring files
* Added Behat tests and unit tests


ernilambar:

includes/Checker/Checks/Abstract_File_Check.php
```diff
@@ -227,6 +227,15 @@ private static function get_files( Check_Context $plugin ) {
 					}
 				}
 
+				$files_to_ignore = Plugin_Request_Utility::get_files_to_ignore();
+
+				foreach ( $files_to_ignore as $ignore_file ) {
+					if ( str_ends_with( $file, $ignore_file ) ) {
```
swissspidy:
```suggestion
					if ( str_ends_with( $file, "/$ignore_file" ) ) {
```

Otherwise `--exclude-files=bar.php` will also exclude `foobar.php`, from what I can see here. That should not happen.

Adding a test for such a scenario would be good.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -165,9 +165,23 @@ private function get_argv_defaults( Check_Result $result ): array {
...
+
+		// Ignore directories.
 		if ( ! empty( $directories_to_ignore ) ) {
-			$defaults[] = '--ignore=*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
+			$ignore_patterns[] = '*/' . implode( '/*,*/', $directories_to_ignore ) . '/*';
+		}
+
+		// Ignore files.
+		if ( ! empty( $files_to_ignore ) ) {
+			$ignore_patterns[] = implode( ',', $files_to_ignore );
```
swissspidy: Do we need to use a pattern here as well to prevent the same `foobar.php` / `bar.php` partial match scenario?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -95,9 +95,12 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * : Include experimental checks.
 	 *
 	 * [--exclude-directories=<directories>]
-	 * : Additional directories to exclude from checks
+	 * : Additional directories to exclude from checks.
 	 * By default, `.git`, `vendor` and `node_modules` directories are excluded.
 	 *
+	 * [--exclude-files=<files>]
+	 * : Additional files to exclude from checks.
+	 *
```
swissspidy, line 103: Not to self: at some point in the future we should just make this one `--exclude` arg for better DX.

swissspidy (approved):

ernilambar:
@mukeshpanchal27 Merge conflict has been fixed.

ernilambar:
@mukeshpanchal27 PR is synced with trunk and `n.e.x.t` has been removed as version.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -161,4 +161,24 @@ public static function get_directories_to_ignore() {
...
+		/**
+		 * Filters the files to ignore.
+		 *
+		 * @since 1.0.0
+		 *
+		 * @param array $default_ignore_files An array of files to ignore.
+		 */
+		$files_to_ignore = (array) apply_filters( 'wp_plugin_check_ignore_files', $default_ignore_files );
+
+		return $files_to_ignore;
```
mukeshpanchal27, line 182:
```suggestion
		/**
		 * Filters the files to ignore.
		 *
		 * @since 1.0.0
		 *
		 * @param array $default_ignore_files An array of files to ignore. Default empty array.
		 */
		return (array) apply_filters( 'wp_plugin_check_ignore_files', array());
```

ernilambar: `@param` without variable name will trigger phpstan error. To accommodate above change you mentioned, when I remove $default_ignore_files variable, there will be error. So I have followed same approach like we have done in `get_directories_to_ignore()` method.

mukeshpanchal27:
@swissspidy Do you know why the Behat test failed?

swissspidy:
As the error message says, the codecov token is missing, because this PR is originating from a fork. The tests are passing.

davidperezgar (approved):
It's very nice this enhacement that will solve false positives.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -161,4 +161,24 @@ public static function get_directories_to_ignore() {
 
 		return $directories_to_ignore;
 	}
+
+	/**
+	 * Gets the files to ignore using the filter.
+	 *
+	 * @since 1.0.0
```
swissspidy, line 168: Note that this is incorrect now

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -161,4 +161,24 @@ public static function get_directories_to_ignore() {
...
+	 */
+	public static function get_files_to_ignore() {
+		$default_ignore_files = array();
+
+		/**
+		 * Filters the files to ignore.
+		 *
+		 * @since 1.0.0
```
swissspidy, line 176: Note that this is incorrect now

includes/Utilities/Plugin_Request_Utility.php
ernilambar: Should it be `1.0.2` or `1.1.0`? What will be our next release?

mukeshpanchal27: As we set `1.0.2` for the most recent merged changes, let's stick with that version for these changes as well.

PR #409, opened 2024-01-31
Fix failing Behat test

ernilambar:
Ref: https://github.com/WordPress/plugin-check/actions/runs/7728228566/job/21068526460

Behat test is failing because Akismet has done some update which we have used in the CLI test.

### DIFF
--- a/tests/behat/features/plugin-check.feature
+++ b/tests/behat/features/plugin-check.feature
@@ -100,11 +100,11 @@ Feature: Test that the WP-CLI command works.
     When I run the WP-CLI command `plugin check akismet`
     Then STDOUT should contain:
       """
-      FILE: views/config.php
+      FILE: views/notice.php
       """
 
     When I run the WP-CLI command `plugin check akismet --exclude-directories=views`
     Then STDOUT should not contain:
       """
-      FILE: views/config.php
+      FILE: views/notice.php
       """
--- a/tests/behat/features/plugin-check.feature
+++ b/tests/behat/features/plugin-check.feature
@@ -94,17 +94,42 @@ Feature: Test that the WP-CLI command works.
       WordPress.Security.EscapeOutput.OutputNotEscaped
       """
 
-  Scenario: Check Akismet
+  Scenario: Exclude directories in plugin check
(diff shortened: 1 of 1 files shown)


ernilambar:

Note: We need to explore option for Behat tests without depending on external plugin.

swissspidy:
That's exactly why I wasn't too happy with using external plugins for tests like this.

The most robust solution would be to use dedicated test plugins, like we do for unit tests already.

Something like this (untested):

```gherkin

  Scenario: Exclude directories
    Given a WP install with the Plugin Check plugin
    And an empty wp-content/plugins/foo-plugin directory
    And an empty wp-content/plugins/foo-plugin/subdir directory
    And a wp-content/plugins/foo-plugin/foo-plugin.php file:
      """
      <?php
      /**
       * Plugin Name: Foo Plugin
       * Plugin URI:  https://example.com
       * Description:
       * Version:     0.1.0
       * Author:
       * Author URI:
       * License:     GPL-2.0+
       * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
       * Text Domain: foo-plugin
       * Domain Path: /languages
       */

      """
    And a wp-content/plugins/foo-plugin/subdir/test.php file:
      """
      <?php
      echo 1234;
      """
    
    When I run the WP-CLI command `plugin check foo-plugin`
    Then STDOUT should contain:
      """
      FILE: subdir/test.php
      """

    When I run the WP-CLI command `plugin check foo-plugin --exclude-directories=subdir`
    Then STDOUT should not contain:
      """
      FILE: subdir/test.php
      """
```

swissspidy (approved):

mukeshpanchal27:
Thanks @ernilambar for the PR. The changes LGTM.

What about `Hello Dolly` plugin? At some point that will also updated?

ernilambar:
@mukeshpanchal27 May be we should replace that also. We can tackle that in separate PR.

mukeshpanchal27:
> @mukeshpanchal27 May be we should replace that also. We can tackle that in separate PR.

Make sense.

mukeshpanchal27 (approved):

PR #410, opened 2024-02-01
Avoid plugin header text domain check for single file plugin

ernilambar:
This `Plugin_Header_Text_Domain_Check` is not applicable to single file plugin because logic of finding plugin slug here will always return `plugins` as plugin slug value. Since we generally don't accept single file plugin in the directory and also message here is of `WARNING` type, bailing this check would be desirable I believe.


ernilambar:

swissspidy (approved):

mukeshpanchal27 (approved):

PR #413, opened 2024-02-05
Fix edit file URL for single file plugin

### LINKED ISSUE #412: Incorrect edit link in single file plugin
Edit Link `View in code editor` is not correct for single file plugin. Clicking in the link takes to an inaccessible page with error `Sorry, that file cannot be edited.`

Eg:

```
/plugin-editor.php?plugin=hello.php&file=plugins%2Fhello.php&line=49
```

`file` parameter is not supposed to have `plugins` here. It should be `hello.php` only.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/412

- Value of `file` is generated conditionally

### DIFF
--- a/includes/Traits/File_Editor_URL.php
+++ b/includes/Traits/File_Editor_URL.php
@@ -101,7 +101,7 @@ trait File_Editor_URL {
 		if ( ! $edit_url && current_user_can( 'edit_plugins' ) ) {
 			$query_args = array(
 				'plugin' => rawurlencode( $result->plugin()->basename() ),
-				'file'   => rawurlencode( str_contains( $result->plugin()->basename(), '/' ) ? $plugin_slug . '/' . $filename : $filename ),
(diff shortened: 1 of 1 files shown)


ernilambar:

includes/Traits/File_Editor_URL.php
```diff
@@ -101,7 +101,7 @@ protected function get_file_editor_url( Check_Result $result, $filename, $line =
 		if ( ! $edit_url && current_user_can( 'edit_plugins' ) ) {
 			$query_args = array(
 				'plugin' => rawurlencode( $result->plugin()->basename() ),
-				'file'   => rawurlencode( $plugin_slug . '/' . $filename ),
+				'file'   => rawurlencode( str_contains( $result->plugin()->basename(), '/' ) ? $plugin_slug . '/' . $filename : $filename ),
```
swissspidy: Should we perhaps use `$result->plugin()->is_single_file_plugin()` instead?

swissspidy (approved):

mukeshpanchal27 (approved):
Nice catch.

PR #414, opened 2024-02-05
Detect missing readme headers

ernilambar:
Related to https://github.com/WordPress/plugin-check/issues/402

* Added `check_headers()` method in `Plugin_Readme_Check`
* Add separate private method for getting ignored filters `get_ignored_warnings()`
* Two headers are checked here:
  - `Tested up to`
  - `Contributors`
* New method will check if those fields are empty
* Checking for the field is enabled/disabled based on `wp_plugin_check_ignored_readme_warnings`
* Since `contributor_ignored` is returned from the above filter, detecting of `Contributors` is disabled.

If the PR is acceptable then I will add tests for the changes and update other tests accordingly. Other tests also need to be updated because currently we have assumed there will be no warning for missing `Tested up to` field.


ernilambar:

frantorres:
Hi @ernilambar thanks for the change. It looks good, if you finish the tests I can continue with the review.

ernilambar:
PR has been rebased with trunk and few more updates have been done.

davidperezgar:
For me It's ok after all changes! In the future, we would need to get the actual version of WordPress somewhere and create a warning if it's not getting an actual version.

davidperezgar (approved):
Everything seems OK!

davidperezgar:
I would add to check the actual version of WordPress. It must be Tested up "actual version". 

In order to have always the latest version, I've [asked to meta](https://wordpress.slack.com/archives/C02QB8GMM/p1719654997418999), and we have some methods to check that:

```
Yes, use one of these:
https://api.wordpress.org/core/stable-check/1.0/
3rd line of https://api.wordpress.org/core/version-check/1.1/
Or just json decode https://api.wordpress.org/core/version-check/1.7/
```

So finally we have a complete check.

The error that we show to the users is something like this `ERROR: Tested up to: 6.4 < 6.5`

swissspidy (approved):

ernilambar:
@davidperezgar Should this be done in this same PR or separate PR?

davidperezgar:
In a separate Issue. I'll create the issue #481

davidperezgar (commented):
Ok

PR #417, opened 2024-02-06
Improve readme plugin name header check

ernilambar:
Few changes are there regarding plugin name header in this update https://meta.trac.wordpress.org/changeset/13162

* Whenever `invalid_plugin_name_header` warning is triggered, `name` property is `false`.
* This warning is triggered in two cases: when there is no valid plugin name header (and there are other content to be parsed in the readme file) AND when header is `=== Plugin Name ===`. So I have updated the message considering both cases.
* When there is no warning `invalid_plugin_name_header` but `$parser->name` is still `empty()` then it means readme file is completely empty.

### DIFF
--- a/tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
+++ b/tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
@@ -67,8 +67,6 @@ class Plugin_Readme_Check_Tests extends WP_UnitTestCase {
 
 		$this->assertNotEmpty( $errors );
 		$this->assertArrayHasKey( 'readme.txt', $errors );
-		// Two errors are 'empty_plugin_name' and 'no_license'.
-		$this->assertEquals( 2, $check_result->get_error_count() );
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -67,8 +67,6 @@ public function test_run_with_errors_empty_name() {
 
 		$this->assertNotEmpty( $errors );
 		$this->assertArrayHasKey( 'readme.txt', $errors );
-		// Two errors are 'empty_plugin_name' and 'no_license'.
```
ernilambar, line 70: @swissspidy I have removed this to fix the PHPUnit error. We are gradually adding new readme checks in the plugin. Every time we add new checks this number does not match. Since our target here is to check for empty plugin name which is below this code, test should be good for this I guess.

ernilambar:

swissspidy (approved):

mukeshpanchal27 (approved):

PR #419, opened 2024-02-07
Rename main plugin file name to match v0.2.3

swissspidy:
If we change the main plugin filename, the plugin would get deactivated when updating it.


westonruter (approved):

ernilambar:
For me, I would not rename it and keep `plugin-check/plugin-check.php` as it is. This is the standard practices we are spreading about plugin development best practices in our handbook and other documents. I would prefer sticking to the best practices what we are advising for other developers.
Since this plugin is targeted for developers and also active installations is also not that high, and also major versions are generally bound to have breaking changes, may be we can keep appropriate Upgrade Notice in readme and keep files as it is. Just my 2 cents.

joemcgill:
I tend to agree with @ernilambar. What if we just added a `plugin.php` file that updated the value in the DB and then loaded the original `plugin-check.php` file? That should keep back-compat in place and eventually get everyone on the correct file.

swissspidy:
The thing with a compat file is that you have to keep it forever, which is a bit unnecessary. It's just a filename, it doesn't really affect anything, and keeping the existing one doesn't hurt anyone.

I just had a quick chat with @bordoni about this and he also prefers not having to deactivate plugins either. Especially since it's already active on WordPress.org as well.

So this simple renaming is the most straightforward way to address this issue.

joemcgill (approved):

bordoni (approved):

mukeshpanchal27 (approved):

PR #428, opened 2024-02-14
Do not send email after “install”

### LINKED ISSUE #424: New user E-mail sent out on accessibility or/and performance test
### Description
When checking accessibility or/and performance test a email is sent out with admin credentials for new site. (System failed to delivery e-mail)

https://wordpress.slack.com/archives/C1LBM36LC/p1707869533873209

### Environment
- WordPress: 6.5-beta1
- PHP: 8.3.2
- Server: LiteSpeed
- Database: mysqli (Server: 10.6.16-MariaDB / Client: mysqlnd 8.3.2)
- Browser: Safari 17.3.1 (Mobile) (iPhone)
- Theme: GeneratePress 3.4.0
- MU-Plugins: 
  * block-automation-by-installatron.php 
  * Health Check Troubleshooting Mode 1.9.0
  * ManageWP - Worker Loader 1.0.0
- Plugins: 
  * Core Rollback 1.3.5
  * Enable jQuery Migrate Helper 1.4.0
  * Health Check & Troubleshooting 1.7.0
  * Limit Login Attempts Reloaded 2.26.4
  * Plugin Check 1.0.0
  * Plugin Compatibility Checker 3.0.1
  * Query Monitor 3.15.0
  * WordPress Beta Tester 3.5.5
  * WP-Cron Status Checker 1.2.5
  * WP Crontrol 1.16.1
  * WP Mail Logging 1.12.0
  * WP SMS 6.5.4
  * WPVulnerability 3.1.1

### Steps to Reproduce
1.  Install and activate PPC version 1.0.0 
2.  Tick accessibility or/and performance test and run check on any plugin
3. 🐞 Bug occurs. 

### Expected Results
1.  ✅  Disable the e-mail sending. Or does a new user need to be created for the test? 

### Actual Results
1.  ❌ System Mail Delivery e-mail sent out on every check - Mail delivery failed: returning message to sender 
 demo@plugincheck.test
   Unrouteable address
   
With a new user (administrator) and password for “ New WordPress Site”

swissspidy:
Fixes #424

### DIFF
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -48,12 +48,17 @@ final class Runtime_Environment_Setup {
 				}
 			);
 
+			// Do not send post-install notification email, see https://github.com/WordPress/plugin-check/issues/424.
+			add_filter( 'pre_wp_mail', '__return_false' );
+
 			wp_install(
 				'Plugin Check',
 				'plugincheck',
 				'demo@plugincheck.test',
 				false
 			);
+
+			remove_filter( 'pre_wp_mail', '__return_false' );
 
 			// Activate the same plugins in the test environment.
 			update_option( 'active_plugins', $active_plugins );


includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -48,13 +48,18 @@ static function () use ( $permalink_structure ) {
...
+			add_filter( 'pre_wp_mail', '__return_false' );
+
 			wp_install(
 				'Plugin Check',
 				'plugincheck',
 				'demo@plugincheck.test',
 				false
 			);
 
+			remove_filter( 'pre_wp_mail', '__return_false' );
```
westonruter, line 61: Would this be safer in case there is an existing plugin that is already disabling mail?

```suggestion
			$return_false = static function (): bool {
				return false;
			};
			add_filter( 'pre_wp_mail', $return_false );

			wp_install(
				'Plugin Check',
				'plugincheck',
				'demo@plugincheck.test',
				false
			);

			remove_filter( 'pre_wp_mail', $return_false );
```

westonruter (approved):
One question, but otherwise looks good.

includes/Checker/Runtime_Environment_Setup.php
swissspidy: Since this runs only for the duration of a runtime check, not sure if there's even any risk associated with that. So while I understand the point, it seems a bit too defensive

bordoni (approved):

PR #435, opened 2024-03-07
Fix readme file issue in Windows

### LINKED ISSUE #434: Windows: Plugin check reports that readme.txt file is missing. Edit Link fails for other files.
When running plugin check in a Windows environment the plugin check doesn't find the readme.txt file.

It's probably a forward slash / backslash problem in trait Find_Readme()

In my example:

`$plugin_relative_path` is `C:\apache\htdocs\wordpress/wp-content/plugins/field-block-for-acf-pro/`
but the $files array has a backslash after the plugin name.

```
[160] => (string) "C:\apache\htdocs\wordpress/wp-content/plugins/field-block-for-acf-pro\README.md"
 [161] => (string) "C:\apache\htdocs\wordpress/wp-content/plugins/field-block-for-acf-pro\readme.txt"
```

See also https://github.com/WordPress/plugin-check/issues/271

### Scenario
- Tools > Plugin Check
- Choose any plugin
- Categories: Plugin Report
- Click on Check it!

The Warning `no_plugin_readme` with message "The plugin readme.txt does not exist". 

The Edit Link works for this file.

### Environment
Windows: 11
WordPress: 6.4.3
PHP: 8.3
plugin-check version: 1.0.1

### Additional observations

The Edit Link doesn't work for the other messages. The path is incorrect.

https://s.b/wordpress/wp-admin/plugin-editor.php?plugin=field-block-for-acf-pro%2Ffield-block-for-acf-pro.php&file=field-block-for-acf-pro%2FC%3A%5Capache%5Chtdocs%5Cwordpress%2Fwp-content%2Fplugins%2Ffield-block-for-acf-pro%5Cphpunit.xml.dist

https://s.b/wordpress/wp-admin/plugin-editor.php?plugin=field-block-for-acf-pro%2Ffield-block-for-acf-pro.php&file=field-block-for-acf-pro%2FC%3A%5Capache%5Chtdocs%5Cwordpress%2Fwp-content%2Fplugins%2Ffield-block-for-acf-pro%5C.git

image

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/434

### DIFF
--- a/includes/Traits/Find_Readme.php
+++ b/includes/Traits/Find_Readme.php
@@ -27,12 +27,15 @@ trait Find_Readme {
 		// Find the readme file.
 		$readme_list = preg_grep( '/readme\.(txt|md)$/i', $files );
 
+		$plugin_relative_path = rtrim( $plugin_relative_path, '/' );
+
 		// Filter the readme files located at root.
 		$potential_readme_files = array_filter(
 			$readme_list,
 			function ( $file ) use ( $plugin_relative_path ) {
 				$file = str_replace( $plugin_relative_path, '', $file );
-				return ! str_contains( $file, '/' );
+				$file = ltrim( $file, '/\\' );
+				return ! ( str_contains( $file, '/' ) || str_contains( $file, '\\' ) );
 			}
 		);
 
--- a/readme.txt
+++ b/readme.txt
@@ -69,6 +69,10 @@ In any case, passing the checks in this tool likely helps to achieve a smooth pl
 
 == Changelog ==
 
+= 1.0.2 =
+
+* Fix - Correct detection of readme files in Windows
+
 = 1.0.1 =
 
 * Fix - Add missing `test-content` folder needed for runtime checks.


bordoni:
@ernilambar can we make sure to add a Changelog here?

includes/Traits/Find_Readme.php
```diff
@@ -27,12 +27,15 @@ protected function filter_files_for_readme( array $files, $plugin_relative_path
 		// Find the readme file.
 		$readme_list = preg_grep( '/readme\.(txt|md)$/i', $files );
 
+		$plugin_relative_path = rtrim( $plugin_relative_path, '/' );
```
felixarntz, line 30: Why is this needed? In which situation would `$plugin_relative_path` end in a forward slash?

includes/Traits/Find_Readme.php
```diff
@@ -27,12 +27,15 @@ protected function filter_files_for_readme( array $files, $plugin_relative_path
...
+
 		// Filter the readme files located at root.
 		$potential_readme_files = array_filter(
 			$readme_list,
 			function ( $file ) use ( $plugin_relative_path ) {
 				$file = str_replace( $plugin_relative_path, '', $file );
-				return ! str_contains( $file, '/' );
+				$file = ltrim( $file, '/\\' );
+				return ! ( str_contains( $file, '/' ) || str_contains( $file, '\\' ) );
```
felixarntz, line 38: Could be simplified to:
```suggestion
				return ! str_contains( $file, '/' ) && ! str_contains( $file, '\\' );
```

felixarntz (commented):
@ernilambar Thank you for the PR!

While this PR addresses the concrete problem you're facing, I wonder whether we should take a step back to find a more holistic solution to the problem.

Paths being different on Windows can affect _any_ checks, and I wonder whether we should cater for that in a more central location, rather than having to adjust any such checks.

WordPress provides a [`wp_normalize_path()` function](https://developer.wordpress.org/reference/functions/wp_normalize_path/) which takes care of normalizing paths so that they always use `/`.

Maybe we should use this function in the `Abstract_File_Check::get_files()` method (and potentially other methods in that class) so that all paths are always normalized? This way we can keep using the common `/` everywhere, no need for workarounds like this one.

Curious to hear other people's thoughts. cc @swissspidy @mukeshpanchal27

ernilambar:
Closing.

PR #449, opened 2024-04-01
Use normalized path

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/434

* Use normalized path when fetching list files and folders

PR #453, opened 2024-04-08
Fix fatal error introduced in #449

swissspidy:
Introduced in #449 

Reported in https://github.com/swissspidy/wp-plugin-check-action/issues/68

To reproduce:

```bash
wp plugin check akismet --require=./wp-content/plugins/plugin-check/cli.php
```

cc @ernilambar


felixarntz (commented):
Thanks @swissspidy. Overall this looks good to me, though I'm not sure I like the idea of the caller of the `Plugin_Context` class constructor having to take care of the normalization.

On the one hand, it makes sense from a separation of concerns perspective, because it works differently for WP core and WP-CLI. On the other hand, it would be great to implement this internally in the class.

Maybe we could go with a pragmatic approach within the `Plugin_Context` constructor? If `wp_normalize_path()` exists, use that, otherwise if we're in CLI and the WP-CLI method exists, use that, otherwise do nothing - or maybe even throw an exception cause this should never happen? WDYT?

swissspidy:
IDK, don't have a strong opinion. We can keep doing it in the constructor, but throwing an exception seems overkill.

includes/Plugin_Context.php
```diff
@@ -42,7 +43,13 @@ class Plugin_Context {
 	 * @param string $main_file The absolute path to the plugin main file.
 	 */
 	public function __construct( $main_file ) {
-		$this->main_file = wp_normalize_path( $main_file );
+		if ( function_exists( '\WP_CLI\Utils\normalize_path' ) ) {
+			$this->main_file = normalize_path( $main_file );
+		} elseif ( function_exists( 'wp_normalize_path' ) ) {
```
felixarntz: Nit-pick: I think the order should be reversed since WP core is more "canonical" than WP-CLI :)

includes/Plugin_Context.php
```diff
@@ -42,7 +43,13 @@ class Plugin_Context {
...
+			$this->main_file = wp_normalize_path( $main_file );
+		} else {
+			$this->main_file = $main_file;
```
felixarntz: Following up on my previous feedback, I do think it's better to throw an exception or at least trigger a PHP warning here because it will result in problems when using Windows if this logic is ever reached.

It should never be reached in the first place, which is why I think even an exception here is fine. But not doing anything and silently let it pass will lead to obscure errors, should anyone ever run into it.

felixarntz (approved):
The current implementation WFM, though I do think being strict here is safer than silently letting a problematic configuration continue. Not a big deal either way, so my feedback is just nit-picks.

Note that there are some lint failures due to the CLI function.

includes/Plugin_Context.php
```diff
@@ -39,10 +41,18 @@ class Plugin_Context {
...
 	 * @param string $main_file The absolute path to the plugin main file.
 	 */
 	public function __construct( $main_file ) {
-		$this->main_file = wp_normalize_path( $main_file );
+		if ( function_exists( 'wp_normalize_path' ) ) {
+			$this->main_file = wp_normalize_path( $main_file );
+		} elseif ( function_exists( '\WP_CLI\Utils\normalize_path' ) ) {
+			$this->main_file = normalize_path( $main_file );
+		} else {
+			throw new Exception( 'Unknown environment, normalize_path function not found' );
```
```suggestion
			throw new Exception(
				__( 'Unknown environment, normalize_path function not found', 'plugin-check' )
			);
```

includes/Plugin_Context.php
```diff
@@ -39,10 +41,18 @@ class Plugin_Context {
 	 *
 	 * @since 1.0.0
 	 *
+	 * @throws Exception Throws exception if not called via regular WP-CLI or WordPress bootstrap order.
+	 *
 	 * @param string $main_file The absolute path to the plugin main file.
```
```suggestion
	 * @param string $main_file The absolute path to the plugin main file.
	 *
	 * @throws Exception Throws exception if not called via regular WP-CLI or WordPress bootstrap order.
```

Re-order for consistency

tests/behat/features/plugin-check.feature
```diff
@@ -155,3 +155,34 @@ Feature: Test that the WP-CLI command works.
...
+          echo $number;
+        }
+      );
+      """
+
+    When I run the WP-CLI command `plugin check foo-single.php --require=./wp-content/plugins/plugin-check/cli.php`
+    Then STDOUT should contain:
+      """
+      mt_rand() is discouraged.
+      """
```
mukeshpanchal27:
```suggestion
      """

```
Missing new line at end.

includes/Plugin_Context.php
```diff
@@ -6,8 +6,10 @@
  */
 
 namespace WordPress\Plugin_Check;
+use Exception;
```
```suggestion
namespace WordPress\Plugin_Check;

use Exception;
```

nit-pick

mukeshpanchal27 (commented):
Thanks @swissspidy for the PR! Left some feedbacks.

ernilambar:
What if we provide polyfill function for `wp_normalize_path()` so that it is always available.

swissspidy:
I think a polyfill for an edge case that should never ever happen is overkill.

Here it‘s only not available because of the hacky object-cache.php drop-in

mukeshpanchal27 (approved):

PR #457, opened 2024-04-23
Bump PHP requirement to 7.2+

### LINKED ISSUE #455: Bump the minimum PHP version to 7.2
As WordPress bumps the [minimum required PHP version to 7.2](https://core.trac.wordpress.org/ticket/58719) for 6.6 version, we should do the same plugin.

swissspidy:
7.2.24 is the same requirement as WP 6.6+

Had to pin afragen/wordpress-plugin-readme-parser because the latest version requires 7.4

Fixes #455


readme.txt
barrykooij, line 1: @swissspidy Did you intend to remove ```Requires PHP:```  from the readme.txt?

swissspidy: @barrykooij Yes. This header should only be in the main plugin file, not in the readme. See https://meta.trac.wordpress.org/changeset/9050

barrykooij (approved):

Issue #448, opened 2024-03-28
Decision making and governance for included checks


felixarntz:
In this issue, I'd like for us to brainstorm and discuss a bit on how we would like this plugin to evolve in the future, in particular regarding additional checks to be added.

There probably needs to be _some_ sort of process, especially as more people may contribute to the project in the future (hopefully!). We certainly shouldn't overcomplicate this, but we may need more than just "any maintainer approving any PR including a new check" as a "decision" to include a new check.

While I'm sure none of us has any malicious intentions, the purpose of this discussion is to find a way to ensure the "integrity" of the plugin checker long-term. We probably want to avoid that, for example, an extremely opinionated check would make it into the plugin checker and then cause confusing errors for tons of plugin developers.

I'm curious to get your perspective on this. Some questions / ideas to discuss:
* Should there be specific `CODEOWNERS` whose approval is required to merge a new check?
* Should PRs that add a new check require having a "check proposal" issue that was formally approved, in some way?
* Should there be specific people or groups assigned to "own" specific "check categories"? (For instance, ownership between different verticals like plugin review, performance, accessibility etc. could differ, based on expertise and involvement.)
* Should there be additional requirements for new checks, such as having a documentation URL attached about how to fix the respective problem?
* Or maybe, is the simple approach of any maintainer approving a PR with a new check sufficient?

cc @bordoni @barrykooij @swissspidy @mukeshpanchal27

barrykooij:
Hey @felixarntz,

I think this would be the most suitable:
> Should there be specific people or groups assigned to "own" specific "check categories"? (For instance, ownership between different verticals like plugin review, performance, accessibility etc. could differ, based on expertise and involvement.)

I can only speak from the plugin review part of course, but for us it would be ideal if we can change and merge our own checks. These checks will reflect our guidelines, which we discuss within the team and publicly already. Since checks in PCP only reflect our guidelines, having discussion here regarding adding them makes no sense to me.

If we need to change structural things that affect other parts of the plugin, we should do this via issues that all related parties can discuss in, in my opinion.

This all is just my personal opinion of course, happy to discuss this further.

chriscct7:
I strongly agree with @barrykooij

I think what makes most sense is:
- Performance checks: merged only by Performance team
- Security checks: merged only by Plugins Team Security members (or users that team designates as needed) as these are currently enforced on plugin re-reviews manually currently by that team
[...]

PR #460, opened 2024-04-29
Add initial code owners configuration

swissspidy:
Moving checks into separate folders per category makes it easy to set up code owners.

Feel free to submit suggestions to the CODEOWNERS file if you like to be included/excluded.

See #448


joemcgill (commented):
I think this is a great idea. To make management easier, we could consider assigning teams like @WordPress/performance-admins to the Performance section. Not sure if other community teams have similar groups that could be used.

Some practical questions/concerns:

1. This should include a changelog entry, since several namespaces are changing.
2. There a chance that someone is already extending this plugin with their own custom checks. Do we need to be concerned about adding backcompat support for the previous  names, or is adding documentation to the release notes enough?

swissspidy:
We can create new teams for each category or something if we want, e.g. `@WordPress/plugin-check-plugins-team`, `@WordPress/plugin-check-performance`. Don't want to use that existing team though, as it includes too many people who aren't contributing this plugin and would just unnecessarily get pinged because of it (like right now :)).

(Edit: also keep in mind mentioning/assigning teams doesn't work for people not in the org or when logged out of GitHub)

Not worried about back compat, a changelog entry should be enough.

westonruter (approved):

joemcgill (approved):

readme.txt
```diff
@@ -68,9 +68,10 @@ In any case, passing the checks in this tool likely helps to achieve a smooth pl
 
 == Changelog ==
 
-= 1.0.2 =
+= 1.1.0 =
 
 * Fix - Correct detection of readme files in Windows by normalizing file paths.
+* Enhancement - Changed the namespace of included checks. [#460](https://github.com/WordPress/plugin-check/pull/460)
```
mukeshpanchal27:
```suggestion
* Enhancement - Changed the namespace of included checks. ([460](https://github.com/WordPress/plugin-check/pull/460))
```
Not blocker: In PL we use above formate for PR links.

mukeshpanchal27 (commented):
Thanks, @swissspidy, for the PR. Overall, it looks good to me. I've left one nit-pick and a couple of questions:

- Who is responsible for the other checks and other infrastructure-related files?
- The documentation at https://github.com/WordPress/plugin-check/tree/trunk/docs should be updated to reflect the new changes related to checks.

readme.txt
swissspidy: Here we don‘t use that format, see older changelog entries

swissspidy:
> Who is responsible for the other checks and other infrastructure-related files?

Everyone and no one :) Something we can define as we go. Suggestions welcome if you think we should define all of that already.

adamsilverstein:
Looks great and this will let individual teams iterate more rapidly. +1.

mukeshpanchal27 (approved):

swissspidy:
@bordoni @chriscct7 @davidperezgar @barrykooij Would appreciate eyes on this from the plugins team as well as it means you'll get notified more often for PRs (every time some touches checks related to your team).

IMHO adding a code owners configuration is a great way to more clearly separate concerns. Even if we end up restructuring the plugin further down the road, this would be a good thing to do.

davidperezgar:
I wait @barrykooij have an opinion of this.

But it looks great as we have very clear information where to see the checks.

davidperezgar (approved):

Issue #316, opened 2023-11-03
Provide links to further documentation for reported errors


swissspidy:
I've heard from some users that error messages they get from Plugin Check are not always very useful and lack further explanation.

I think it would be helpful to point to resources such as https://developer.wordpress.org/plugins and https://learn.wordpress.org/ where applicable. In some cases new documentation might need to be written and error messages expanded a bit.

davidperezgar:
Hello, We are already doing that in our check reviews. We can provide that info if we have some place to do it. The best way would be an accordion with all information about the check.

swissspidy:
That sounds great! The more pre-existing documentation we can use, the better. Let's start assembling that in a spreadsheet or so? 1 description/documentation/explanation per check class. Ideally the text itself is relatively short and links to the plugin handbook for further information.

Every check class could then have a new `get_description()` method. Then we need to come up with a good way to show this description in both the admin UI and when using via WP-CLI. Ideally not to repetitive.

Maybe in the admin a `<details>` block below the error message is an OK start as a proof of concept?

davidperezgar:
Hello, it's an HTML content. It would be a work to move it in the excel, and then to the plugin. I think it's better to say where is going to be saved, and we can pass all explanations when we have time.

swissspidy:
I see. Well, my suggestion would be for every check to have a `get_description()` method (and that method to be defined in the `Check` interface. That can contain HTML, no problem. I'd just suggest that this description should be relatively short (1 paragraph) and a longer version (if there is one) should live in the plugin handbook. Does that make sense?

davidperezgar:
No, they are some long with links and documentation. That gives the user all the information to explain it and solve it. So, that's why I suggest an accordion or something similar.

swissspidy:
If the long form documentation lives in the handbook or some other website, it's much easier to update and share, without having to release a new plugin version every time you fix a typo. It also makes it easier to read the documentation without first having to run the checks :)

But if you prefer to have all of that in the plugin we can figure something out as well, be it an accordion or something else.

Let's start by assembling the content here — in whatever form you prefer — and we can go from there.

davidperezgar:
I can give you some example:

```
## Generic function/class/define/namespace/option names

All plugins must have unique function names, namespaces, defines, class and option names. This prevents your plugin from conflicting with other plugins or themes. We need you to update your plugin to use more unique and distinct names.

[...]

PR #461, opened 2024-04-29
Add some initial minimal docs for each check

swissspidy:
Each _**check**_ should have a short description that can be displayed in the admin / CLI, and a link to a web page with a more detailed documentation, which can be more easily maintained.

**Note:** in the future we can add docs for each _error_ or _warning_ reported by a check. But that's out of scope for this PR.

In the admin it could look like this:

Screenshot 2024-07-29 at 14 51 31

Screenshot 2024-07-29 at 14 51 40

See #448
See #316

**To-do:**

- [ ] Rephrase short descriptions, update URLs
- [ ] Create handbook pages for some checks which don't have any matching documentation pages yet
- [ ] Maybe add a `get_name()` method so every check has a human-readable name
- [ ] Maybe move the check slug to a `get_slug()` method
- [x] Maybe centralize documentation somewhere, e.g. under https://developer.wordpress.org/plugins/
- [x] Actually display the documentation in the admin / CLI context


swissspidy:
> Actually display the documentation in the admin / CLI context

Briefly looked into this and and this needs some work for sure.  The data from`Check_Result` is essentially what's being returned via admin-ajax, and `Check_Result` can _theoretically_ hold errors & warnings from multiple checks (Even though it's currently only 1 at the time). So we would need to somehow attach more information to the `Check_Result` (check name, description, and URL) for each check and display that in the admin in the right place. Aside: there's also no connection between an added error and which check it's coming from.

Then there are also some detailed descriptions for specific errors, as shown in #316. Those are better put on the handbook website.

davidperezgar:
Hello, you're right, every check has their own description. I'd prefer to have them all in a folder with html files. That's how we have it already in our tool. And it's much easier to maintain these files.

swissspidy:
I still think it makes more sense to put the long versions on the handbook website for easier maintenance and reference.

If you want them as part of the plugin, then they all need to be translatable, so an HTML file is not good enough.

davidperezgar:
Yes, you're right in make it translatable. It's not that long an is a content of years and years reviewing, so I think is very important to have it.

davidperezgar:
I'm reviewing this PR but I see that description is for all checks, and We do need to have a description in every check. Maybe we could change the description function to have parameters, so we can pass a slug of check and it give the solution. 

I give an example of a description of check:

> ## Not permitted files
> 
> A plugin typically consists of files related to the plugin functionality (php, js, css, txt, md) and maybe some multimedia files (png, svg, jpg) and / or data files (json, xml).
> 
> We have detected files that are not among of the files normally found in a plugin, are they necessary? If not, then those won't be allowed.
> 


What do you think?

swissspidy:
In the admin it could look like this:

Screenshot 2024-07-29 at 14 51 31

Screenshot 2024-07-29 at 14 51 40

davidperezgar:
Yes, that's good option, but I'd add it as well in the results of the checks.

swissspidy:
> Yes, that's good option, but I'd add it as well in the results of the checks.

Let's do that in a follow-up issue.

As per https://github.com/WordPress/plugin-check/pull/461#issuecomment-2206239406 that is going to be more complicated. The results do not have any information about the check attached, and we don't currently send that information via the Ajax responses.

Also keep in mind that result ≠ check. 1 Check can have hundreds of results. So we'd need to figure out how to best display the information without adding too much noise.

includes/Admin/Admin_Page.php
```diff
@@ -91,6 +94,80 @@ public function add_and_initialize_page() {
...
+			<dd>
+				<?php echo wp_kses( $check->get_description(), array( 'code' => array() ) ); ?>
+				<br>
+				<a href="<?php echo esc_url( $check->get_documentation_url() ); ?>">
+					<?php esc_html_e( 'Learn more', 'plugin-check' ); ?>
+				</a>
+			</dd>
+			<?php
+
+			echo '';
```
mukeshpanchal27:
```suggestion
```

includes/Checker/Checks/Performance/Enqueued_Resources_Check.php
```diff
@@ -47,4 +47,30 @@ protected function get_args() {
...
+	 * Gets the description for the check.
+	 *
+	 * Every check must have a short description explaining what the check does.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return string Description.
+	 */
+	public function get_description(): string {
+		return __( 'Checks whether scripts and styles are properly enqueued using the recommended way', 'plugin-check' );
```
mukeshpanchal27:
```suggestion
		return __( 'Checks whether scripts and styles are properly enqueued using the recommended way.', 'plugin-check' );
```

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -240,4 +240,30 @@ private function get_viewable_post_types() {
...
+	 * Gets the description for the check.
+	 *
+	 * Every check must have a short description explaining what the check does.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return string Description.
+	 */
+	public function get_description(): string {
+		return __( 'Checks whether scripts and styles are enqueued using a recommended loading strategy', 'plugin-check' );
```
mukeshpanchal27:
```suggestion
		return __( 'Checks whether scripts and styles are enqueued using a recommended loading strategy.', 'plugin-check' );
```

includes/Checker/Checks/Security/Direct_DB_Queries_Check.php
```diff
@@ -47,4 +47,30 @@ protected function get_args() {
...
+	 * Gets the description for the check.
+	 *
+	 * Every check must have a short description explaining what the check does.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return string Description.
+	 */
+	public function get_description(): string {
+		return __( 'Checks the usage of direct database queries, which should be avoided', 'plugin-check' );
```
mukeshpanchal27:
```suggestion
		return __( 'Checks the usage of direct database queries, which should be avoided.', 'plugin-check' );
```

mukeshpanchal27 (commented):
The changes looks good to me. Left some nit-pick

Could you please update https://github.com/WordPress/plugin-check/tree/trunk/docs, the static and runtime check?

davidperezgar:
I propose this. Make the title of the result, description and link in the same page. 

Yes it's longer, but our experience reviewing plugins is that developers need help solving the issues.

screenshot-upload

swissspidy:
@davidperezgar but that‘s the message for an individual error, not a whole check. Let‘s handle that in a follow-up issue, as per my above comment

includes/Checker/Checks/Performance/Enqueued_Styles_Size_Check.php
```diff
@@ -264,4 +264,34 @@ private function get_viewable_post_types() {
...
+	 *
+	 * Every check must have a short description explaining what the check does.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return string Description.
+	 */
+	public function get_description(): string {
+		return sprintf(
+		/* translators: %s: Script size threshold. */
```
westonruter:
```suggestion
			/* translators: %s: Script size threshold. */
```

westonruter (approved):
One tiny nit, but pre-approving pending applying Mukesh's suggestions.

davidperezgar:
Ok @swissspidy , make that follow up.

davidperezgar:
We need for results, not whole check. The explanations for the results vary a lot between them.

ernilambar (approved):

mukeshpanchal27 (approved):
The documents changes can we done in followup PR.

PR #473, opened 2024-06-19
Fix relative wp-env paths when host filepaths contain spaces

joemcgill:
This fixes the path passed to `wp-env` via the --env-cwd arg when `pwd` returns a file path containing a space.

## Testing instructions

1. Clone this repo to a folder locally that includes a space, like `~/Local Sites/plugin-check`.
2. Run a script like `npm run lint-php` and see that it's trying to run from `/var/www/html/wp-content/plugins/Local`
3. Check out this PR
4. Run the same script and see that it is running successful from `/var/www/html/wp-content/plugins/plugin-check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,14 +19,14 @@
     "wp-env": "wp-env",
     "lint-js": "wp-scripts lint-js",
     "format-js": "npm run lint-js -- --fix",
-    "phpstan": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename "$(pwd)")/ composer run-script phpstan",
-    "lint-php": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename "$(pwd)")/ composer run-script lint",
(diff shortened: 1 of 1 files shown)


joemcgill:
This came up while helping @davidperezgar run PHPUnit tests locally, see https://github.com/WordPress/plugin-check/issues/466#issuecomment-2179449685

joemcgill:
The Unit tests seem to be failing due to an old version of the wp-env package. I've raised a separate PR that should fix that issue here: #474.

swissspidy (approved):

mukeshpanchal27 (approved):

joemcgill:
Thanks, all!

davidperezgar:
Nice! thanks!!!

davidperezgar:
That will prevent a lot of users having random problem when using unit tests...

Issue #478, opened 2024-06-27
Allow static checks to be run against an arbitrary path


joemcgill:
Currently the WP CLI command is limited to only run checks against plugins that are installed on the same site as Plugin Check. 

This is because it eventually calls the `Plugin_Request_Utility::get_plugin_basename_from_input()` method during the setup process which throws an Exception if the plugin is not installed.

To satisfy the requirements of https://github.com/WordPress/plugin-check/issues/441#issuecomment-2190961036, we need to be able to run checks against a plugin that is located outside the WordPress filesystem, e.g. `/tmp`.

Ideally, this would work by passing a CLI command like:

```
wp plugin check /tmp/plugin-path
```

When the path is not to an active plugin, only static checks should be eligible to be run.

swissspidy:
Related:

* #153

davidperezgar:
Yes! It's necessary for the form submission.

davidperezgar:
I've commited first approach how to use in an arbitrary path. It's working. Only rename variables would need it.

Issue #479, opened 2024-06-27
Create a better solution for severity levels


barrykooij:
Another thing we'd have to change before we can integrate on w.org is a more flexible "severity levels" solutions. Currently, we have 2 array variables `$errors` and `$warnings` but we need to be able to extend this.

For the w.org integration specifically we need a higher level than error, called (something along the lines of) "blocking". If this is > 0, the submission will not be allowed to .org. This is mainly because we can't automatically block every error, because of false positives. And reducing errors to warning because of this reason feels wrong.

Also, in the future, I'd love to see something like a "notice" added with suggestions / best practices.
I suggest we move away from having a variable per "level" and move into a more extendable way of setting this up via a "severity level collection" like solution. This way other people can even add (or remove) levels via a filter for example.

Relevant file: https://github.com/WordPress/plugin-check/blob/trunk/includes/Checker/Check_Result.php#L31

I'll also add this in the doc / plan. Love to hear your thoughts on this.

---

Talked this over with @joemcgill in Slack, we discussed the possibility for having different ruleset xml files and/or using PHPCS severity levels. This solves the PHPCS side of things, but PCP has more (and will have a lot more in the future) checks not related to PHPCS. We need to support multiple severity levels for this as well.

joemcgill:
Our conversation can be found [here](https://wordpress.slack.com/archives/C079B9FMBJ9/p1719512889233049).

To summarize, I'd suggest trying to adapt the existing [severity levels strategy used by PHPCS](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#changing-the-default-severity-levels).

We would still sort failed checks into two groups, Errors and Warning, but each would include a severity level ranging from 0–10, with 5 being the default level that is shown in reports, unless otherwise specified. This would allow the output of checks to be limited to only show errors or warnings that meet a specific severity threshold:

ex: `wp plugin check ${slug} --ignore-warnings --severity=8`

For any checks that are extending the `Abstract_PHP_CodeSniffer_Check` class, we could pass the severity levels as args to the PHPCS Runner, and/or add a new param to the `Amend_Check_Result::add_result_message_for_file()` so that the severity is collected along with other info, like file, line, column, etc.

We would need to adapt non PHPCS checks to make use of this severity system so that any messages not meeting the specified severity level would either be filtered out upon output or not saved to the arrays in the first place (probably the former).

swissspidy:
[...]

Issue #440, opened 2024-03-20
Plugin Review Team: Ensure top level of "Error" includes no false positives.


bordoni:
As we talked last week @barrykooij will own this task to making sure that we have either:
- A new level of error that includes no false positives.
- Modify the current top level items to not have any false positives.

That is important to enables us to have it be part of the submission on WordPress.org, which should enable us to reduce the Queue and maintain it at a good level.

Currently there is no way to appeal specific items that might be a false positive so this is a must.

felixarntz:
@bordoni @barrykooij Curious to learn more about this, can you provide a bit more context on what this is about? How could we possibly ensure there are no false positives? Maybe I'm misunderstanding the issue description.

barrykooij:
@felixarntz This issue is about identifying what phpcs rules are 100% blocking for new submission. These should not have any false positives.

An example of a 100% blocking check is `DisallowShortOpenTag`. We never accept this, no exceptions and no false positives. This issue is to identify all rules that are like this so we can use these rules as a first step in the plugin upload process.

felixarntz:
@barrykooij Thanks, makes sense!

Is the idea to put those rules into a new PHPCS config once they're identified? Or a new check group? How would those rules technically be annotated / grouped so that only those would run in the new submission process?

barrykooij:
@felixarntz I'm not sure yet, whatever works best. I'm currently leaning towards creating a set (check group?) of "blocking" PHPCS rules and if any of these fails, a submission is blocked. Another thing we need to in mind is that we're going to expand the checks to other form of checks than PHPCS, like a readme parser.

I think it would be good to also show this in the "regular" PCP wp-admin result, so a developer can see that the current version of the plugin contains blocking issues. 

I'm currently creating a meta trac ticket to discuss how to run only these checks in a new plugin submission. Related to this is also #441

felixarntz:
@barrykooij 
> I'm currently leaning towards creating a set (check group?) of "blocking" PHPCS rules and if any of these fails, a submission is blocked. Another thing we need to in mind is that we're going to expand the checks to other form of checks than PHPCS, like a readme parser.

In that case a "check category" probably makes most sense. For instance, we could add a "Plugin Submission" or "Plugin Repo (blocking)" category to https://github.com/WordPress/plugin-check/blob/trunk/includes/Checker/Check_Categories.php#L48 and assign the relevant checks to that category.

barrykooij:
[...]

PR #480, opened 2024-06-28
Results from WPCLI that allows implement in w.org

### LINKED ISSUE #441: Plugin Review Team: Output for WP.org form
We need a way to output a simplified version of the table of checks, by passing a couple of params so we can hook into the WP.org output with an error.

No false positives are permitted here, we will only output things that are required to pass. Which will be handled on #440 by @barrykooij.

---

More details to come on this.

davidperezgar:
We need to give a result from WPCLI that fits meta team in order to execute this plugin in form submission.

This results fits the need asked by Meta [#441 (comment)](https://github.com/WordPress/plugin-check/issues/441#issuecomment-2190961036), so we could implement as well as this #478 , and it would be the minimal needs to achieve that.

The results has a json format like this:
`{"pass":true,"errors":[],"warnings":[]}` 

The Pass value is true if it has no errors. 

I'd need some feedback if the approach is correct, or we could do something more.


joemcgill:
As asked in https://github.com/WordPress/plugin-check/issues/441#issuecomment-2195224108

> I'm curious if the presence of any errors returned by the CLI command would be sufficient as a pass/fail signal, or if the command needs to also return a specific error code as well?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -98,6 +98,9 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * : Additional directories to exclude from checks
 	 * By default, `.git`, `vendor` and `node_modules` directories are excluded.
 	 *
+	 * [--wporg]
```
swissspidy: Instead of a new argument for this, it would make more sense to add another option to `--format`, e.g. `--format=wporg`

davidperezgar: That was my first approach, but it gives more things related with that argument.

davidperezgar: And We will for sure make changes in the plugin for needs from Meta, and It will easier for us keeping the command.

davidperezgar:
> As asked in [#441 (comment)](https://github.com/WordPress/plugin-check/issues/441#issuecomment-2195224108)
> 
> > I'm curious if the presence of any errors returned by the CLI command would be sufficient as a pass/fail signal, or if the command needs to also return a specific error code as well?

I think is the easiest way. We have to be sure that errors haven't got any false positive.

includes/CLI/Plugin_Check_Command.php
swissspidy: I suppose long term, if you need more changes, you could also look into extending this class with your own custom command (e.g. `class Plugin_Review_Plugin_Check_Command extends Plugin_Check_Command`) or something.

Just from a code quality perspective a custom flag like this isn't exactly a best practice :)

davidperezgar: done

includes/CLI/Plugin_Check_Command.php
```diff
@@ -77,6 +78,7 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 *   - table
 	 *   - csv
 	 *   - json
+	 *   - wporg
```
mukeshpanchal27, line 84: Add new option in docblock.

`Format to display the results. Options are table, csv, json, and wporg. The default will be a table.`

includes/CLI/Plugin_Check_Command.php
```diff
@@ -210,6 +214,18 @@ static function ( $dirs ) use ( $excluded_directories ) {
...
+		if ( $run_wporg ) {
+			$all_results = array(
+				'pass'     => empty( $errors ) ? true : false,
+				'errors'   => $errors,
+				'warnings' => $warnings,
+			);
+
+			WP_CLI::line( json_encode( $all_results ) );
+			return;
+		}
```
```suggestion
		// Print the json formatted results for WPORG.
		if ( isset( $options['format'] ) && 'wporg' === $options['format'] ) {
			$wporg_results = array(
				'pass'     => empty( $errors ) ? true : false,
				'errors'   => $errors,
				'warnings' => $warnings,
			);

			WP_CLI::line( json_encode( $wporg_results ) );
			return;
		}
```

- Remove local variable and incorporate the condition inline.
- Slightly update the comment.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -133,6 +135,8 @@ public function check( $args, $assoc_args ) {
 		$plugin = isset( $args[0] ) ? $args[0] : '';
 		$checks = wp_parse_list( $options['checks'] );
 
+		$run_wporg = isset( $options['format'] ) && 'wporg' === $options['format'] ? true : false;
+
 		// Create the categories array from CLI arguments.
```
```suggestion
		// Create the categories array from CLI arguments.
```

mukeshpanchal27 (changes requested):
Thanks @davidperezgar for the PR! Let some feedbacks other then it looks good to me.

- Fix `composer phpmd` errors.
   - Add `@SuppressWarnings(PHPMD.ExcessiveClassComplexity)` in class `Plugin_Check_Command` docblock.
   - Add `@SuppressWarnings(PHPMD.CyclomaticComplexity)` in `check` method docblck.
-  Additionally if you add Behat tests for new args then it would be great.

davidperezgar:
Thanks for your suggestions!

joemcgill:
I don't particularly like the approach of creating a new output format specific for wporg if we are instead able to consume one of the exiting formats when passing the correct arguments to the CLI script.

Before trying to implement a solution here, can we get an answer from systems (e.g., @dd32) to the question asked in https://github.com/WordPress/plugin-check/issues/441#issuecomment-2195224108? We also need to decide if #479 is a blocker to this issue, because we may not need a special CLI command at all if we can configure what errors are returned based on severity levels.

davidperezgar:
Related with #479 and #440 .

mukeshpanchal27:
@davidperezgar I have fixed PHPMD issues and added behat test for new argument.

ernilambar:
I tried this command:

```
wp plugin check foo-single.php --fields=line,column,code --format=wporg
```

Argument `--fields` is not respected when using `wporg` format.

davidperezgar:
Yes, and it's correct. The format for wporg is very strict and I shouldn't be customized.

dd32:
I don't see anything "wrong" here, but I don't see the purpose of it, other than to fix the fact that `--format=json` isn't a pure JSON output.

Instead of adding this for the purposes of adding it to w.org plugin submission, I'd probably just suggest fixing the existing json output, or request that we parse the existing json output.

My main issue with this PR is that:
> 'pass'     => empty( $errors ) ? true : false,

Is completely useless. Pretty much every plugin throws up errors in plugin-check that are considered errors that aren't actually errors.

The original "A command that can be run, parsed, and provides a clear block or accept result" really still stands, and while this does do that, it still doesn't provide a the last part in any form that's inline with the plugin submission guidelines IMHO.

For example; Are this really something that should block a plugin submission?
> | 0    | 0      | ERROR   | outdated_tested_upto_header | Tested up to: 6.5 < 6.6                                                                     |
> | 37   | 25     | ERROR | WordPress.WP.I18n.MissingTranslatorsComment | A function call to `__()` with texts containing placeholders was found, but was not accompanied by a "translators:" comment on the line above to clarify the meaning of the placeholders. |
> | 59   | 21     | ERROR | WordPress.Security.EscapeOutput.ExceptionNotEscaped | All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found `'__'`.            |
> | 114  | 30     | ERROR   | WordPress.WP.AlternativeFunctions.json_encode_json_encode | json_encode() is discouraged. Use wp_json_encode() instead. |

So I guess actually, this all comes back to #479 instead as the main blocker.

davidperezgar:
Yes, We're working on severity levels, and the team where agreed equals and more than 7, should be an error. 

So, I see:
- We should have a format=wporg as json is for users.
- We have we give you then after the severity level an ERROR or WARNING
- Do you need then the pass variable?

For the example that you give: 
1. ERROR
2. WARNING
3. ERROR
4. ERROR

So we have to work on severity levels and then give you the result.

davidperezgar:
Hello, 
I've updated the code with the severity level and wporg format. I've made within iterative for all formats, but creating an array for errors and warning after severity filter.

I've removed pass as is useless.

We can discuss what's your opinion.

davidperezgar:
Thanks Nilambar for the changes, I think is ready for a review from @dd32

dd32:
Am I understanding this correctly?

```
$output = json_decode( wp plugin-check .... )
$error_types = wp_list_pluck( $output->errors, 'type' );
$pass = ! in_array( 'ERROR', $error_types );
```

If so, yep, that's fine.

davidperezgar:
Yes Dion! ERRORS will block an upload. But please show as well warnings as they will help to reduce the time for reviewing.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -248,6 +250,15 @@ static function ( $dirs ) use ( $excluded_files ) {
 		$error_severity   = ! empty( $options['error-severity'] ) ? $options['error-severity'] : $options['severity'];
 		$warning_severity = ! empty( $options['warning-severity'] ) ? $options['warning-severity'] : $options['severity'];
 
+		$format_is_wporg = ( isset( $options['format'] ) && 'wporg' === $options['format'] ) ? true : false;
```
swissspidy:
```suggestion
		$format_is_wporg = ( isset( $options['format'] ) && 'wporg' === $options['format'] );
```

davidperezgar:
@dd32 Do you need something else about this PR? Thanks.

dd32 (approved):
> @dd32 Do you need something else about this PR? Thanks.

Nope, It was just waiting on merging the other needed PRs AFAIK.

ernilambar (approved):

includes/CLI/Plugin_Check_Command.php
```diff
@@ -248,6 +250,15 @@ static function ( $dirs ) use ( $excluded_files ) {
 		$error_severity   = ! empty( $options['error-severity'] ) ? $options['error-severity'] : $options['severity'];
 		$warning_severity = ! empty( $options['warning-severity'] ) ? $options['warning-severity'] : $options['severity'];
 
+		$format_is_wporg = ( isset( $options['format'] ) && 'wporg' === $options['format'] );
```
```suggestion
		$is_wporg_format = isset( $options['format'] ) && 'wporg' === $options['format'];
```

Nit. name should be `is_wporg_format`

includes/CLI/Plugin_Check_Command.php
```diff
@@ -676,4 +700,33 @@ function ( $item ) use ( $warning_severity ) {
 
 		return array_merge( $errors, $warnings );
 	}
+
+	/**
+	 * Returns separated results by ERROR and WARNING.
+	 *
+	 * @since 1.1.0
```
```suggestion
	 * @since 1.2.0
```

Should be upcoming version.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -676,4 +700,33 @@ function ( $item ) use ( $warning_severity ) {
...
+			function ( $item ) {
+				return ( 'WARNING' === $item['type'] );
+			}
+		);
+
+		return array(
+			'errors'   => array_values( $errors ),
+			'warnings' => array_values( $warnings ),
+		);
+	}
```
```suggestion
	private function get_separated_results( $results ) {
		// Initialize arrays for errors and warnings.
		$errors   = array();
		$warnings = array();

		// Return early if $results is not an array or is empty.
		if ( ! is_array( $results ) || empty( $results ) ) {
			return array(
				'errors'   => $errors,
				'warnings' => $warnings,
			);
		}

		// Separate errors and warnings in one pass.
		foreach ( $results as $item ) {
			if ( ! isset( $item['type'] ) ) {
				continue;
			}

			switch ( $item['type'] ) {
				case 'ERROR':
					$errors[] = $item;
					break;
				case 'WARNING':
					$warnings[] = $item;
					break;
			}
		}

		return array(
			'errors'   => $errors,
			'warnings' => $warnings,
		);
	}
```

- Early return if no error.
- Refactor code.

dd32 (changes requested):
Let's not merge this, until it's determined to be needed. See my comment above on #441

davidperezgar:
Ok!

includes/CLI/Plugin_Check_Command.php
davidperezgar: ok!

ernilambar:
Probably we dont need this PR now. It seems it is possible to integrate in org without new format.

dd32:
Closing, we've taken a different direction in implementation. See https://meta.trac.wordpress.org/ticket/7778

davidperezgar:
Ok! No problem, if the result is the same! ;)

PR #483, opened 2024-07-01
Readme Check: Compare `Tested up to` value with latest stable version

### LINKED ISSUE #481: Check: Tested up to Latest version of WordPress
I would add to check the actual version of WordPress. It must be Tested up "actual version".

In order to have always the latest version, I've [asked to meta](https://wordpress.slack.com/archives/C02QB8GMM/p1719654997418999), and we have some methods to check that:

> Yes, use one of these:
> https://api.wordpress.org/core/stable-check/1.0/
> 3rd line of https://api.wordpress.org/core/version-check/1.1/
> Or just json decode https://api.wordpress.org/core/version-check/1.7/
> So finally we have a complete check.

The error that we show to the users is something like this ERROR: Tested up to: 6.4 < 6.5

And we don't compare with minor version, just major.

It should have a transient with expired date 1 day.

Description for this issue:

> ## Tested Up To Value is Out of Date, Invalid, or Missing
> 
> The tested up to value in your plugin is not set to the current version of WordPress. This means your plugin will not show up in searches, as we require plugins to be compatible and documented as tested up to the most recent version of WordPress.
> 
> Please update your readme to show that it is tested up to the most recent stable major version of WordPress. For example if WordPress said that version 99.0.1 was the current release, you would be able to set your value to 99.0 or 99.0.1, as 99.0.1 would be considered a minor release.
> 
> The following links will assist you in understanding WordPress's versioning and the latest version:
> 
> https://wordpress.org/download/
> https://make.wordpress.org/core/handbook/about/release-cycle/version-numbering/
> 
> You cannot set it beyond the current version, as that will cause your plugin not to be available on searches. Also remember that WordPress has made late changes in releases, so claiming compatibility with an incomplete version is likely to land you in trouble.
> 
> From your readme file:
> 
> ERROR: Tested up to: 6.4 < 6.5

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/481

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -477,11 +477,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
 					$latest_release = reset( $body['offers'] );
 
-					$version = $latest_release['current'];
-
-					if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
-						$version = $matches[0];
-					}
+					$version = $latest_release['new_bundled'];
 
 					set_transient( 'wp_plugin_check_latest_wp_version', $version, DAY_IN_SECONDS );
 				}
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -150,7 +150,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		foreach ( $fields as $field_key => $field ) {
 			if ( ! in_array( $field['ignore_key'], $ignored_warnings, true ) && ! isset( $parser_warnings[ $field['ignore_key'] ] ) ) {
 
-				if ( 'tested' === $field_key ) {
+				if ( ! empty( $parser->{$field_key} ) && 'tested' === $field_key ) {
 					$latest_wordpress_version = $this->get_wordpress_stable_version();
 					if ( version_compare( $parser->{$field_key}, $latest_wordpress_version, '<' ) ) {
 						$this->add_result_error_for_file(
@@ -477,7 +477,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
 					$latest_release = reset( $body['offers'] );
 
-					$version = $latest_release['new_bundled'];
+					$version = $latest_release['current'];
 
 					set_transient( 'wp_plugin_check_latest_wp_version', $version, DAY_IN_SECONDS );
 				}
@@ -490,10 +490,10 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 
 			// Strip off any -alpha, -RC, -beta suffixes.
 			list( $version, ) = explode( '-', $version );
+		}
 
-			if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
-				$version = $matches[0];
-			}
+		if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
+			$version = $matches[0];
 		}
 
 		return $version;
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -148,17 +148,37 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		$parser_warnings = is_array( $parser->warnings ) ? $parser->warnings : array();
 
 		foreach ( $fields as $field_key => $field ) {
-			if ( empty( $parser->{$field_key} ) && ! in_array( $field['ignore_key'], $ignored_warnings, true ) && ! isset( $parser_warnings[ $field['ignore_key'] ] ) ) {
-				$this->add_result_warning_for_file(
-					$result,
-					sprintf(
-						/* translators: %s: plugin header tag */
-						__( 'The "%s" field is missing.', 'plugin-check' ),
-						$field['label']
-					),
-					'missing_readme_header',
-					$readme_file
-				);
+			if ( ! in_array( $field['ignore_key'], $ignored_warnings, true ) && ! isset( $parser_warnings[ $field['ignore_key'] ] ) ) {
+
+				if ( 'tested' === $field_key ) {
+					$latest_wordpress_version = $this->get_wordpress_stable_version();
+					if ( version_compare( $parser->{$field_key}, $latest_wordpress_version, '<' ) ) {
+						$this->add_result_error_for_file(
+							$result,
+							sprintf(
+								/* translators: 1: currently used version, 2: latest stable WordPress version */
+								__( 'Tested up to: %1$s < %2$s', 'plugin-check' ),
+								$parser->{$field_key},
+								$latest_wordpress_version
+							),
+							'outdated_tested_upto_header',
+							$readme_file
+						);
+					}
+				} else {
+					if ( empty( $parser->{$field_key} ) ) {
+						$this->add_result_error_for_file(
+							$result,
+							sprintf(
+								/* translators: %s: plugin header tag */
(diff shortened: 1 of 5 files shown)


includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -148,17 +148,37 @@ private function check_headers( Check_Result $result, string $readme_file, Parse
...
+								$parser->{$field_key},
+								$latest_wordpress_version
+							),
+							'outdated_tested_upto_header',
+							$readme_file
+						);
+					}
+				} else {
+					if ( empty( $parser->{$field_key} ) ) {
+						$this->add_result_warning_for_file(
```
davidperezgar: I'd change for error. Thanks

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -457,7 +477,11 @@ private function get_wordpress_stable_version() {
 				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
 					$latest_release = reset( $body['offers'] );
 
-					$version = $latest_release['new_bundled'];
+					$version = $latest_release['current'];
+
+					if ( preg_match( '#^\d.\d#', $version, $matches ) ) {
+						$version = $matches[0];
+					}
```
swissspidy: I assume you changed this because `new_bundled` seems to be outdated in the API?

Let's revert this change here and instead make sure the API is fixed :-)

ernilambar:
Refinement in Unit tests:

We could remove over-scoping of test methods so that we would not have to update tests in every readme related PR. 

Example:

```
1) Plugin_Readme_Check_Tests::test_run_with_errors_invalid_name
Failed asserting that 2 matches expected 1.
```
 
When we are testing ERROR for invalid plugin name, we should not be caring how many warnings are there. We should focus on whether our ERROR is there or not.

Related: https://github.com/WordPress/plugin-check/actions/runs/9743511198/job/26887103212#step:7:270

swissspidy:
Agreed, I don't know why all these tests are doing that.

davidperezgar:
I've made some changes: Detect major version (for example 6.5) and prevent some readme file without version added.

swissspidy:
> I've made some changes: Detect major version (for example 6.5) and prevent some readme file without version added.

Let's undo the changes to `get_wordpress_stable_version`.

We should rely on https://api.wordpress.org/core/version-check/1.7/ providing a correct value for `new_bundled`, which is always the latest major (e.g. 6.5).

(Note: for some reason it says 6.4 right now but I [already reported this bug](https://wordpress.slack.com/archives/C02QB8GMM/p1719833280497749?thread_ts=1719654997.418999&cid=C02QB8GMM)).

includes/Checker/Checks/Plugin_Readme_Check.php
davidperezgar: But I don't understand current is current version of WordPress, right?

swissspidy: But `current` is `6.5.5`. We only need `6.5.`

We should not need to do any regex whatsoever to get from `6.5.5` to `6.5`.

The `new_bundled` field already provides the right value for us.

davidperezgar: But it's not giving. I think the fastest way is making a regex as it's in the PR

ernilambar:
I refactored the test `test_run_with_errors_invalid_name` like this:
```
public function test_run_with_errors_invalid_name() {
    $readme_check  = new Plugin_Readme_Check();
    $check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-plugin-readme-errors-invalid-name/load.php' );
    $check_result  = new Check_Result( $check_context );

    $readme_check->run( $check_result );

    $errors = $check_result->get_errors();

    $this->assertNotEmpty( $errors );
    $this->assertArrayHasKey( 'readme.txt', $errors );

    // Check for invalid name error.
    $this->assertArrayHasKey( 0, $errors['readme.txt'] );
    $this->assertArrayHasKey( 0, $errors['readme.txt'][0] );
    $this->assertSame( 1, count( wp_list_filter( $errors['readme.txt'][0][0], array( 'code' => 'invalid_plugin_name' ) ) ) );
}
```

Here:
* I have removed assertions related to warnings
* Also for error, I am not considering it is in either `0` or `1` position because position of the error could change depending upon order of methods `$this->check_name()` and `$this->check_headers` called.

Errors array:

```
(
    [readme.txt] => Array
        (
            [0] => Array
                (
                    [0] => Array
                        (
                            [0] => Array
                                (
                                    [message] => Plugin name header in your readme is missing or invalid. Please update your readme with a valid plugin name header. Eg: "=== Example Name ==="
                                    [code] => invalid_plugin_name
                                    [link] => 
                                )

                            [1] => Array
                                (
                                    [message] => Tested up to: 6.1 < 6.5
                                    [code] => outdated_tested_upto_header
                                    [link] => 
                                )

                        )

                )

        )

)
```

Is this approach good?

CC @swissspidy @mukeshpanchal27

includes/Checker/Checks/Plugin_Readme_Check.php
swissspidy: But we're not in a rush, no need to use the "fastest" way. Also, arguably, simply fixing the API is faster.

Definitely an opponent of doing unnecessary regex checks like this.

swissspidy:
You can use `assertCount()` to be more precise, instead of `assertSame( 1, count( `

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -457,7 +477,7 @@ private function get_wordpress_stable_version() {
 				if ( isset( $body['offers'] ) && ! empty( $body['offers'] ) ) {
 					$latest_release = reset( $body['offers'] );
 
-					$version = $latest_release['new_bundled'];
+					$version = $latest_release['current'];
```
swissspidy, line 482: I stand corrected, `current` is actually the correct field to use, not `new_bundled`.

swissspidy:
In the future we could think about custom PHPUnit assertions to make our lives easier.

ernilambar:
> In the future we could think about custom PHPUnit assertions to make our lives easier.

I tried this approach also. https://github.com/ernilambar/plugin-check/pull/10/files

ernilambar:
@swissspidy For tests like `public function test_run_without_errors()`, how can we update test such that we wont have to fix test when we add new check giving error/warning?

swissspidy:
You mean specifically for `\Plugin_Readme_Check_Tests::test_run_without_errors()`? Well, it depends.

Given that this PR updates the check and the `readme.txt` used in tests (`tests/phpunit/testdata/plugins/test-plugin-plugin-readme-without-errors/readme.txt`) is very basic, it makes sense that the test now fails.

If you mean specifically for the `Tested up to` field in that file, we could think about some automation to keep that up-to-date with the latest WP version. But that can be discussed in a new issue.

ernilambar:
Not specific to `Tested up to`. Lets say in the future `Short Description` will be **required**. Then these tests like `test_run_without_errors`, `test_run_md_without_errors`, `test_run_root_readme_file_without_errors`, etc will fail.

What if have an fixed array of codes which we check against. Though, in that case `test_run_without_errors` name wont be appropriate coz we will be just checking few codes only.

swissspidy:
There's definitely room for improvement in how tests are structured and how we check for error codes.

And tests like "runs without errors" are a bit too broad. Better to check for good and bad cases for each individual sub-check.

But let's discuss that in a new issue :-)

swissspidy (approved):

davidperezgar:
What happen with the version? We keep regex then?

swissspidy:
Yes. Turns out I was wrong and we have to keep using `current`

includes/Checker/Checks/Plugin_Readme_Check.php
swissspidy, line 1: Need to add `@SuppressWarnings(PHPMD.ExcessiveClassComplexity)` to the class doc block

davidperezgar (approved):
I see everything is OK.

PR #485, opened 2024-07-01
Implement `Enqueued_Scripts_Scope_Check`

### LINKED ISSUE #23: Create Enqueued_Scripts_Scope_Check
## Description

Checks that scripts are only enqueued in certain contexts and not just everywhere. While a theme often enqueues scripts or styles that are needed for the entire site, plugins usually have specific behaviors that often do not affect the entire site. Such plugins should not just call wp_enqueue_script() on the wp_enqueue_scripts action, but have conditional logic in there to only enqueue the script when needed based on context. This check would load different contexts (simulating different URLs) of different kinds and warn if a script is enqueued on all of them.

## Acceptance Criteria

- Class `Enqueued_Scripts_Scope_Check ` should be created and exists at `includes/Checker/Checks`
- Class `Enqueued_Scripts_Scope_Check ` should extend the `Runtime_Check` abstract class which extends `Check` interface
- Method `run` should be implemented from the `Runtime_Check` abstract class if needed

## Test Coverage

- `Enqueued_Scripts_Scope_Check ` class is expected to extend `Runtime_Check` class
- `Enqueued_Scripts_Scope_Check ` class is expected to be instance of `Runtime_Check` class

swissspidy:
Note: this is basically an exact copy of `Enqueued_Styles_Scope_Check` right now. And the other script/style-related checks are very similar too.

I suggest creating some abstraction in a new, separate PR to remove repeated code. 

Fixes #23

### DIFF
--- a/includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
+++ b/includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
@@ -1,0 +1,276 @@
+<?php
+/**
+ * Class WordPress\Plugin_Check\Checker\Checks\Enqueued_Scripts_Scope_Check
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks;
+
+use Exception;
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation;
+use WordPress\Plugin_Check\Checker\With_Shared_Preparations;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+use WordPress\Plugin_Check\Traits\URL_Aware;
+
+/**
+ * Check if a script is present in all URLs.
+ *
+ * @since 1.0.2
+ */
+class Enqueued_Scripts_Scope_Check extends Abstract_Runtime_Check implements With_Shared_Preparations {
+
+	use Amend_Check_Result;
+	use Stable_Check;
+	use URL_Aware;
+
+	/**
+	 * List of viewable post types.
+	 *
+	 * @since 1.0.2
+	 * @var array
+	 */
+	private $viewable_post_types;
+
+	/**
+	 * List of plugin scripts.
+	 *
+	 * @since 1.0.2
+	 * @var array
+	 */
+	private $plugin_scripts = array();
+
+	/**
+	 * Plugin script counter.
+	 *
+	 * @since 1.0.2
+	 * @var array
+	 */
+	private $plugin_script_count = array();
+
+	/**
+	 * Gets the categories for the check.
+	 *
+	 * Every check must have at least one category.
+	 *
+	 * @since 1.0.2
+	 *
+	 * @return array The categories for the check.
+	 */
+	public function get_categories() {
+		return array( Check_Categories::CATEGORY_PERFORMANCE );
+	}
+
+	/**
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since 1.0.2
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		$orig_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
+
+		// Backup the original values for the global state.
+		$this->backup_globals();
+
+		return function () use ( $orig_scripts ) {
+			if ( is_null( $orig_scripts ) ) {
+				unset( $GLOBALS['wp_scripts'] );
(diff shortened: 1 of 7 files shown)


includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+	 * Runs this preparation step for the environment and returns a cleanup function.
+	 *
+	 * @since 1.0.2
+	 *
+	 * @return callable Cleanup function to revert any changes made here.
+	 *
+	 * @throws Exception Thrown when preparation fails.
+	 */
+	public function prepare() {
+		$orig_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
```
westonruter, line 77: This can be simplified to:

```suggestion
		$orig_scripts = $GLOBALS['wp_scripts'] ?? null;
```

But not required.

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+	 * @since 1.0.2
+	 *
+	 * @return array Returns a map of $class_name => $constructor_args pairs. If the class does not
+	 *               need any constructor arguments, it would just be an empty array.
+	 */
+	public function get_shared_preparations() {
+		$demo_posts = array_map(
+			static function ( $post_type ) {
+				return array(
+					'post_title'   => "Demo {$post_type} post",
```
westonruter, line 105: Advised by PhpStorm:

```suggestion
					'post_title'   => "Demo $post_type post",
```

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+				'post_type'              => $post_type,
+				'post_status'            => array( 'publish', 'inherit' ),
+				'ignore_sticky_posts'    => true,
+				'no_found_rows'          => true,
+				'lazy_load_term_meta'    => false,
+				'update_post_meta_cache' => false,
+				'update_post_term_cache' => false,
+			);
+
+			$the_query = new \WP_Query( $args );
```
westonruter, line 173: If `use WP_Query` is added to the top:

```suggestion
			$the_query = new WP_Query( $args );
```

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+
+			if ( $the_query->have_posts() ) {
+				while ( $the_query->have_posts() ) {
+					$the_query->the_post();
+
+					$urls[]         = get_permalink();
+					$post           = get_post();
+					$taxonomy_names = get_post_taxonomies( $post );
```
westonruter, line 181:
```suggestion
					$the_query->next_post();
					$urls[]         = get_permalink( $the_query->post );
					$taxonomy_names = get_post_taxonomies( $the_query->post );
```

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+					sprintf(
+						/* translators: %s: The Post Type name. */
+						__( 'Unable to retrieve post URL for post type: %s', 'plugin-check' ),
+						$post_type
+					)
+				);
+			}
+
+			/* Restore original Post Data */
+			wp_reset_postdata();
```
westonruter, line 210: Now unneeded if we don't call `the_post()` above.

```suggestion
```

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
```diff
@@ -0,0 +1,276 @@
...
+			}
+		);
+
+		if ( ! empty( $this->plugin_scripts ) ) {
+			$url_count = count( $urls );
+			foreach ( $this->plugin_scripts as $plugin_script ) {
+				if ( isset( $plugin_script['count'] ) && ( $url_count === $plugin_script['count'] ) ) {
+					$this->add_result_warning_for_file(
+						$result,
+						__( 'This script is being loaded in all contexts.', 'plugin-check' ),
```
westonruter, line 141: Why is it a problem if a script is loaded in all contexts?

swissspidy:
@westonruter Just noting that I agree with the suggestions, but since they would need to be done in `Enqueued_Styles_Scope_Check` too, might be best to do that in a follow-up PR converging the two classes.

includes/Checker/Checks/Enqueued_Scripts_Scope_Check.php
swissspidy: It's not _always_ a problem, but often times this is done in error and not actually on purpose. This can also be often observed in the wp-admin, when a plugin loads JS on every page instead of only its own settings page.

This check is aimed to detect the former, but of course it can't be bulletproof. Hence this is a _warning_ and not an _error_.

westonruter (approved):

mukeshpanchal27 (approved):

PR #490, opened 2024-07-03
Allow static checks to be run against an arbitrary path

### LINKED ISSUE #478: Allow static checks to be run against an arbitrary path

davidperezgar:
What we need for the [submission form from wordpress.org](https://wordpress.org/plugins/developers/add/) is run static checks in an arbitrary path. It's helpful as well for any developer user to directly run static checks in any directory.

Resolves #478

### DIFF
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -365,7 +365,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		}
 
 		// Check the readme file Stable tag against the plugin's main file version.
-		$plugin_data = get_plugin_data( WP_PLUGIN_DIR . '/' . $result->plugin()->basename() );
+		$plugin_data = get_plugin_data( $result->plugin()->main_file() );
 
 		if (
 			! empty( $plugin_data['Version'] ) &&
--- a/includes/Checker/Checks/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Readme_Check.php
@@ -225,7 +225,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	private function check_license( Check_Result $result, string $readme_file, Parser $parser ) {
 		$license          = $parser->license;
 		$matches_license  = array();
-		$plugin_main_file = WP_PLUGIN_DIR . '/' . $result->plugin()->basename();
+		$plugin_main_file = $result->plugin()->main_file();
 
 		// Filter the readme files.
 		if ( empty( $license ) ) {
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -83,7 +83,7 @@ class Plugin_Context {
 	/**
 	 * Returns the plugin main file.
 	 *
-	 * @since 1.0.0
+	 * @since 1.0.2
 	 *
 	 * @return string Plugin main file.
 	 */
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -70,6 +70,26 @@ class Plugin_Context {
 	}
 
 	/**
+	 * Returns the plugin main file.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @return string Plugin main file.
+	 */
+	public function main_file() {
+		if ( false === strpos( $this->main_file, '.php' ) ) {
+			$files = glob( $this->main_file . '/*.php' );
+			foreach ( $files as $file ) {
+				$plugin_data = get_plugin_data( $file );
+				if ( ! empty( $plugin_data['Name'] ) ) {
+					return $file;
+				}
+			}
+		}
+		return $this->main_file;
+	}
+
+	/**
 	 * Returns the absolute path for a relative path to the plugin directory.
 	 *
 	 * @since 1.0.0
--- a/includes/Plugin_Context.php
+++ b/includes/Plugin_Context.php
@@ -78,7 +78,11 @@ class Plugin_Context {
 	 * @return string Absolute path.
 	 */
 	public function path( $relative_path = '/' ) {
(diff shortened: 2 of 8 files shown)


swissspidy:
I'm out of office until Monday but happy to review then.

At first glance I'd say the `is_dir` logic is at the wrong level. If you pull it higher up, then you don't to worry about naming or anything. For example doing that in `Abstract_Check_Runner::get_plugin_basename` would be better. No need to touch `get_plugin_basename_from_input()` at all.

Also, this would be an excellent candidate to write Behat tests for.

Happy to help with both once I'm back.

ernilambar:
I tested PR and it is not working for plugin in external location as expected.
Few checks are working though.

In Plugin_Context there is `protected $main_file;` which is used in several checks. This expects the main file of the plugin but in external plugin, this variable just has plugin path. So there are few errors in the output.

Screenshot 2024-07-04 at 4 51 43 PM

Like in `Plugin_Header_Text_Domain_Check`, there is like this:
```
$plugin_main_file = WP_PLUGIN_DIR . '/' . $result->plugin()->basename();
```

May be we should fix `$main_file` in the beginning and access same property in all checks rather than evaluating main file like above.

davidperezgar:
Thanks @ernilambar for your feedback. I've detected that problem, and I've committed a fix. Update the PR and it should work.

includes/Plugin_Context.php
```diff
@@ -69,6 +69,26 @@ public function basename() {
...
 
+	/**
+	 * Returns the plugin main file.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @return string Plugin main file.
+	 */
+	public function main_file() {
+		if ( false === strpos( $this->main_file, '.php' ) ) {
```
ernilambar: Finding actual main file here in this method is too late. May be it should be done in the constructor of `Plugin_Context`.

includes/Checker/Checks/Plugin_Readme_Check.php
```diff
@@ -225,7 +225,7 @@ private function check_default_text( Check_Result $result, string $readme_file,
 	private function check_license( Check_Result $result, string $readme_file, Parser $parser ) {
 		$license          = $parser->license;
 		$matches_license  = array();
-		$plugin_main_file = WP_PLUGIN_DIR . '/' . $result->plugin()->basename();
+		$plugin_main_file = $result->plugin()->main_file();
```
ernilambar, line 231: Similar update should be done in following files also.
- Plugin_Header_Text_Domain_Check
- Plugin_Readme_Check
- Plugin_Updater_Check
- Trademarks_Check

davidperezgar: Yes, you're right. I'll check it out

includes/Plugin_Context.php
davidperezgar: ok, let's do it.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -40,12 +40,16 @@ class Plugin_Request_Utility {
...
-	 * @param string $plugin_slug The plugin slug or basename.
+	 * @param string $plugin_slug The plugin slug, basename or arbitrary path.
 	 * @return string The plugin basename.
 	 *
 	 * @throws Exception Thrown if an invalid basename or plugin slug is provided.
 	 */
 	public static function get_plugin_basename_from_input( $plugin_slug ) {
+		if ( is_dir( $plugin_slug ) ) {
+			return $plugin_slug;
+		}
```
swissspidy: We should not need to change this method.

This change should happen in `Abstract_Check_Runner::get_plugin_basename()`

davidperezgar: I don't understand. This conditional jump the test if the plugin is activated and get actived plugins, that is necessary.

swissspidy: This conditional should be done in `Abstract_Check_Runner::get_plugin_basename()`, not here.

ernilambar: May be like this in `Abstract_Check_Runner::get_plugin_basename()`:
```
			if ( is_dir( $plugin ) ) {
			} else {
				$this->plugin_basename = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin );
			}

```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -40,7 +40,7 @@ class Plugin_Request_Utility {
 	 *
 	 * @since 1.0.0
 	 *
-	 * @param string $plugin_slug The plugin slug or basename.
+	 * @param string $plugin_slug The plugin slug, basename or arbitrary path.
```
swissspidy:
```suggestion
	 * @param string $plugin_slug The plugin slug or basename.
```

swissspidy (approved):

ernilambar (approved):

includes/Plugin_Context.php
```diff
@@ -69,6 +80,17 @@ public function basename() {
 		return plugin_basename( $this->main_file );
 	}
 
+	/**
+	 * Returns the plugin main file.
+	 *
+	 * @since 1.0.0
```
```suggestion
	 * @since 1.0.2
```

davidperezgar:
None, it's working

PR #508, opened 2024-07-08
Make default selected check category filterable

ernilambar:
Related https://github.com/WordPress/plugin-check/issues/494

* Introduce filter `wp_plugin_check_default_categories`
* Add new method `get_default_check_categories_to_be_selected()` in `Admin_Page` class
* Add test for related changes

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -310,7 +310,7 @@ final class Admin_Page {
 	 *
 	 * @return string Categories separated by double underscores.
 	 */
-	public static function get_default_check_categories_to_be_selected() {
+	public function get_default_check_categories_to_be_selected() {
 		$default_check_categories = array(
 			'plugin_repo',
 		);
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -196,8 +196,8 @@ final class Admin_Page {
 
 		$category_labels = Check_Categories::get_category_labels();
 
-		// Get user settings for category preferences and set a default value to check plugin_repo by default.
-		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'plugin_repo' );
+		// Get user settings for category preferences.
+		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', $this->get_default_check_categories_to_be_selected() );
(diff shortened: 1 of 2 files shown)


includes/Admin/Admin_Page.php
```diff
@@ -300,4 +302,28 @@ public function admin_footer() {
...
 	}
+
+	/**
+	 * Gets default check categories to be selected.
+	 *
+	 * @since 1.0.2
+	 *
+	 * @return string Categories separated by double underscores.
+	 */
+	public function get_default_check_categories_to_be_selected() {
```
swissspidy: No need for this to be public

ernilambar: Yah, I understood. But when I kept private/protected I could not implement unit test for it as it gave fatal error. Any suggestions?

swissspidy: We could test it via `render_page()` (like the `test_render_page` tests)

includes/Admin/Admin_Page.php
```diff
@@ -300,4 +302,28 @@ public function admin_footer() {
...
+		$default_check_categories = array(
+			'plugin_repo',
+		);
+
+		/**
+		 * Filters the default check categories to be selected.
+		 *
+		 * @since 1.0.2
+		 *
+		 * @param array $default_check_categories An array of categories.
```
swissspidy:
```suggestion
		 * @param string[] $default_check_categories An array of category slugs.
```

includes/Admin/Admin_Page.php
```diff
@@ -300,4 +302,28 @@ public function admin_footer() {
...
+		 */
+		$default_categories = (array) apply_filters( 'wp_plugin_check_default_categories', $default_check_categories );
+
+		return ! empty( $default_categories ) ? implode( '__', $default_categories ) : '';
```
swissspidy: No need for the `empty` check. `implode` already returns an empty string when the array is empty.

swissspidy: I don't like doing the `implode` here. Let's do it above.

```php
$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', implode( '__', $this->get_default_check_categories_to_be_selected() ) );

swissspidy (approved):

mukeshpanchal27 (approved):

PR #509, opened 2024-07-08
Allow registering custom categories through filter

### LINKED ISSUE #494: Default selected check category could be customizable using filter
Currently default selected category is set to `plugin_repo`. It would be great if we could provide filter for that so that I can assign my custom category/categories as default.

```
$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'plugin_repo' );
```

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/494

* Introduce filter wp_plugin_check_categories
* Add new method `get_check_categories()` in `Check_Categories` class
* Add test for related changes

Example implementation: https://github.com/ernilambar/fresh-check/blob/main/fresh-check.php

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -194,7 +194,7 @@ final class Admin_Page {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$category_labels = Check_Categories::get_category_labels();
+		$category_labels = Check_Categories::get_categories();
 
 		// Get user settings for category preferences and set a default value to check plugin_repo by default.
 		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'plugin_repo' );
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -194,7 +194,7 @@ final class Admin_Page {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$category_labels = Check_Categories::get_categories();
+		$categories = Check_Categories::get_categories();
 
 		// Get user settings for category preferences and set a default value to check plugin_repo by default.
 		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', 'plugin_repo' );
--- a/includes/Checker/Check_Categories.php
+++ b/includes/Checker/Check_Categories.php
@@ -64,7 +64,7 @@ class Check_Categories {
 		 *
 		 * @since 1.0.2
 		 *
-		 * @param array $default_categories An array of check categories.
+		 * @param array<string, string> $default_categories Associative array of category slugs to labels.
 		 */
 		$check_categories = (array) apply_filters( 'wp_plugin_check_categories', $default_categories );
 
--- a/includes/Checker/Check_Categories.php
+++ b/includes/Checker/Check_Categories.php
@@ -29,13 +29,7 @@ class Check_Categories {
 	 * @return array An array of available categories.
 	 */
 	public static function get_categories() {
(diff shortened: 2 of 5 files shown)


includes/Checker/Check_Categories.php
```diff
@@ -46,13 +40,35 @@ public static function get_categories() {
...
 			self::CATEGORY_PERFORMANCE   => __( 'Performance', 'plugin-check' ),
 			self::CATEGORY_ACCESSIBILITY => __( 'Accessibility', 'plugin-check' ),
 		);
+
+		/**
+		 * Filters the check categories.
+		 *
+		 * @since 1.0.2
+		 *
+		 * @param array $default_categories An array of check categories.
```
swissspidy: Nit: let’s explain that it‘s an associative array

includes/Checker/Check_Categories.php
```diff
@@ -46,13 +40,35 @@ public static function get_categories() {
...
+	}
+
+	/**
+	 * Returns an array of check categories.
+	 *
+	 * @since 1.0.2
+	 *
+	 * @return array An array of check categories.
+	 */
+	public static function get_check_categories() {
```
swissspidy: Confusing to have get_check_categories and get_categories methods. Can we rename them?

includes/Checker/Check_Categories.php
ernilambar: Sure. Would it make more sense following array structure like `block_categories_all`? It would be more future proof and we can easily add attributes for each category in the future if needed.

swissspidy: I don't think we need that. If we ever need more complex data structure we can still add it later.

What I mean here is something like 
```suggestion
		 * @param array<string, string> $default_categories Associative array of category slugs to labels.
```

includes/Checker/Check_Categories.php
swissspidy: * `get_categories`
* `get_category_slugs`
* `get_category_labels`

includes/Checker/Check_Categories.php
```diff
@@ -46,13 +40,35 @@ public static function get_categories() {
 	 * @return array An array of category labels.
 	 */
 	public static function get_category_labels() {
-		return array(
+		return self::get_categories();
```
swissspidy:
```suggestion
		return array_values( self::get_categories() );
```

swissspidy (approved):

ernilambar:
Looks like we have used key/value pair, like:

```
$category_labels  = $check_categories->get_category_labels();

$categories = array();

foreach ( $category_labels as $slug => $label ) {
	$categories[] = array(
		'slug' => $slug,
		'name' => $label,
	);
}
```

So I think we need to adjust in other places also.

swissspidy:
Then these places should call `get_categories()`

includes/Admin/Admin_Page.php
```diff
@@ -194,7 +194,7 @@ public function render_page() {
 
 		$selected_plugin_basename = filter_input( INPUT_GET, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		$category_labels = Check_Categories::get_category_labels();
+		$category_labels = Check_Categories::get_categories();
```
mukeshpanchal27: Update the variable to `$categories` and same `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'templates/admin-page.php'`

mukeshpanchal27:
In https://github.com/WordPress/plugin-check/pull/509/commits/37aeb20e90194881d8b5cb497445866d8be5ada2, I have removed the `test_get_category_labels` test as they test the same `get_categories()` data values, making them redundant.

ernilambar:
May be we could remove that labels function. It is not used anywhere I guess.

mukeshpanchal27:
Yes but that function introduce in 1.0, leave it for now.

mukeshpanchal27 (approved):

PR #510, opened 2024-07-08
Revert "Update yoast/phpunit-polyfills requirement from ^1.0 to ^2.0 in /build-phpunit"

mukeshpanchal27:
Reverts WordPress/plugin-check#497

After the merge the unit tests goes failed so need some more debugging.

### DIFF
--- a/build-phpunit/composer.json
+++ b/build-phpunit/composer.json
@@ -1,6 +1,6 @@
 {
   "require-dev": {
     "wp-phpunit/wp-phpunit": "^6.1",
-    "yoast/phpunit-polyfills": "^2.0"
+    "yoast/phpunit-polyfills": "^1.0"
   }
 }


swissspidy (approved):

swissspidy:
Confused, where are they failing? 🤔

swissspidy:
Ah, I see one failure now regarding the `--verbose` option which was removed in PHPUnit 10. 

Let's just remove this argument then? 🤔

mukeshpanchal27:
I did that in local system and got similar error shown here https://github.com/WordPress/plugin-check/actions/runs/9837501136/job/27155469465?pr=512

swissspidy:
Yeah seeing it now too. Just strange that a PR like this one does not trigger unit tests, otherwise we would have caught this before merging.

PR #518, opened 2024-07-11
Add feature tests for checking plugin with addon enabled

ernilambar:
* Add feature tests with addon enabled (which includes new category and extra checks under that category)


ernilambar:
Note: Behat tests are passing in my local setup with following specs.
```
PHP: 8.2.19
Node: 20.11.1
```

ernilambar:
Problem I am facing is classes on main plugin are not loading in the addon. Since WordPress not necessarily loads plugin files in our required order, how can I make sure that all classes of main plugin are available in the addon?

CC @swissspidy

ernilambar:
May be we should also add instructions in our docs on how to create addon for Plugin Check.

swissspidy:
Yeah simply defining the classes like that would be error prone.

First, you'll want to use the `Requires Plugins` header.

Second, I would recommending moving the classes to separate files and then load them only when needed and when you know the parent classes exist. For example like so:

```php
add_filter(
        'wp_plugin_check_checks',
        function ( array $checks ) {
          require_once './class-prohibited-text-check.php';
          require_once './class-postsperpage-check.php';

          return array_merge(
            $checks,
            array(
              'prohibited_text' => new Prohibited_Text_Check(),
              'postsperpage'    => new PostsPerPage_Check(),
            )
          );
        }
      );
```

swissspidy:
> May be we should also add instructions in our docs on how to create addon for Plugin Check.

Yeah @davidperezgar brought that up too. First we can start small, explaining the filters and giving a short example. We can reuse some of the code you're writing here for that.

Not really urgent though, as I doubt there is huge demand for writing those 😄

davidperezgar:
Yes! That would be very helpful.

ernilambar:
When I `plugin check` command is run, may be there needs to cleaned up. Because when same command run twice, in the second time external checks are not detected. For example, in line 463 and 477 I have kept same command but output is not as expected.

swissspidy:
That sounds odd 🤔 Sounds like some debugging is in order

ernilambar:
@swissspidy I tried this to reproduce this in separate setup. After first command run, those checks are not included until I delete file `object-cache.php`.

swissspidy:
So it's an issue with Plugin Check / the `object-cache.php` drop-in and not with the test?

Want me to take a look at it to see if I can find the issue, or what's the next step?

ernilambar:
Yes, it does not seem to be related to Behat test. I tried to debug but could not find the exact issue. May be you can give a shot to find issue quicker as you are more familiar with the architecture 😊

swissspidy:
Well, I'm not _that_ familiar with that part 😅 

The gist is that`object-cache.php` is used as a "hook" to short-circuit regular WordPress loading order and manually execute the PCP CLI command before WordPress has fully loaded. That also means before any plugins have actually loaded...

From what I can see, this hack is used so we can perform runtime checks for a plugin without the plugin having to be active on the site. But not entirely sure.

When I activate `pcp-addon` and `foo-plugin` and remove all of this `object-cache.php` stuff, the runtime checks and the custom addon checks work just fine. See #533 for where I attempted this.

Related: #153

ernilambar:
I removed `before_wp_load` hook and all tests are passing. Did we miss to add test for CLI_Runner or it is not needed now?

swissspidy:
It's still needed and used by `Plugin_Check_Command`. Now it's just instantiated later and not during the object-cache setup. See also #533.

davidperezgar:
Behat tests are not passing.

ernilambar:
> Behat tests are not passing.

Yes. See https://github.com/WordPress/plugin-check/pull/518#issuecomment-2232943393 above and related conversation. That needs to be fixed first.

ernilambar:
I have created separate feature file to pass the feature test. For blocker, I have created separate issue for tracking. https://github.com/WordPress/plugin-check/issues/555

swissspidy:
But separating the tests like this now gives the false impression that it works, when in reality it actually doesn't (well, unless you delete the drop-in between runs).

The object-cache drop-in simply prevents add-ons from easily integrating with PCP, due to the change in execution/loading order. That blocks this PR.

davidperezgar:
So, is it ok?

swissspidy:
No, IMHO this PR is currently blocked because the feature it is testing is not actually working as expected. It is blocked by #555

cli.php
```diff
@@ -34,8 +34,17 @@
...
-WP_CLI::add_command( 'plugin', $plugin_command );
-
+WP_CLI::add_command(
+	'plugin',
+	$plugin_command,
+	array(
+		'after_invoke' => function () {
+			if ( file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+				unlink( ABSPATH . 'wp-content/object-cache.php' );
+			}
```
swissspidy: That sounds too easy to be true 😅

tests/behat/features/plugin-check.feature
```diff
@@ -374,3 +374,283 @@ Feature: Test that the WP-CLI command works.
...
+      <?xml version="1.0"?>
+      <ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="PCPAddon" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/squizlabs/PHP_CodeSniffer/master/phpcs.xsd">
+        <rule ref="WordPress.WP.PostsPerPage">
+          <type>error</type>
+          <severity>9</severity>
+        </rule>
+      </ruleset>
+      """
+    And I run the WP-CLI command `plugin activate pcp-addon`
+    And a wp-content/plugins/foo-sample/foo-sample.php file:
```
swissspidy, line 382: Can we add something that is flagged by a runtime check to this plugin? To verify that runtime checks are running as expected

cli.php
ernilambar: Yah, looks like it does not work :-D

ernilambar: Screenshot 2024-08-20 at 3 21 39 PM

ernilambar: But tests are passing in GH Action.

swissspidy: Same here. The "Perform runtime check for multi-file plugin" scenario is also failing locally.

When I manually go to the temp directory for that WP install and run the CLI command, the runtime checks are indeed not running.

The drop-in is successfully copied though.

So something is definitely not right 🤔

felixarntz: Can you point me to where in the Behat tests the failing call is defined? I'm not familiar with Behat _at all_, so not sure whether I'm looking at the right place. Is it the _second_ `plugin check` execution (in https://github.com/WordPress/plugin-check/pull/518/files#diff-303aba0b6bdf4d673c51a01f6bb7b91eccfc4717908349148e02927f9f3ac223R602) specifically that's failing? Or is even the first `plugin check` failing?

ernilambar: Yes. First command run good. When same command run again, issue appears.

felixarntz: Thanks. Can you also clarify how that relates to runtime checks? As far as I can tell, the addon checks are all non-runtime checks, so what does that have to do with runtime checks / setting up the `object-cache.php` drop-in? As far as I can see, the drop-in is only used in the very last execution, since only there the CLI command is run with `--require`ing of that `cli.php` file.

ernilambar: If that file is deleted manually then above issue does not appear. This is not specific to runtime checks.

felixarntz: But isn't that file only created _after_ the failing second run, in https://github.com/WordPress/plugin-check/pull/518/files#diff-303aba0b6bdf4d673c51a01f6bb7b91eccfc4717908349148e02927f9f3ac223R672? Since only there the `--require` is used, so only that should place the drop-in. 🤔

felixarntz:
@swissspidy @ernilambar I've been looking through the codebase a bit more to find potential problems, and so far I believe I found at least one, in `Abstract_Check_Runner::get_checks_to_run()`:
* There's a check `$this->initialized_early || $this->runtime_environment->can_set_up()`, based on which runtime checks are being used. This makes sense for the AJAX based workflow because:
    * Checks are being retrieved as a first step (before anything is set up).
    * Therefore we need to check whether the environment is already set up or, if not, whether it _can_ be set up.
        * The `$this->initialized_early` is basically whether the environment is already set up, since that can only be `true` if the `object-cache.php` drop-in is already present.
        * Even if `$this->initialized_early` is `false` at this point, it's fine as long as the environment can be set up, as in the AJAX workflow this happens in a separate request afterwards, but before any request that runs any checks.
* For the CLI based workflow, I believe this condition is wrong though. With CLI, it all runs in a single request so if the environment is not set up already (which requires the `--require` to place the `object-cache.php` drop-in early), we _can't_ do this later, since the CLI command runs after WordPress, plugins etc have already been loaded.

In other words, I think for CLI we either need to reduce this to just `$this->initialized_early`, or we go with `$this->initialized_early && $this->runtime_environment->can_set_up()` (which shouldn't make any difference in outcome, but may be better for extra safety).

I don't know if this bug is related to the failure here, but it's worth hacking it in temporarily to try. If that's it, I think we should open a separate PR to address the bug. I think a clean implementation would be to replace the current `$this->initialized_early || $this->runtime_environment->can_set_up()` check with a call to an abstract method like `can_use_runtime_checks()`, so that it can be implemented differently between the AJAX and CLI runners.

Let me know if you have any questions. FWIW I'm thinking it might be useful to make a flowchart of how this works in the different circumstances to make it easier to follow.

swissspidy:
> FWIW I'm thinking it might be useful to make a flowchart of how this works in the different circumstances to make it easier to follow.

Yeah as you can see, neither @ernilambar nor I are really familiar with that flow as it's not really documented and hard to follow. If someone from the original implementers group can do that, this would be helpful. 

tbh, I really wish all this magic wasn't needed, and that we could address #153 instead and can come up with a simpler solution for things like the GitHub Action where it's a controlled environment.

swissspidy:
> I've been looking through the codebase a bit more to find potential problems, and so far I believe I found at least one, in `Abstract_Check_Runner::get_checks_to_run()`: [...]

Could you perhaps make those changes against this PR yourself? That way it's easier to follow.

cli.php
```diff
@@ -34,8 +34,17 @@
...
 $plugin_command = new Plugin_Check_Command( $context );
-WP_CLI::add_command( 'plugin', $plugin_command );
-
+WP_CLI::add_command(
+	'plugin',
+	$plugin_command,
+	array(
+		'after_invoke' => function () {
+			if ( file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
+				unlink( ABSPATH . 'wp-content/object-cache.php' );
```
swissspidy: Note to self: We also need to check whether it's actually our object-cache drop-in

swissspidy:
I just pushed some changes to make the tests easier to reason about. 
Found a couple of issues along the way, like how the drop-in wasn't actually copied correctly because `ABSPATH` isn't defined `before_wp_load`, so I switched to `after_wp_config_load`.

Also included a runtime check in the example add-on so we can verify those work too (they don't).

We'd need to figure out how we could possibly load an add-on in CLI when doing runtime checks. Maybe the add-ons to load need to be explicitly passed as an arg or something?

felixarntz:
@swissspidy @ernilambar Please review #597, which is about documenting all the current process flows. While doing that, I identified a number of problems (mostly related about addon checks indeed). I believe with the additional context from that documentation (potentially after a few reviews to clarify things more), we can collaborate in finding solutions to these problems.

swissspidy:
Proposed next steps:

1. Update the PHP tests so that they pass again
	* Happens because I'm checking for the `WP_CLI` constant now
2. Merge the PR with some Behat tests commented out
3. Address https://github.com/WordPress/plugin-check/issues/597#issuecomment-2327483452 in follow-up PRs, re-enabling the commented out Behat tests

swissspidy (approved):

includes/Checker/CLI_Runner.php
```diff
@@ -32,12 +32,15 @@ class CLI_Runner extends Abstract_Check_Runner {
 	 * @return bool Returns true if is an CLI request for the plugin check else false.
 	 */
 	public static function is_plugin_check() {
+		if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
+			return false;
```
felixarntz, line 36: 👍

cli.php
```diff
@@ -34,16 +34,29 @@
 
 // Create the CLI command instance and add to WP CLI.
 $plugin_command = new Plugin_Check_Command( $context );
-WP_CLI::add_command( 'plugin', $plugin_command );
-
+WP_CLI::add_command(
+	'plugin',
+	$plugin_command,
+	array(
+		'after_invoke' => function () {
```
felixarntz, line 49: Can you add a comment above this to explain why it's here? For example reference the hook below, that this is a cleanup for.

cli.php
```diff
@@ -34,16 +34,29 @@
...
- * @since 1.0.0
+ * Runs after wp-config.php is loaded and thus ABSPATH is defined,
+ * but before any plugins are actually loaded.
  */
 WP_CLI::add_hook(
-	'before_wp_load',
+	'after_wp_config_load',
 	function () {
 		if ( CLI_Runner::is_plugin_check() ) {
 			if ( ! file_exists( ABSPATH . 'wp-content/object-cache.php' ) ) {
```
felixarntz, line 70: In principle, this should only be placed if runtime checks are run. If only static checks should be run against a plugin, this isn't needed. But I'm not sure it's reasonably possible to detect that from here.

FWIW at the moment users of WP-CLI have to know about the `--require` portion anyway, so because of that it's probably _okay_ to also expect of them to only include the `--require` if they want to run any runtime checks. Something to improve elsewhere, as we've already discussed a few times.

felixarntz (approved):
LGTM, just minor notes.

PR #519, opened 2024-07-11
Add `Non_Blocking_Scripts_Check`

### LINKED ISSUE #467: `Non_Blocking_Scripts_Check`: Warn about enqueued scripts that use neither `defer` nor `async`
As proposed by @felixarntz in https://github.com/WordPress/plugin-check/issues/443, this check would help encourage user to switch to non blocking enqueues.

Potentially we could also accept a blocking script as long as it's `$in_footer`. But it would definitely warn about a blocking in head script.

@swissspidy pointed out that The corresponding PHPCS sniff [does not yet support](https://github.com/WordPress/WordPress-Coding-Standards/issues/2420) the new $args param we added in 6.3. It was also indicated that the sniff might actually be removed because it's harder to detect. So probably needs to be a runtime check.

swissspidy:
This runtime check is pretty basic right now but does the following:

1. Test various URLs, similar to our other styles/scripts checks
2. Enqueues all scripts
3. Checks whether each enqueued script match the plugin (based on URL)
4. For matching scripts, checks whether they're enqueued in the footer or the header and if they have a loading strategy

Warnings are reported if either is the case:

a) The script is enqueued in the header, without async/defer -> suggest adding async/defer
b) The script is enqueued in the footer -> suggest adding async/defer

Fixes #467


includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+	protected function check_url( Check_Result $result, $url ) {
+		// Reset the WP_Scripts instance.
+		unset( $GLOBALS['wp_scripts'] );
+
+		// Run the 'wp_enqueue_script' action, wrapped in an output buffer in case of any callbacks printing scripts
+		// directly. This is discouraged, but some plugins or themes are still doing it.
+		ob_start();
+		wp_enqueue_scripts();
+		wp_scripts()->do_head_items();
+		wp_scripts()->do_footer_items();
```
adamsilverstein: what about scripts added to content or on the `wp_body_open` hook? should we try to catch those as well?

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+					$script_path
+				);
+			} else {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+					/* translators: 1: tested URL. 2: 'async'. 3: 'defer' */
+						__( 'This script on %1$s is loaded in the footer. Consider an %2$s or %3$s script loading strategy instead.', 'plugin-check' ),
+						$url,
+						'async',
```
adamsilverstein: i would suggest these in the reverse order" defer, async" since defer is likely the better choice for most cases

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+			if ( ! empty( $script->extra['strategy'] ) ) {
+				continue;
+			}
+
+			if ( ! $is_script_in_footer ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: 1: tested URL. 2: 'async'. 3: 'defer' */
+						__( 'This script on %1$s is potentially blocking. Consider an %2$s or %3$s script strategy or moving it to the footer', 'plugin-check' ),
```
adamsilverstein: I'm not sure we should recommend moving it to the footer. curious what others think on this point.

adamsilverstein (commented):
Looks good! only some small feedback and a question.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
swissspidy: Good question, not sure how feasible that is given our current runtime check setup and the various ways for enqueuing scripts.

Note that this applies to all our performance checks related to scripts and styles. They all do the same thing, so we would need to change all of them. This therefore might be better suited for a new issue.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation;
+use WordPress\Plugin_Check\Checker\With_Shared_Preparations;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+use WordPress\Plugin_Check\Traits\URL_Aware;
+
+/**
+ * Check for enqueued script sizes.
```
ernilambar: Doc should be updated for the class.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
westonruter: I think this is right. If a script has to be blocking, then moving it to the footer will prevent it from blocking rendering.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+	 */
+	protected function get_urls() {
+		$urls = array( home_url() );
+
+		foreach ( $this->get_viewable_post_types() as $post_type ) {
+			$posts = get_posts(
+				array(
+					'posts_per_page' => 1,
+					'post_type'      => $post_type,
+					'post_status'    => array( 'publish', 'inherit' ),
```
westonruter: Why inherit?

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+		// Reset the WP_Scripts instance.
+		unset( $GLOBALS['wp_scripts'] );
+
+		// Run the 'wp_enqueue_script' action, wrapped in an output buffer in case of any callbacks printing scripts
+		// directly. This is discouraged, but some plugins or themes are still doing it.
+		ob_start();
+		wp_enqueue_scripts();
+		wp_scripts()->do_head_items();
+		wp_scripts()->do_footer_items();
+		ob_get_clean();
```
westonruter: Since the return value isn't needed:

```suggestion
		ob_end_clean();
```

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,242 @@
...
+						'defer'
+					),
+					'NonBlockingScripts.BlockingHeadScript',
+					$script_path
+				);
+			} else {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+					/* translators: 1: tested URL. 2: 'async'. 3: 'defer' */
```
westonruter:
```suggestion
						/* translators: 1: tested URL. 2: 'async'. 3: 'defer' */
```

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
westonruter: Maybe this is what you're hinting at, but wouldn't a more ideal runtime check setup actually do a loopback request to the URL and examine the results of the frontend request?

swissspidy: Exactly, but Plugin Check doesn‘t do that right now

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
swissspidy: Not sure, to be honest, I copied this from the other checks.

swissspidy:
Worth noting that this has some overlap with the `WordPress.WP.EnqueuedResourceParameters.NotInFooter` PHPCS sniff which we also run. But as per https://github.com/WordPress/WordPress-Coding-Standards/issues/2420 it is limited and might even get removed.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,243 @@
+<?php
+/**
+ * Class Non_Blocking_Scripts_Check
```
mukeshpanchal27:
```suggestion
 * Class Non_Blocking_Scripts_Check.
```

Add end period per WPCS.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
mukeshpanchal27: In PR #76, we added a first run time check. The https://github.com/WordPress/plugin-check/pull/76/commits/b6e5e564aa9af66ba2d99b96cc648f803c93fd51 added the "inherit" post status. I don't think this check is necessary, and there has been no feedback or discussion on the PR regarding it. It appears to have been added by mistake, so we should remove it in a follow-up PR.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,243 @@
...
+			if ( ! empty( $script->extra['strategy'] ) ) {
+				continue;
+			}
+
+			if ( ! $is_script_in_footer ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: 1: tested URL. 2: 'defer'. 3: 'async' */
+						__( 'This script on %1$s is potentially blocking. Consider an %2$s or %3$s script strategy or moving it to the footer', 'plugin-check' ),
```
mukeshpanchal27:
```suggestion
						__( 'This script on %1$s is potentially blocking. Consider a %2$s or %3$s script strategy or moving it to the footer.', 'plugin-check' ),
```

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,243 @@
...
+					),
+					'NonBlockingScripts.BlockingHeadScript',
+					$script_path
+				);
+			} else {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: 1: tested URL. 2: 'defer'. 3: 'async' */
+						__( 'This script on %1$s is loaded in the footer. Consider an %2$s or %3$s script loading strategy instead.', 'plugin-check' ),
```
mukeshpanchal27:
```suggestion
						__( 'This script on %1$s is loaded in the footer. Consider a %2$s or %3$s script loading strategy instead.', 'plugin-check' ),
```

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,243 @@
...
+
+			$script_path = str_replace( $result->plugin()->url(), $result->plugin()->path(), $script->src );
+
+			$is_script_in_footer = in_array( $handle, wp_scripts()->in_footer, true );
+
+			if ( ! empty( $script->extra['strategy'] ) ) {
+				continue;
+			}
+
+			if ( ! $is_script_in_footer ) {
```
mukeshpanchal27:
```suggestion
			if ( ! empty( $script->extra['strategy'] ) ) {
				continue;
			}

			$script_path = str_replace( $result->plugin()->url(), $result->plugin()->path(), $script->src );

			if ( ! in_array( $handle, wp_scripts()->in_footer, true ) ) {
```

mukeshpanchal27 (commented):
Thanks for the PR! Overall looks good to me. Left some feedbacks.

includes/Checker/Checks/Non_Blocking_Scripts_Check.php
swissspidy: Opened https://github.com/WordPress/plugin-check/issues/526 to track this.

tests/phpunit/tests/Checker/Checks/Non_Blocking_Scripts_Check_Tests.php
```diff
@@ -0,0 +1,89 @@
...
+		$this->assertTrue( $test_get );
+		$this->assertTrue( $test_post );
+		$this->assertTrue( $test_server );
+		$this->assertSame( 'test_prepare', $test_globals );
+	}
+
+	public function test_run_with_warnings() {
+		require UNIT_TESTS_PLUGIN_DIR . 'test-plugin-non-blocking-scripts-check/load.php';
+
+		$check   = new Non_Blocking_Scripts_Check( 1 );
```
mukeshpanchal27:
```suggestion
		$check   = new Non_Blocking_Scripts_Check();
```

We don't have threshold value for the check.

mukeshpanchal27 (approved):
Approved. Left one non-blocking feedback.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,241 @@
...
+		ob_start();
+		wp_enqueue_scripts();
+		wp_scripts()->do_head_items();
+		wp_scripts()->do_footer_items();
+		ob_end_clean();
+
+		foreach ( wp_scripts()->done as $handle ) {
+			$script = wp_scripts()->registered[ $handle ];
+
+			// TODO: Somehow detect inline scripts added by the plugin that don't have a `src`.
```
westonruter, line 185: Is this referring to before/after inline scripts? Or to scripts that are manually printed with `wp_print_inline_script_tag()` or otherwise?

swissspidy: Both I suppose :-)

Right now the detection whether a script is coming from a plugin is very basic. Something like Origination + loopback requests would be more robust.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,241 @@
...
+			$posts = get_posts(
+				array(
+					'posts_per_page' => 1,
+					'post_type'      => $post_type,
+					'post_status'    => array( 'publish', 'inherit' ),
+				)
+			);
+
+			if ( ! isset( $posts[0] ) ) {
+				throw new Exception(
```
adamsilverstein, line 142: maybe just skip silently here, sicne a post type could have no posts?

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,241 @@
...
+	 * @since 1.1.0
+	 *
+	 * @return array List of URL strings (either full URLs or paths).
+	 *
+	 * @throws Exception Thrown when a post type URL cannot be retrieved.
+	 */
+	protected function get_urls() {
+		$urls = array( home_url() );
+
+		foreach ( $this->get_viewable_post_types() as $post_type ) {
```
adamsilverstein, line 132: maybe add a sanity limit here, check at most 5 or 10 post types, I can imagine some site has 10,000 post types and this breaks site health for them

adamsilverstein (commented):
overall looks good, left some small feedback

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
swissspidy: Hmm no other plugins are active during checks and we're checking only a specific plugin here, I doubt that a single plugin adds that many post types.

If we add a limit, then we'd need to add it to all the runtime checks, but I don't see a need for it.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
swissspidy: This runtime check creates exactly 1 post in every post type, so it is guaranteed to have one.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,241 @@
...
+
+		// Run the 'wp_enqueue_script' action, wrapped in an output buffer in case of any callbacks printing scripts
+		// directly. This is discouraged, but some plugins or themes are still doing it.
+		ob_start();
+		wp_enqueue_scripts();
+		wp_scripts()->do_head_items();
+		wp_scripts()->do_footer_items();
+		ob_end_clean();
+
+		foreach ( wp_scripts()->done as $handle ) {
```
westonruter, line 182: I just realized what may be a more robust way to handle this, which will also include support for inline scripts: use HTML Tag Processor to iterate over generated markup instead of iterating over the `done` handles. There could be one output buffer for the head and another for the footer. This would then handle cases where a plugin is adding the strategy attribute manually via the `script_loader_tag` filter. It would also flag cases where a script fails to get a strategy due to a blocking dependent script depending on an async/defer script.

But that can be done in a subsequent PR.

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
```diff
@@ -0,0 +1,241 @@
...
+						/* translators: 1: tested URL. 2: 'defer'. 3: 'async' */
+						__( 'This script on %1$s is loaded in the footer. Consider a %2$s or %3$s script loading strategy instead.', 'plugin-check' ),
+						$url,
+						'defer',
+						'async'
+					),
+					'NonBlockingScripts.NoStrategy',
+					$script_path
+				);
+			}
```
westonruter: Wouldn't it be particularly helpful for the script handle to be indicated?

```suggestion
			if ( ! in_array( $handle, wp_scripts()->in_footer, true ) ) {
				$this->add_result_warning_for_file(
					$result,
					sprintf(
						/* translators: 1: tested URL. 2: the script handle. 3: 'defer'. 4: 'async' */
						__( 'This script on %1$s (with handle %2$s) is potentially blocking. Consider a %3$s or %4$s script strategy or moving it to the footer.', 'plugin-check' ),
						$url,
						$handle,
						'defer',
						'async'
					),
					'NonBlockingScripts.BlockingHeadScript',
					$script_path
				);
			} else {
				$this->add_result_warning_for_file(
					$result,
					sprintf(
						/* translators: 1: tested URL. 2: the script handle. 3: 'defer'. 4: 'async' */
						__( 'This script on %1$s (with handle %2$s) is loaded in the footer. Consider a %3$s or %4$s script loading strategy instead.', 'plugin-check' ),
						$url,
						$handle,
						'defer',
						'async'
					),
					'NonBlockingScripts.NoStrategy',
					$script_path
				);
			}
```

westonruter (approved):
A suggestion, but LGTM

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
swissspidy: In the admin/CLI output the annotation will be under the script file name, isn't that enough context?

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
swissspidy: A new issue for this sounds like a good idea 👍

includes/Checker/Checks/Performance/Non_Blocking_Scripts_Check.php
westonruter: Yeah, but since the script handles are string literals that are used for registering and enqueueing scripts, where the src may not always be immediately apparent, it seems having the handle too would be useful context.

PR #521, opened 2024-07-12
Conditional display of edit links

### LINKED ISSUE #515: Hide "View in code editor" link for uneditable files
Errors like `hidden_files`, `application_detected`, `no_plugin_readme`, etc should not have **View in code editor** link as those files are not editable. Clicking on those links are now just going to error page.

Example:

Screenshot 2024-07-09 at 11 01 46 AM

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/515

* Display edit link only if file is editable


includes/Plugin_Context.php
```diff
@@ -183,4 +191,35 @@ public function minimum_supported_wp() {
...
 	}
+
+	/**
+	 * Gets the list of editable file names of the plugin.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return string[] Array of editable file names relative to the plugin root.
+	 */
+	public function editable_files() {
```
swissspidy: Hmm not sure we really need a list of all editable files this way.

What do you think about `public function is_file_editable( string $file )` instead?

ernilambar: Make this method `protected`/`private` and add new method `public function is_file_editable( string $file )`?

swissspidy: No, what I meant is **only** `is_file_editable()`. No `editable_files` method.

Just something like this:

```php
public function is_file_editable( string $filename ) {
	$editable_extensions = wp_get_plugin_file_editable_extensions( $this->basename() );

	$extension  = strtolower( pathinfo( $filename, PATHINFO_EXTENSION ) );

	return in_array( $extension, $editable_extensions, true );
}
```

ernilambar: Hmm, let me try this in my local setup.

ernilambar: I tested in local setup. This will fix half of the issue. This does not handle the case of `no_plugin_readme`. Here in this case, there is no actual `readme.txt` file in the plugin. But the value of `filename` is still `readme.txt`. So edit link is still displayed for this error.

swissspidy: So we can add a `file_exists` check?

ernilambar: PR updated.

swissspidy (approved):

includes/Plugin_Context.php
```diff
@@ -183,4 +183,20 @@ public function minimum_supported_wp() {
...
+
+	/**
+	 * Checks if the file is editable.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $file Filename.
+	 * @return bool true if the file is editable, otherwise false.
+	 */
+	public function is_file_editable( $file ) {
```
mukeshpanchal27, line 195: The function still allowed edit link for `.eslintrc.js` and `.readme.txt` files while the plugin editor didn't allow it.

The plugin editor use `get_plugin_files` function to get the list of files in which it use `list_files` function that they [exclude the hidden files](https://github.com/WordPress/wordpress-develop/blob/6.5/src/wp-admin/includes/file.php#L161C4-L164C5). We should incorporate those changes so it will not include hidden files.

```
// Skip hidden and excluded files.
if ( ( ! $include_hidden && '.' === $file[0] ) || in_array( $file, $exclusions, true ) ) {
	continue;
}
```

**Plugin check:**
Screenshot 2024-07-15 at 10 00 18 AM

**WP plugin editor:**
Screenshot 2024-07-15 at 10 00 37 AM

mukeshpanchal27 (changes requested):
Thanks @ernilambar for the PR. Left one feedback other then it looks good to me.

includes/Plugin_Context.php
ernilambar: PR updated.

mukeshpanchal27:
Thanks for the update.

It still adds links to the `.husky/_/husky.sh` file. When we click on the link, it shows "Sorry, that file cannot be edited."

@swissspidy, do we need to exclude the `.husky` folder? Is it worth checking the files from that folder? We already exclude some folders in PR #237.

ernilambar:
We are currently excluding filename starting with `.`. May be we should also exclude folders starting with `.` like core does.

mukeshpanchal27:
> We are currently excluding filename starting with `.`. May be we should also exclude folders starting with `.` like core does.

+1 That also works.

mukeshpanchal27 (approved):

PR #525, opened 2024-07-15
Strengthen readme finder to exclude invalid readme files

### LINKED ISSUE #524: Do not treat `.readme.txt` as valid readme file
If there is `.readme.txt` file in the plugin, readme related checks are run which should not be running.

Edit: Looks like those checks are triggered by `test.readme.txt` file also.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/524

* Make sure readme file has valid file name and extension.

### DIFF
--- a/includes/Traits/Find_Readme.php
+++ b/includes/Traits/Find_Readme.php
@@ -25,7 +25,7 @@ trait Find_Readme {
 	 */
 	protected function filter_files_for_readme( array $files, $plugin_relative_path ) {
 		// Find the readme file.
-		$readme_list = preg_grep( '/readme\.(txt|md)$/i', $files );
+		$readme_list = preg_grep( '/\/readme\.(txt|md)$/i', $files );
 
 		// Filter the readme files located at root.
 		$potential_readme_files = array_filter(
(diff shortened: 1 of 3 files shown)


includes/Traits/Find_Readme.php
```diff
@@ -32,7 +32,13 @@ protected function filter_files_for_readme( array $files, $plugin_relative_path
 			$readme_list,
 			function ( $file ) use ( $plugin_relative_path ) {
 				$file = str_replace( $plugin_relative_path, '', $file );
-				return ! str_contains( $file, '/' );
+
+				$info = pathinfo( $file );
```
swissspidy: Can't we just fix the regex above?

```diff
-$readme_list = preg_grep( '/readme\.(txt|md)$/i', $files );
+$readme_list = preg_grep( '/\/readme\.(txt|md)$/i', $files );
```

A test could help verify this.

ernilambar: Ok, I will check.

ernilambar: Updated and test added.

swissspidy (approved):

PR #533, opened 2024-07-17
Remove `object-cache.php` hack

swissspidy:
Found out in #518 that this `object-cache.php` drop-in prevents other plugins from adding custom checks that can then be run via CLI.

From what I can see,`object-cache.php` is used as a "hook" to short-circuit regular WordPress loading order and manually execute the PCP CLI command before WordPress has fully loaded. That also means before any plugins have actually loaded...

This "hack" appears to be used so we can perform runtime checks for a plugin without the plugin having to be active on the site. If we can simply make this a requirement for the plugin to be active, then we can simplify this _a lot_.

I do lack full context though on why this was implemented this way, so appreciate any feedback.

----

Related:

* #153


swissspidy:
@joemcgill @mukeshpanchal27 Do you perhaps more context on this whole early initialization thing with this `object-cache.php` drop-in? We could save ourselves some headaches by removing that...

mukeshpanchal27:
@swissspidy I found #42, where it was introduced and discussed why it was added. Hope this helps to get more context.

JJJ:
Due to the nature of how Plugin Check needs to work, only other way to load super early would be to `require` a stub from inside of `wp-config.php`, and do all the same work on either the `mu_plugins_loaded` or `shutdown` hooks, and that seems overly fragile & yucky considering Plugin Check is a core utility.

WordPress Core should just add support for a new drop-in: `wp-content/plugin-check.php`.

Plugin Check is the type of library that drop-in plugins are designed to accommodate; intentionally forked loading paths. Every existing drop-in overrides a low-level API or traps a related low-level problem, and it is an easy argument to make that Plugin Check has similarly low-level requirements.

Naming it something like I suggested above would also set the expectation to everyone that its usage is *reserved*, and if that file already exists it had better be compatible with Plugin Check, the same way all `db.php` drop-ins must extend `wpdb`.

swissspidy:
I like the `wp-content/plugin-check.php` suggestion, but... do we really need to load super early? 🤔  That's what I am trying to understand better.

ernilambar:
This issue came here also. https://github.com/WordPress/plugin-check/pull/543 😟

swissspidy:
Closing for now, can be revisited later

PR #536, opened 2024-07-18
Remove broad and time dependent tests in plugin readme

ernilambar:
Related https://github.com/WordPress/plugin-check/pull/483#issuecomment-2202589513

Since we have already separate test for each code, may be we don't need this type of broad tests like `without errors`.


swissspidy:
Let me guess, they are now failing because of `Tested up to: 6.5`?

Agree that these tests are too broad. Better to test for specific errors or absence thereof.

swissspidy (approved):

ernilambar:
> Let me guess, they are now failing because of `Tested up to: 6.5`?
> 
Yes

> Agree that these tests are too broad. Better to test for specific errors or absence thereof.

mukeshpanchal27 (approved):

PR #537, opened 2024-07-18
Add PHP 8.4 as experimental in tests

ernilambar:
Follow up to https://github.com/WordPress/plugin-check/pull/535


ernilambar:
Error:

```
 ERROR: docker.io/library/wordpress:php8.4: not found
```

swissspidy:
Yeah those images don't exist yet because 8.4 is still in alpha.

If we'd run our tests simply with `install-wp-tests.sh` then that wouldn't be a problem.

swissspidy:
Closing for now since this is blocked

PR #541, opened 2024-07-20
Making check Generic names

### LINKED ISSUE #523: Check: Generic function/class/define/option prefix names
This check aims to detect short or common prefixes that could cause fatal errors in WordPress installations. 
We consider as an error for this check.

## How could develop this check?

We need to have a white list of common function starts. Actually we have in our internal scanner: `__,_,-,set,get,is,save,show,update,add,wordpress,wp,woocommerce,wc,table,html,css,js,input,output,plugin,plugins,my_plugin,myplugin,prefix,my_custom,custom,as,widget,oauth2,handle,generate,post,site,remove,filter,display,init,start,check,sync,cache,phpmailer,declare,register,enable,include,search,upgrade,update,setup,create,admin,load,theme,fetch,apply,clear,verify,test,insert,acme,app,render,rest`

And after, We check the list of named functions that are outside a Class, and a list o named Classes. Maybe we can go to Namespaces as well.

Our description to developers:

> ## Generic function/class/define/namespace/option names 
>  
> All plugins must have unique function names, namespaces, defines, class and option names. This prevents your plugin from conflicting with other plugins or themes. We need you to update your plugin to use more unique and distinct names.
> 
> A good way to do this is with a prefix. For example, if your plugin is called "Easy Custom Post Types" then you could use names like these:
>  
> function ecpt_save_post()
> define( ‘ECPT_LICENSE’, true );
> class ECPT_Admin{}
> namespace EasyCustomPostTypes;
> update_option( 'ecpt_settings', $settings );
>  
> Don't try to use two (2) or three (3) letter prefixes anymore. We host nearly 100-thousand plugins on WordPress.org alone. There are tens of thousands more outside our servers. Believe us, you’re going to run into conflicts. 
>  
> You also need to avoid the use of __ (double underscores), wp_ , or _ (single underscore) as a prefix. Those are reserved for WordPress itself. You can use them inside your classes, but not as stand-alone function. 
> 
> Please remember, if you're using _n() or __() for translation, that's fine. We're only talking about functions you've created for your plugin, not the core functions from WordPress. In fact, those core features are why you need to not use those prefixes in your own plugin! You don't want to break WordPress for your users.
> 
> Related to this, using if (!function_exists(‘NAME ‘)) { around all your functions and classes sounds like a great idea until you realize the fatal flaw. If something else has a function with the same name and their code loads first, your plugin will break. Using if-exists should be reserved for shared libraries only.
> 
> Remember: Good prefix names are unique and distinct to your plugin. This will help you and the next person in debugging, as well as prevent conflicts. 
>  

davidperezgar:
Related #523


ernilambar:
I dont think we can use `WordPress.NamingConventions.PrefixAllGlobals` sniff directly here for our purpose. This sniff requires an array of prefixes to check against. If not passed this sniff does not run.

```
<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
	<properties>
		<property name="prefixes" type="array">
			<element value="ecpt"/>
			<element value="EasyCustomPostTypes"/>
		</property>
	</properties>
</rule>
```

davidperezgar:
Yes, you're right, so we should make it manually without PHPCS? then?

davidperezgar:
So, we have to make it manual. I'd need if to have this function: From the line in the file, detect the **function name** and **the arguments**.

I've reviewed and I don't find it. Any guidance?

davidperezgar:
How could I run the WPCS?

davidperezgar:
Related WPCS [#2481](https://github.com/WordPress/WordPress-Coding-Standards/issues/2481)

davidperezgar:
As @frantorres told me, this check has to be done by PCP, as WPCS won't be able to check all files and prefixes of the plugin. PCP does. Let's make it with a custom sniff.

tests/phpunit/tests/Checker/Checks/Prefix_Globals_Check_Tests.php
swissspidy, line 1: This file seems obsolete now? `Prefix_All_Globals_Check` does not exist.

phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/GenericNamesUnitTest.inc
swissspidy, line 1: You'll also need a `GenericNamesUnitTest.php` test file

tests/phpunit/tests/Checker/Checks/Prefix_Globals_Check_Tests.php
davidperezgar: yes!

phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/GenericNamesUnitTest.inc
davidperezgar: yes, it's on draft... I'm making it.. thanks

PR #542, opened 2024-07-21
CLI: Support checking a plugin from a URL

### LINKED ISSUE #476: Make enhacements for Plugin Team to use it as a tool
We do have an internal scanner, and we could develop our checks to this Plugin, but It would be a good idea to make enhancements to make it available to be used as a tool. We will necessary have internal checks that maybe in the future will be in an Internal Addon.

What we would need for using as a tool:
- The use of WP CLI need to have an option to include an URL, so it can be downloaded to the WordPress installation and run the checker after. This URL has a second part to get API information about the plugin and the user.
- After the process, we need a report based in HTML that will have all issues and explanations about how to solve them. It have to be opened in a browser.

I don't see more functionalities necessary for the team. I ping them, so they can add more ideas in this issue.

@WordPress/plugin-review-team

davidperezgar:
Related #476


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -486,6 +486,9 @@ final public function get_plugin_basename() {
 			if ( is_dir( $plugin ) ) {
 				$this->plugin_basename = $plugin;
 			} else {
+				if ( filter_var( $plugin, FILTER_VALIDATE_URL ) ) {
+					$plugin = Plugin_Request_Utility::download_plugin( $plugin );
```
ernilambar: `$plugin` value could be a string or boolean. May be we should handle case of false value also.

ernilambar:
I tested PR in my local setup also. Check is working correctly for external zip file like:

```
wp plugin check https://github.com/ernilambar/foo-bar-wp/releases/latest/download/foo-bar-wp.zip
```

includes/Checker/Abstract_Check_Runner.php
davidperezgar: If it's false, won't be a url. So, I suppose there is no problem.

ernilambar: While testing I gave valid URL but file system could extract the zip so returned false in that case. Another case, even if zip is extracted, it could not move folder using move_dir().

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -181,4 +181,79 @@ public static function get_files_to_ignore() {
...
+
+	/**
+	 * Downloads the plugin from the URL and installs it.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string|bool The plugin basename if the plugin is downloaded and installed successfully, false otherwise.
+	 */
+	public static function download_plugin( $plugin_url ) {
```
swissspidy, line 197: Could this use the `download_url()` function provided by WordPress? And then maybe `wp_handle_sideload()` with the custom upload dir. This would allow removing a lot of code here and we wouldn't have to worry about `WP_Filesystem` etc.

davidperezgar: I didn't know that function. For sure, I'll check it and propose a change.

davidperezgar: It's giving me a fatal error download_url, even if I use the require...

davidperezgar: PHP Fatal error:  Uncaught Error: Call to undefined function WordPress\Plugin_Check\Utilities\download_url() in /wp-content/plugins/plugin-check/includes/Utilities/Plugin_Request_Utility.php:194

ernilambar: @davidperezgar I am checking it.

includes/Checker/Abstract_Check_Runner.php
ernilambar: I have drafted a PR to use download_dir() and directly unzip file to the target destination. You can take reference from the PR. https://github.com/WordPress/plugin-check/pull/543/files

ernilambar:
One query here. Why are we copying plugin files to the `/plugins/` folder? When I run command like this:

```
wp plugin check https://github.com/ernilambar/foo-bar-wp/releases/latest/download/foo-bar-wp.zip
```
I would expect command to check the zip file. In this case I would not expect to have that plugin inside my  `/plugins/` folder. 
Since we now support running plugin from arbitrary folder, may be we can explore to use that approach. 
Thoughts?

davidperezgar:
Yes. You're right. Why should we run in plugins folder. Let's keep it in the temporary folder then.

ernilambar:
I have created PR to accept plugin zip URL as parameter. https://github.com/WordPress/plugin-check/pull/546 After this is merged we can rebase this PR to use that feature here.

includes/Checker/Abstract_Check_Runner.php
davidperezgar: I think that is better to have in one PR. I'm confused to what to do, merge or keep one?

davidperezgar:
Should we use the plugins folder as a target? 
Well, this method is mainly for using for the Plugin Review Team. It could be helpful to have it as a plugin. What if exists the folder? Should we delete it?

ernilambar:
Using remote zip URL in the command is very common in WP-CLI and it is frequently used feature. So here also, I think first step could be supporting remote zip URL in the command. After we finalize where to download zip, where to keep the plugin, how to run checks, etc, implementing custom URL targeted for Plugin Review would be just adding conditionals for `#wporgapi:` parameter.

davidperezgar:
Ok, I'm agree.

davidperezgar:
I updated the code from @ernilambar 's PR and changed the temp directory. I think that this is a good a approach.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,89 @@ public static function get_files_to_ignore() {
...
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string The plugin basename after downloading and installing the plugin.
+	 *
+	 * @throws Exception Thrown if an invalid URL given or zip could be extracted properly.
+	 */
+	public static function download_plugin( $plugin_url ) {
+		$response = wp_remote_get( $plugin_url );
+
+		if ( is_wp_error( $response ) || 200 !== wp_remote_retrieve_response_code( $response ) ) {
+			throw new Exception( 'Downloading the zip file failed.' );
```
mukeshpanchal27:
```suggestion
			throw new Exception( 
				__( 'Downloading the zip file failed.', 'plugin-check' )
			);
```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,91 @@ public static function get_files_to_ignore() {
...
+		$unzip_file = unzip_file( $file_path, $temp_dir );
+
+		if ( true !== $unzip_file ) {
+			throw new Exception( $unzip_file->get_error_message() );
+		}
+
+		// Remove zip file.
+		unlink( $file_path );
+
+		if ( ! empty( $plugin_info_url ) ) {
```
swissspidy: Would be good to validate this URL before using, maybe also with `filter_var( $plugin_info_url, FILTER_VALIDATE_URL )`

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,91 @@ public static function get_files_to_ignore() {
...
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string The plugin basename after downloading and installing the plugin.
+	 *
+	 * @throws Exception Thrown if an invalid URL given or zip could be extracted properly.
+	 */
+	public static function download_plugin( $plugin_url ) {
+		$response = wp_remote_get( $plugin_url );
```
swissspidy: Should we use `wp_safe_remote_get`?

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,91 @@ public static function get_files_to_ignore() {
...
+
+		if ( true !== $unzip_file ) {
+			throw new Exception( $unzip_file->get_error_message() );
+		}
+
+		// Remove zip file.
+		unlink( $file_path );
+
+		if ( ! empty( $plugin_info_url ) ) {
+			$response_json = wp_remote_get( $plugin_info_url );
```
swissspidy: Should we use `wp_safe_remote_get`?

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,91 @@ public static function get_files_to_ignore() {
...
+			$file_path     = $plugin_check_dir . 'plugin-info.json';
+			$file_process  = fopen( $file_path, 'w' );
+			fwrite( $file_process, $response_json['body'] );
```
swissspidy: We'd want to validate the response code first and whether it is not a `WP_Error`. The body can also be access via `wp_remote_retrieve_body`.

Then I'd recommend validating whether it is valid JSON.

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -182,6 +182,91 @@ public static function get_files_to_ignore() {
...
+		$file_process = fopen( $file_path, 'w' );
+		fwrite( $file_process, $response['body'] );
+		fclose( $file_process );
+
+		$temp_dir = $plugin_check_dir . strtotime( 'now' ) . '/';
+
+		// Unzip file.
+		require_once ABSPATH . 'wp-admin/includes/file.php';
+
+		WP_Filesystem();
```
swissspidy: This doesn't seem to be used at all.

Either we should use the filesystem API or `fwrite` & co, but not both.

ernilambar: I think `unzip_file()` requires this call. https://developer.wordpress.org/reference/functions/unzip_file/

swissspidy: Then we should use the filesystem API methods instead of `fwrite` :)

ernilambar:
@davidperezgar I have updated the method and added feature tests. Please check if `plugin-info.json` is correctly positioned.

davidperezgar:
Thanks @ernilambar . For me looks great.

I have a question. Should we clean this folder? Should we explain in the readme this folder?

ernilambar:
@davidperezgar Ideally I think it should be deleted. When I just try to check a remote ZIP in my CLI, I don't want those files under my WP system. That is why I was proposing using `get_temp_dir()` instead of upload directory, so that we wont have to worry about deleting ourself. Every OS handles and cleans up `tmp/` directory itself. But I am not sure how reliable is `get_temp_dir()` as I see its usage very less.

CC @swissspidy

swissspidy:
> But I am not sure how reliable is get_temp_dir() as I see its usage very less.

Nothing wrong with `get_temp_dir()` IMHO, WP definitely uses that a lot too.

> That is why I was proposing using `get_temp_dir()` instead of upload directory, so that we wont have to worry about deleting ourself. Every OS handles and cleans up `tmp/` directory itself.

IMHO we should still clean it up ourselves, even if it's in a temporary directory.

ernilambar:
@swissspidy I have updated the location to use `get_temp_dir()`. Where is the best place to keep the code for deleting the temp folder **/tmp/plugin-check/**? Any suggestions?

swissspidy:
Good question. Somewhere in `Abstract_Check_Runner` makes the most sense, as that's where the download is initiated. I'll push a quick change with what I have in mind.

davidperezgar:
For me is looking great. I've added timeout, because it was failing during the download from wporg. It're ready to be merged. Great work guys!

ernilambar (approved):

swissspidy (approved):

PR #543, opened 2024-07-24
WIP: Update URL implementation

ernilambar:


davidperezgar:
Is it works for you? It gives me a fatal error for download_url.

ernilambar:
Yes. Behat test is running fine with this.

davidperezgar:
But test it with wpcli

ernilambar:
Remove `object-cache.php` file from `/wp-content/` and it will work.  
Related https://github.com/WordPress/plugin-check/pull/533

swissspidy:
> Remove `object-cache.php` file from `/wp-content/` and it will work. Related #533

Let me guess, the command was loaded too early because of `object-cache.php` so the `download_url` function didn't exist yet? You might want to move up the `require_once ABSPATH . 'wp-admin/includes/file.php';` call in that case.

ernilambar:
Moving it on top `download_url()` is found but error comes: `wp_generate_password()` not found. Keeping `require_once ABSPATH . 'wp-includes/pluggable.php';` then error comes:

```
Uncaught Error: Undefined constant "AUTH_COOKIE" in /var/www/html/wp-includes/pluggable.php:932
```

PR #546, opened 2024-07-24
Accept plugin zip URL in check command

ernilambar:
* Zip is downloaded and kept in temp folder
* Check will run from the temp folder

### DIFF
--- a/includes/Utilities/Plugin_Request_Utility.php
+++ b/includes/Utilities/Plugin_Request_Utility.php
@@ -188,12 +188,12 @@ class Plugin_Request_Utility {
 	 * @since 1.1.0
 	 *
 	 * @param string $plugin_url The URL of the plugin to download.
-	 * @return string|bool The plugin basename if the plugin is downloaded successfully, false otherwise.
+	 * @return string|bool The full plugin path if the plugin is downloaded successfully, false otherwise.
 	 */
 	public static function download_plugin( $plugin_url ) {
 		$response = wp_remote_get( $plugin_url );
 
-		if ( is_wp_error( $response ) ) {
+		if ( is_wp_error( $response ) || 200 !== wp_remote_retrieve_response_code( $response ) ) {
 			return false;
 		}
 
--- a/includes/Utilities/Plugin_Request_Utility.php
+++ b/includes/Utilities/Plugin_Request_Utility.php
@@ -181,4 +181,63 @@ class Plugin_Request_Utility {
 
 		return $files_to_ignore;
 	}
+
+	/**
+	 * Downloads the plugin from the URL.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string|bool The plugin basename if the plugin is downloaded successfully, false otherwise.
+	 */
+	public static function download_plugin( $plugin_url ) {
+		$response = wp_remote_get( $plugin_url );
+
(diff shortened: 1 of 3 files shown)


tests/behat/features/plugin-check.feature
```diff
@@ -283,3 +283,16 @@ Feature: Test that the WP-CLI command works.
       """
       no_plugin_readme
       """
+
+  Scenario: Check a plugin from external zip
+    Given a WP install with the Plugin Check plugin
+
+    When I run the WP-CLI command `plugin check https://downloads.wordpress.org/plugin/hello-dolly.1.7.3.zip --fields=code,type --format=csv`
```
ernilambar, line 290: I have again used external zip here. In the followup PR we should create our own test plugin zip and keep in `testdata` and use from there.

swissspidy (approved):

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -181,4 +181,63 @@ public static function get_files_to_ignore() {
...
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string|bool The plugin basename if the plugin is downloaded successfully, false otherwise.
+	 */
+	public static function download_plugin( $plugin_url ) {
+		$response = wp_remote_get( $plugin_url );
+
+		if ( is_wp_error( $response ) ) {
```
mukeshpanchal27:
```suggestion
		if ( is_wp_error( $response ) || 200 !== wp_remote_retrieve_response_code( $response ) ) {
```

includes/Utilities/Plugin_Request_Utility.php
```diff
@@ -181,4 +181,63 @@ public static function get_files_to_ignore() {
...
+	/**
+	 * Downloads the plugin from the URL.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $plugin_url The URL of the plugin to download.
+	 * @return string|bool The plugin basename if the plugin is downloaded successfully, false otherwise.
+	 */
+	public static function download_plugin( $plugin_url ) {
+		$response = wp_remote_get( $plugin_url );
```
mukeshpanchal27, line 194: What happen if someone add zip that is not plugin? For example theme zip?

ernilambar: Hmm, I have not checked. This PR uses https://github.com/WordPress/plugin-check/pull/490 arbitrary path approach. I will create follow-up issue for that PR.

davidperezgar:
Why don't we make it in one PR?

includes/Utilities/Plugin_Request_Utility.php
ernilambar: Issue - https://github.com/WordPress/plugin-check/issues/547

ernilambar:
I tested this PR few more times with different use cases. It has few issues:

* If URL is invalid, `hello.php` is being checked. (May be due to the reason that is the file in the main plugins folder)
* Once I run command with valid URL like `https://downloads.wordpress.org/plugin/debug-bar.1.1.6.zip` and again run command with another URL but with same plugin slug like `https://downloads.wordpress.org/plugin/debug-bar.1.1.6111.zip`, it still checks the plugin files from earlier run.

This warrants more investigation. Closing this for now.

PR #548, opened 2024-07-25
Add arbitrary folder validation

### LINKED ISSUE #547: Validate arbitrary path before running the checks
Follow-up of [Allow static checks to be run against an arbitrary path ](https://github.com/WordPress/plugin-check/pull/490)

Proper validation should be done before start checking arbitrary path. If not a proper WordPress plugin folder, checks should not run.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/547

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -477,16 +477,50 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 *
 	 * @since 1.0.0
 	 *
+	 * @throws Exception Throws exception if input is invalid.
+	 *
 	 * @return string The plugin basename to check.
 	 */
 	final public function get_plugin_basename() {
 		if ( null === $this->plugin_basename ) {
 			$plugin = null !== $this->plugin ? $this->plugin : $this->get_plugin_param();
 
-			if ( is_dir( $plugin ) ) {
-				$this->plugin_basename = $plugin;
-			} else {
+			// Check if a slug.
+			if ( preg_match( '/^[a-z0-9]+(-?[a-z0-9]+)*(\.php)?$/i', $plugin ) ) {
 				$this->plugin_basename = Plugin_Request_Utility::get_plugin_basename_from_input( $plugin );
+			} else {
+				if ( is_dir( $plugin ) ) {
+
+					$files = glob( $plugin . '/*.php' );
+
+					$main_file_found = false;
+
+					foreach ( $files as $file ) {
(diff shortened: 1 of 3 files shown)


includes/Checker/Abstract_Check_Runner.php
```diff
@@ -477,16 +477,50 @@ private function get_check_exclude_slugs() {
...
 	 */
 	final public function get_plugin_basename() {
 		if ( null === $this->plugin_basename ) {
 			$plugin = null !== $this->plugin ? $this->plugin : $this->get_plugin_param();
 
-			if ( is_dir( $plugin ) ) {
-				$this->plugin_basename = $plugin;
-			} else {
+			// Check if a slug.
+			if ( preg_match( '/^[a-z0-9]+(-?[a-z0-9]+)*(\.php)?$/i', $plugin ) ) {
```
swissspidy: A regex check is error prone. Can you elaborate on this?

Why not also just check if it contains a plugin header?

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -477,16 +477,50 @@ private function get_check_exclude_slugs() {
...
+						if ( ! empty( $plugin_data['Name'] ) ) {
+							$main_file_found = true;
+							break;
+						}
+					}
+
+					if ( true === $main_file_found ) {
+						$this->plugin_basename = $plugin;
+					} else {
+						throw new Exception(
```
swissspidy: Is this method the right place to do this kind of validation?

Why not move this to `Plugin_Request_Utility`?

includes/Checker/Abstract_Check_Runner.php
ernilambar: We could have inputs like this:
```
sample-slug
hello.php
/var/www/path/plugin/
https://example.com/files/plugin-slug.zip (In the future PR)
```
This is for checking which routes should we use for finding basename.

ernilambar: Now wondering how WP-CLI handles this. 😊

ernilambar: CLI does not seem to accept folder as input. `wp plugin install` **plugin|zip|url** - Accepts a plugin slug, the path to a local zip file, or a URL to a remote zip file.

ernilambar: Removed the regex.

swissspidy (approved):

davidperezgar (approved):

PR #557, opened 2024-07-31
Support passing a URL for individual check results

### LINKED ISSUE #355: Allow HTML tags in error/warning messages
Followup from https://github.com/WordPress/plugin-check/issues/312

Currently we dont allow HTML tags in messages.

Opinions:

> Using late escaping principles, it would make sense to me that if HTML is being stored in messages that each output target would be responsible for its own escaping, so we could use `wp_kses` to define the subset of HTML that is supported in the admin, while CLI strips all tags.

_Originally posted by @joemcgill in https://github.com/WordPress/plugin-check/issues/312#issuecomment-1864730259_

> I like the idea of using `wp_kses()` instead of `esc_html()`:
> * For regular UI output, support a few basic inline tags (e.g. `a[href]`, `strong`, `em`, `code`, ...)
> * For CLI, support none, effectively the same as stripping tags (i.e. we probably would just use [`wp_strip_all_tags()`](https://developer.wordpress.org/reference/functions/wp_strip_all_tags/)

_Originally posted by @felixarntz in https://github.com/WordPress/plugin-check/issues/312#issuecomment-1864803007_

We need to finalize the Acceptance Criteria including list of tags those should be allowed.

swissspidy:
Follow-up to #461, see #316. Fixes #355 by not escaping the result message in the admin.

Specifically implements [this suggestion](https://github.com/WordPress/plugin-check/pull/461#issuecomment-2258578037). Here's a screenshot:

Screenshot 2024-07-31 at 11 16 32

It's up to the team to add longer error messages and add docs URLs where applicable (either in this PR or new PRs).

### DIFF
--- a/includes/Traits/Amend_Check_Result.php
+++ b/includes/Traits/Amend_Check_Result.php
@@ -23,15 +23,16 @@ trait Amend_Check_Result {
 	 *
 	 * @since 1.0.0
 	 *
-	 * @param Check_Result $result  The check result to amend, including the plugin context to check.
-	 * @param bool         $error   Whether it is an error or notice.
-	 * @param string       $message Error message.
-	 * @param string       $code    Error code.
-	 * @param string       $file    Absolute path to the file where the issue was found.
-	 * @param int          $line    The line on which the message occurred. Default is 0 (unknown line).
-	 * @param int          $column  The column on which the message occurred. Default is 0 (unknown column).
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param bool         $error     Whether it is an error or notice.
+	 * @param string       $message   Error message.
+	 * @param string       $code      Error code.
+	 * @param string       $file      Absolute path to the file where the issue was found.
+	 * @param int          $line      The line on which the message occurred. Default is 0 (unknown line).
+	 * @param int          $column    The column on which the message occurred. Default is 0 (unknown column).
+	 * @param string       $docs      URL for further information about the message.
 	 */
-	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0 ) {
+	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0, string $docs = '' ) {
 		$result->add_message(
 			(bool) $error,
 			$message,
@@ -41,6 +42,7 @@ trait Amend_Check_Result {
 				'line'   => $line,
 				'column' => $column,
 				'link'   => $this->get_file_editor_url( $result, $file, $line ),
+				'docs'   => $docs,
 			)
 		);
 	}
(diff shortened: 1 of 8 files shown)


davidperezgar:
Thanks for this PR @swissspidy I think it now looks great. I recommend a strong in the first line.

result-check-strong

What do you think @frantorres @barrykooij ?

davidperezgar (approved):

ernilambar (approved):

includes/Traits/Amend_Check_Result.php
```diff
@@ -23,15 +23,16 @@ trait Amend_Check_Result {
...
-	 * @param int          $line    The line on which the message occurred. Default is 0 (unknown line).
-	 * @param int          $column  The column on which the message occurred. Default is 0 (unknown column).
+	 * @param Check_Result $result    The check result to amend, including the plugin context to check.
+	 * @param bool         $error     Whether it is an error or notice.
+	 * @param string       $message   Error message.
+	 * @param string       $code      Error code.
+	 * @param string       $file      Absolute path to the file where the issue was found.
+	 * @param int          $line      The line on which the message occurred. Default is 0 (unknown line).
+	 * @param int          $column    The column on which the message occurred. Default is 0 (unknown column).
+	 * @param string       $docs      URL for further information about the message.
```
mukeshpanchal27, line 33:
```suggestion
	 * @param Check_Result $result  The check result to amend, including the plugin context to check.
	 * @param bool         $error   Whether it is an error or notice.
	 * @param string       $message Error message.
	 * @param string       $code    Error code.
	 * @param string       $file    Absolute path to the file where the issue was found.
	 * @param int          $line    The line on which the message occurred. Default is 0 (unknown line).
	 * @param int          $column  The column on which the message occurred. Default is 0 (unknown column).
	 * @param string       $docs    URL for further information about the message.
```

Nit.

includes/Traits/Amend_Check_Result.php
```diff
@@ -23,15 +23,16 @@ trait Amend_Check_Result {
...
 	 */
-	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0 ) {
+	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0, string $docs = '' ) {
```
mukeshpanchal27, line 35: `@since` annotation is missing for new variable.

includes/Traits/Amend_Check_Result.php
```diff
@@ -56,9 +58,10 @@ protected function add_result_message_for_file( Check_Result $result, $error, $m
 	 * @param string       $file    Absolute path to the file where the error was found.
 	 * @param int          $line    The line on which the error occurred. Default is 0 (unknown line).
 	 * @param int          $column  The column on which the error occurred. Default is 0 (unknown column).
+	 * @param string       $docs    URL for further information about the message.
 	 */
-	protected function add_result_error_for_file( Check_Result $result, $message, $code, $file, $line = 0, $column = 0 ) {
-		$this->add_result_message_for_file( $result, true, $message, $code, $file, $line, $column );
+	protected function add_result_error_for_file( Check_Result $result, $message, $code, $file, $line = 0, $column = 0, string $docs = '' ) {
```
mukeshpanchal27, line 63: Check similar comment above.

barrykooij (approved):

barrykooij:
Looks great, this will be super helpful!

mukeshpanchal27 (approved):
Thanks for the PR! left some non blocking feedbacks.

davidperezgar (approved):

includes/Traits/Amend_Check_Result.php
swissspidy: Don't think we really need that

PR #561, opened 2024-08-02
Add testdata for feature tests

ernilambar:
Follow up of https://github.com/WordPress/plugin-check/pull/542

* Add JSON file (`user-detail.json`)
* Add plugin zip file (`foo-bar-wp.zip`)

`user-detail.json` content:

```
{
  "username": "johndoe",
  "first_name": "John",
  "last_name": "Doe"
}
```

Plugin content:
`foo-bar-wp.php`:

```php
<?php
/**
 * Plugin Name: Foo Bar WP
 * Plugin URI: https://example.com
 * Description: Custom plugin.
 * Version: 0.1.0
 * Requires at least: 6.0
 * Requires PHP: 7.0
 * Author: WordPress Team
 * Author URI: https://make.wordpress.org/plugins/
 * License: GPL-2.0+
 * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
 * Text Domain: foo-bar-wp
 */

add_action(
  'init',
  function () {
    $number = mt_rand( 10, 100 );
    echo $number;

    // By default I am not that bad. Only addon tells me such.
    echo 'I am bad';

    $qargs = array(
      'post_type'      => 'post',
      'post_status'    => 'publish',
      'posts_per_page' => 1000,
      'no_found_rows'  => true,
    );
  }
);
```

`readme.txt`

```
=== Foo Bar WP ===

Contributors: wordpressdotorg
Tags: foo, bar
Tested up to: 6.5
Stable tag: 0.1.0
License: GPLv2 or later
License URI: http://www.gnu.org/licenses/gpl-2.0.html

Short description will be here. This short description has more than 150 characters to trigger the warning. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

== Description ==

Long description will be here. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

```


ernilambar:
May be we should also add here one runtime checks.

Edit: On second thought, we cannot run runtime checks from remote zip file, I guess.

swissspidy:
Where is this data used?

ernilambar:
> Where is this data used?

Here https://github.com/WordPress/plugin-check/blob/trunk/tests/behat/features/plugin-check-remote.feature#L28

ernilambar:
These testdata will replace files currently fetched from my personal repo.

swissspidy (approved):

PR #564, opened 2024-08-04
rename plugin.php to plugin-check.php

monzuralam:
# Rename plugin.php to plugin-check.php
As per the plugin development Best Practices [Folder structure](https://developer.wordpress.org/plugins/plugin-basics/best-practices/#folder-structure) recommended, the plugin's main file name should be plugin folder name.php. `Plugin Check` is our plugin name, So, `plugin.php` file should be `plugin-check.php`.

I also provide a reference from `WordPress Plugin Review Team`:
## The main file of the plugin has a name that does not follow the convention.

We expect the main plugin file (the file containing the plugin headers) to have the same name as the plugin folder, which is also the same name as the slug / permalink of the plugin.

For example, if your plugin slug is ecpt-social-manager we expect your main plugin filename to be ecpt-social-manager.php

Note that using some common names as the filename for the main plugin file can lead to issues in some configurations.

Please check out our tips on how to [structure files and folders in a plugin](https://developer.wordpress.org/plugins/plugin-basics/best-practices/#folder-structure).

The main file of this plugin is named plugin.php


swissspidy:
We know :-)

Thanks for your contribution, but we can‘t do this for backward compatibility reasons, as this would cause the plugin to deactivate automatically for all users.

PR #566, opened 2024-08-10
Add new `Offloading_Files_Check` check

### LINKED ISSUE #486: Check: Calling remote files (js, css, images, etc). Offloading / External
This objective of this check is to detect external assets being loaded in WordPress. The users have to make it local, and we consider as an `error` for this check. 

## How could develop this check?
We could use typical assets extensions. The list could start with: css, svg, jpg, jpeg, gif, png, webm, mp4, mpg, mpeg, mp3, json, fonts, etc.

Then, we have to find if they are using external source or is being loaded locally.

Our description to developers:

> ## Calling files remotely 
> 
> Offloading images, js, css, and other scripts to your servers or any remote service (like Google, MaxCDN, jQuery.com etc) is disallowed. When you call remote data you introduce an unnecessary dependency on another site. If the file you're calling isn't a part of WordPress Core, then you should include it -locally- in your plugin, not remotely. If the file IS included in WordPress core, please call that instead.
> 
> An exception to this rule is if your plugin is performing a service. We will permit this on a case by case basis. Since this can be confusing we have some examples of what are not permitted: 
>  
> Offloading jquery CSS files to Google - You should include the CSS in your plugin.
> Inserting an iframe with a help doc - A link, or including the docs in your plugin is preferred.
> Calling images from your own domain - They should be included in your plugin.
>  
> Here are some examples of what we would permit: 
>   
> Calling font families from Google or their approved CDN (if GPL compatible) 
> API calls back to your server to process possible spam comments (like Akismet) 
> Offloading comments to your own servers (like Disqus)
> oEmbed calls to a service provider (like Twitter or YouTube) 
> 
> Please remove external dependencies from your plugin and, if possible, include all files within the plugin (that is not called remotely). If instead you feel you are providing a service, please re-write your readme.txt in a manner that explains the service, the servers being called, and if any account is needed to connect.
> 

davidperezgar:
Related #486 

This adds a new PHPCS `Offloading_Files_Check` check for detecting asset offloading (styles, scripts, images, etc.)

It does so by creating 2 custom PHPCS sniffs that are inspired by ones from WPCS:

* `EnqueuedResourceOffloadingSniff` - Detect arbitrary URLs in `wp_enqueue_*` calls
* `OffloadingSniff` - Detect arbitrary URLs in HTML strings

These new sniffs are developed according to PHP_CodeSniffer requirements in a new Composer package that is symlinked from the `phpcs-sniffs` folder. We should consider moving those to a new repository or contribute them to WPCS.

The sniffs need to be available in the final Plugin Check plugin, but I don't think we need to include the `phpcs-sniffs` folder in there. It should be included as part of `vendor`. Needs testing though!

To-do:

* [x] Add PHPCS-style tests for these sniffs
* [ ] Figure out `vendor` situation for final plugin ZIP
* [x] Manual testing
* [x] Polishing
* [x] Consider proper location for these new sniffs
* [x] Run tests for these sniffs in CI


davidperezgar:
I'd need some help. I don't find functions that will help to do:
1. Detect functions and their arguments even if they are multiline. Our internal scanner has a function from the parser that detects all functions from the plugin. 
2. Detect if an URL is external (where to do it?)

swissspidy:
This is the first time we're writing custom sniffs for Plugin Check so there's lots to figure out still.

davidperezgar:
Yes, for sure we have to figure out deeply how it fits with the checks. This is a good start.

swissspidy:
>  Run tests for these sniffs in CI

FYI, this is not done yet.

What I meant with that to-do is that there ideally should be dedicated PHPCS-style unit tests under `phpcs-sniffs/PluginCheck`

We probably also need to add `phpcs-sniffs` to `.distignore` (needs testing)

swissspidy:
This is getting closer. Unit tests now run and even code coverage works.

What's left is just manual testing. Ideally by generating a production-ready ZIP file and then testing that on a new site to see whether the sniffs are really included.

ernilambar:
Amazing. I did not even know we could use separate composer package list this. 💗

davidperezgar:
I've made a zip for plugin-check and run in a clean site, and I couldn't manage to work. 

For test purposes, I use this [plugin](https://github.com/davidperezgar/break-guidelines-test), and I'm afraid it does not detect offloaded.

swissspidy:
> I've made a zip for plugin-check and run in a clean site

How did you do it? Trying to reproduce.

Which means it would be an issue with the build process.

I am trying the following steps:

1. `composer install --no-dev`
2. `wp dist-archive . /path/to/write/the/plugin-check.zip`
3. Upload `break-guidelines-test` plugin to new site
4. Upload generated `plugin-check.zip` to new site
5. Run Plugin Check against `break-guidelines-test` in the admin

Result:

Screenshot 2024-09-13 at 10 39 19

davidperezgar:
That's how I do it! Let me test it again.

davidperezgar:
Anyway, if you manage to work, it's ok.

ernilambar:
I tested it in my local setup. `PluginCheck.CodeAnalysis.EnqueuedResourceOffloading.OffloadedContent` errors are there with that test plugin.

swissspidy (approved):

ernilambar (approved):

PR #567, opened 2024-08-12
Add feature test for runtime checks

ernilambar:


ernilambar:
Output is correct but there are PHP warnings. Eg:

```
PHP Notice:  Undefined index: HTTP_HOST in /tmp/wp-cli-test-run--66b9a7bb1dfe98.30388170/wp-includes/functions.php on line 6298
```

swissspidy:
Looks like that's coming from `wp_guess_url()`

Do you get the warnings locally too?

ernilambar:
> Looks like that's coming from `wp_guess_url()`
> 
> Do you get the warnings locally too?

No, it does not appear locally. It only appears in Behat test.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -48,6 +48,10 @@ static function () use ( $permalink_structure ) {
 				}
 			);
 
+			if ( ! isset( $_SERVER['HTTP_HOST'] ) ) {
+				$_SERVER['HTTP_HOST'] = '';
+			}
```
swissspidy: This is lacking an explanatory comment.

Why are we running into this problem though in the first place? It would be great if we could address the root cause rather than papering over it.

From what I saw last time, the notice came from `wp_guess_url()`. A stack trace could tell us why it's being called.

My _guess_ from looking at the WordPress source code is as follows though:

1. `Runtime_Environment_Setup::set_up()` calls `wp_install()` to set up a new database for the runtime check
2. `wp_install()` calls `populate_options()`, which in turn calls `wp_guess_url()`. It also calls `wp_guess_url()` itself to set the `siteurl` option.

So IMHO the better way to fix this is by setting `$_SERVER['HTTP_HOST']` in `Runtime_Environment_Setup::set_up()`. Not here.

We can set it to the existing `get_option( 'siteurl' )` value from the actual WordPress site.

What do you think?

ernilambar: Yes. Also this would probably have other side effects also. I added this in a guess as I am not being able to reproduce this in my local setup or even local Behat test. I will update PR as per your feedback.

ernilambar: This code is actually in the `Runtime_Environment_Setup::set_up()`. Is it meant to be kept somewhere else?

swissspidy: Oh wow 🤦  I didn't realize that, sorry! For some reason I thought this was in `FeatureContext`.

swissspidy (approved):

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -48,6 +48,10 @@ static function () use ( $permalink_structure ) {
 				}
 			);
 
+			if ( ! isset( $_SERVER['HTTP_HOST'] ) ) {
+				$_SERVER['HTTP_HOST'] = get_option( 'siteurl' );
```
mukeshpanchal27: It seems like Behat issue, found one https://groups.google.com/g/behat/c/n4ZYbpIRNVo/m/Q34pjvbA1ikJ that we are fixing here as `wp_guess_url()` called in `populate_options()` at first place in WP and Behat not set proper server variables.

See Behat Server variable:
```php
Array(
    [SERVER_SOFTWARE] => 
    [REQUEST_URI] => 
    [HOME] => /tmp/wp-cli-home
    [BEHAT_SCENARIO_TITLE] => Perform runtime check for multi-file plugin
    [BEHAT_FEATURE_TITLE] => Test that the WP-CLI command works.
    [BEHAT_PROJECT_DIR] => /home/runner/work/plugin-check/plugin-check
    [PATH] => bin:vendor/bin:/home/runner/.composer/vendor/bin:/snap/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
    [PWD] => /tmp/wp-cli-test-run--66bdf869cd91a9.96760688
    [BEHAT_RUN] => 1
    [PHP_SELF] => bin/wp
    [SCRIPT_NAME] => bin/wp
    [SCRIPT_FILENAME] => bin/wp
    [PATH_TRANSLATED] => bin/wp
    [DOCUMENT_ROOT] => /tmp/wp-cli-test-run--66bdf869cd91a9.96760688
    [REQUEST_TIME_FLOAT] => 172[37](https://github.com/mukeshpanchal27/plugin-check/actions/runs/10403977362/job/28811496240?pr=12#step:13:38)25930.8629
    [REQUEST_TIME] => 1723725930
    [argv] => Array
        (
            [0] => bin/wp
            [1] => plugin
            [2] => check
            [3] => foo-sample
            [4] => --fields=code,type
            [5] => --format=csv
            [6] => --require=./wp-content/plugins/plugin-check/cli.php
        )

    [argc] => 7
    [SERVER_PROTOCOL] => HTTP/1.0
    [HTTP_USER_AGENT] => 
    [REQUEST_METHOD] => GET
    [REMOTE_ADDR] => 127.0.0.1
)
```

The propose changed set full URL? ex. `http://localhost:8888` that is wrong. Instead could we update.

```suggestion
				$site_url             = get_option( 'siteurl' );
				$_SERVER['HTTP_HOST'] = preg_replace( '#^https?://#', '', rtrim( $site_url, '/' ) );
```

mukeshpanchal27 (changes requested):
Thanks for the PR. Left one feedback with details.

mukeshpanchal27 (approved):

PR #575, opened 2024-08-16
Fix CLI output

### LINKED ISSUE #574: Issues in check results message CLI
Issue 1:

We have replaced `<br>` with `\n`. This seems to give problem in some terminal.

Screenshot 2024-08-16 at 12 49 37 PM

Issue 2:

`wp_strip_all_tags()` seems to remove those characters in the message which we actually want to show to the user.

Actual message:

```
Short PHP opening tag used with echo; expected "<?php echo $title ..." but found "<?= $title ..." 
```

Displayed message:

```
Short PHP opening tag used with echo; expected "
```

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/574


swissspidy (approved):
Should we perhaps replace `<code>` with backticks?

Looks good either way

PR #598, opened 2024-08-30
Prevent that runtime checks are allowed in CLI context when they cannot be used

felixarntz:
The `Abstract_Check_Runner::get_checks_to_run()` method has logic to determine whether to allow runtime checks, depending on the environment. Generally, runtime checks can only be used if the `object-cache.php` drop-in is already present (which will lead to early initialization) or if we know that the `object-cache.php` file _can_ be placed.

This works as expected for the WP Admin flow, but not for WP-CLI: In WP-CLI, everything happens in a single request, so we can't just place the `object-cache.php` drop-in later. This means PCP will currently attempt to include runtime checks in WP-CLI context even when it should not.

This PR fixes this by moving the logic into its own method, which the `CLI_Runner` class then overrides to make the logic "stricter": It can only use runtime checks if the `object-cache.php` drop-in is already present, i.e. the runner was initialized early.

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -416,14 +416,11 @@ abstract class Abstract_Check_Runner implements Check_Runner {
 	 * @throws Exception Thrown when invalid flag is passed, or Check slug does not exist.
 	 */
 	final public function get_checks_to_run() {
-		// Include file to use is_plugin_active() in CLI context.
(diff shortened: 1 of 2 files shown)


swissspidy (approved):

ernilambar (approved):
We could fix `n.e.x.t` since version in bulk in the time of new release. Lets not hold this now for that issue.

davidperezgar (approved):

mukeshpanchal27 (approved):

PR #604, opened 2024-09-02
Add admin notice for live preview blueprint

huzaifaalmesbah:
This pull request introduces an admin notice to the live preview blueprint, designed to remind users that they are viewing a preview and to highlight the Playground feature.

Huzaifa-20240902094322

### Testing Instructions:
To test this feature:
1. Navigate to the [Blueprint Builder](https://playground.wordpress.net/builder/builder.html) tool.
2. Replace the default JSON with the `blueprint.json` file from this plugin.
3. Click `Run It` to see the preview in action.

### DIFF
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -19,6 +19,11 @@
       "options": {
         "activate": true
       }
+    },
+    {
+      "step": "writeFile",
+      "path": "/wordpress/wp-content/mu-plugins/playground-notice.php",
+      "data": "<?php add_action( 'admin_notices', function() { echo '<div class=\"notice notice-info is-dismissible\"><p>This is a live preview of <strong>Plugin Check</strong>, powered by <a href=\"https://wordpress.org/playground/\" target=\"_blank\">WordPress Playground</a>.</p></div>'; } );"
     }
   ]
 }


rejaulalomkhan:
It's working and very helpful for developers. Thanks for the excellent PR @huzaifaalmesbah

mukeshpanchal27:
Why do you think it's necessary to include this additional notice? If it were necessary, the playground would have already done that. Also, I don't think PCP uses the `Live Preview` feature.

huzaifaalmesbah:
> Why do you think it's necessary to include this additional notice? If it were necessary, the playground would have already done that. Also, I don't think PCP uses the `Live Preview` feature.

Thanks for the feedback, @mukeshpanchal27. The notice is meant to improve clarity for users in preview mode, especially those new to the Playground feature. Even if PCP doesn't use Live Preview, this helps maintain consistency. If you think it's not needed, we can avoid it.

swissspidy:
I agree with Mukesh here, we don‘t even use Live Preview.

If you think people using Live Preview need a little reminder thaf they are using Playground (why would they though?), I recommend opening a feature request on Playground itself. Adding such a notice to every single plugin is not scalable.

davidperezgar:
Yes, It's much better to propose it in Playground.

PR #605, opened 2024-09-02
Add check for plugin headers

### LINKED ISSUE #591: Warning about `Requires Plugins` plugin header
According to doc of `Requires Plugins`:

```
This must contain a comma-separated list of WordPress.org-formatted slugs for its dependencies.
```

We could warn if value of `Requires Plugins` is not as per the requirement.

Reference - https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/

### LINKED ISSUE #600: Check for Plugin Header fields validation
Introduce check for plugin header fields

* **Plugin Name** 
  - Should not be `Plugin Name` or `My Basics Plugin`
* **Plugin URI**
  - Valid URL
  - WordPress.org URL not allowed
  - Should not be example domain `example.com`
* **Description**
  - Should not contain - `This is a short description of what the plugin does`
  - Should not contain - `Here is a short description of the plugin`
  - Should not contain - `Handle the basics with this plugin`
* **Author URI**
  - Valid URL
* **Requires at least**
  - Similar to validation done in readme
* **Requires PHP**
  - Similar to validation done in readme

This way we could also incorporate https://github.com/WordPress/plugin-check/issues/591 in this same check.

Ref: https://developer.wordpress.org/plugins/plugin-basics/header-requirements/

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/600
Fixes https://github.com/WordPress/plugin-check/issues/591


includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,270 @@
...
+
+		$labels = array(
+			'Name'            => 'Plugin Name',
+			'PluginURI'       => 'Plugin URI',
+			'Version'         => 'Version',
+			'Description'     => 'Description',
+			'Author'          => 'Author',
+			'AuthorURI'       => 'Author URI',
+			'TextDomain'      => 'Text Domain',
+			'DomainPath'      => 'Domain Path',
```
swissspidy, line 69: This may be out of scope, but we could add extra validation for `DomainPath`.

If it is set to a folder (e.g. to `/languages`), then that folder must exist in the plugin. Otherwise recommend removing. Plugins in the repo don't need it.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,270 @@
...
+			'Network'         => 'Network',
```
swissspidy, line 70: This may be out of scope, but we could add extra validation for `Network` as well. 
Can only be set to `true`, and should be left out when not needed.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,270 @@
...
+		$plugin_header    = get_plugin_data( $plugin_main_file );
+
+		$labels = array(
+			'Name'            => 'Plugin Name',
+			'PluginURI'       => 'Plugin URI',
+			'Version'         => 'Version',
+			'Description'     => 'Description',
+			'Author'          => 'Author',
+			'AuthorURI'       => 'Author URI',
+			'TextDomain'      => 'Text Domain',
```
swissspidy, line 68: We could move the check from `Plugin_Header_Text_Domain_Check` into this class, and remove `Plugin_Header_Text_Domain_Check`, WDYT?

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
ernilambar: I would like to see all header related checks in this class. So, it will make sense I guess.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
ernilambar: Yes, we could. We should at least use same error code here for that check to reduce possible side effect.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
ernilambar: With get_plugin_data(), this will not be possible as `Network` value comes as `true` or `false` from this function. It does not return actual header value. `$plugin_data['Network'] = ( 'true' === strtolower( $plugin_data['Network'] ) );`

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,335 @@
...
+			} elseif ( str_contains( $plugin_header['PluginURI'], '//wordpress.org/' ) || str_contains( $plugin_header['PluginURI'], '//example.com/' ) ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: %s: plugin header field */
+						__( 'The "%s" header in the plugin file uses restricted domain.', 'plugin-check' ),
+						esc_html( $labels['PluginURI'] )
+					),
+					'plugin_header_restricted_plugin_uri',
+					$plugin_main_file,
```
swissspidy: Not sure "restricted" is the right word.

We could just use the same error message as above, or say it's not "allowed".

Curious to hear what others think.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,335 @@
...
+					0,
+					'',
+					6
+				);
+			} elseif ( str_contains( $plugin_header['PluginURI'], '//wordpress.org/' ) || str_contains( $plugin_header['PluginURI'], '//example.com/' ) ) {
+				$this->add_result_warning_for_file(
+					$result,
+					sprintf(
+						/* translators: %s: plugin header field */
+						__( 'The "%s" header in the plugin file uses restricted domain.', 'plugin-check' ),
```
swissspidy:
```suggestion
						__( 'The "%s" header in the plugin file uses a restricted domain.', 'plugin-check' ),
```

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -0,0 +1,335 @@
...
+				$target_path = wp_normalize_path( $result->plugin()->path() . $domain_path );
+
+				if ( ! is_dir( $target_path ) ) {
+					$this->add_result_warning_for_file(
+						$result,
+						sprintf(
+							/* translators: %s: plugin header field */
+							__( 'The "%s" header in the plugin file must point to the existing folder.', 'plugin-check' ),
+							esc_html( $labels['DomainPath'] )
+						),
```
swissspidy:
```suggestion
						sprintf(
							/* translators: %s: plugin header field */
							__( 'The "%1$s" header in the plugin file must point to an existing folder. Found: "%2$s"', 'plugin-check' ),
							esc_html( $labels['DomainPath'] ),
							$domain_path							
						),
```

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
davidperezgar: I'm agree with @swissspidy, but I'd propose to say It's not "valid".

davidperezgar (approved):
I approve, but review the sentences that pointed @swissspidy

ernilambar:
PR has been updated.

swissspidy (approved):
Nice work, thank you!

ernilambar:
Follow up:

There is `get_wordpress_stable_version()` private method in `Plugin_Readme_Check`. If we could put it in global scope `Plugin_Header_Fields_Check` class could also use that method. Where and how could it be placed to be available from any Check class?

@swissspidy

swissspidy:
@ernilambar Maybe make it a trait?

davidperezgar:
Yes, we can use it in other checks...

Issue #597, opened 2024-08-30
Documentation of PCP check process flows in all scenarios


felixarntz:
The process flows in the different scenarios in which PCP can be used are quite complex to understand, most notably when runtime checks are involved. The complexities for runtime checks mostly stem from the fact that part of the functionality needs to be initialized very early, before plugins are loaded, which is achieved by temporarily placing an `object-cache.php` drop-in (if possible).

This issue is about documenting those flows to help contributors better understand these flows and potentially uncover current bugs.

Since I was heavily involved in the architecture of the plugin and making runtime checks possible in safe way, I went over the code in depth again (which was also very helpful to refresh my memory!) and documented the current flows individually. For now, I put it all in this public document, for easy discussion via contextual comments: https://docs.google.com/document/d/1wDGZBwWB2WAxfbHE3lygIzQFK8IssCa5apOyaBolukQ/edit
Please request comment access if you'd like to leave comments.

Note that I _did_ encounter some problems while documenting the flows, particularly about supporting checks added via other plugins ("addon checks"). I added comments on the doc for all of those problems. I would like to use this documentation issue also as an opportunity to discuss solutions to those problems.

For reference, listing the problems found here (roughly ordered by severity):
* [x] `Force_Single_Plugin_Preparation` prevents plugins other than PCP and the checked plugin from being active when any runtime checks are included, preventing addon checks to work in that context. [See doc comment.](https://docs.google.com/document/d/1wDGZBwWB2WAxfbHE3lygIzQFK8IssCa5apOyaBolukQ/edit?disco=AAABUjom0Pw) and #608 for a PR that fixes this.
* [x] When running runtime checks, the database table prefix is only modified when actually running the checks, but active plugins are still filtered. This can lead to unexpected modification of the _actual_ site database. A solution would be to modify the database table prefix more or less throughout the entire request. [See doc comment.](https://docs.google.com/document/d/1wDGZBwWB2WAxfbHE3lygIzQFK8IssCa5apOyaBolukQ/edit?disco=AAABUzHEE-g) and #768 for a PR that fixes this.
* [x] In WP-CLI context, runtime checks may be run when they shouldn't, which could lead to unexpected results with other plugins active. See #598 for a PR that fixes this.
[...]

PR #608, opened 2024-09-05
Improve `active_plugins` filter

swissspidy:
See #518 and #597.

Filters active plugins to only include required ones. This means:

* The plugin being tested
* All dependencies of the plugin being tested
* Plugin Check itself
* All active dependents of Plugin Check (they could be adding new checks)

To-do:

* [x] Add Behat tests (ideally after #518 is merged).
	* IMO such end-to-end tests are best for testing this scenario.

### DIFF
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -129,8 +129,7 @@ class Force_Single_Plugin_Preparation implements Preparation {
 				array_filter(
 					WP_Plugin_Dependencies::get_dependents( dirname( $plugin_check_basename ) ),
 					static function ( $dependent ) use ( $active_plugins, $new_active_plugins ) {
-						return in_array( $dependent, $active_plugins, true ) ||
-								in_array( $dependent, $new_active_plugins, true );
+						return in_array( $dependent, $active_plugins, true );
 					}
 				)
 			);
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -103,8 +103,7 @@ class Force_Single_Plugin_Preparation implements Preparation {
 		if ( defined( 'WP_PLUGIN_CHECK_MAIN_FILE' ) ) {
 			$plugin_check_file = WP_PLUGIN_CHECK_MAIN_FILE;
 		} else {
-			$plugins_dir       = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : WP_CONTENT_DIR . '/plugins';
-			$plugin_check_file = $plugins_dir . '/plugin-check/plugin.php';
+			$plugin_check_file = basename( dirname( __DIR__, 3 ) ) . '/plugin.php';
 		}
 
 		$plugin_check_basename = plugin_basename( $plugin_check_file );
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -127,7 +127,7 @@ class Force_Single_Plugin_Preparation implements Preparation {
 				// Include any dependents of Plugin Check, but only if they were already active.
 				array_filter(
 					WP_Plugin_Dependencies::get_dependents( dirname( $plugin_check_basename ) ),
-					static function ( $dependent ) use ( $active_plugins, $new_active_plugins ) {
+					static function ( $dependent ) use ( $active_plugins ) {
 						return in_array( $dependent, $active_plugins, true );
 					}
 				)
--- a/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
+++ b/includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
@@ -113,24 +113,27 @@ class Force_Single_Plugin_Preparation implements Preparation {
 			$plugin_check_basename, // Plugin Check itself.
 		);
 
-		WP_Plugin_Dependencies::initialize();
+		// Plugin dependencies support was added in WordPress 6.5.
+		if ( class_exists( 'WP_Plugin_Dependencies' ) ) {
+			WP_Plugin_Dependencies::initialize();
 
-		$new_active_plugins = array_merge(
-			$new_active_plugins,
-			WP_Plugin_Dependencies::get_dependencies( $this->plugin_basename )
-		);
+			$new_active_plugins = array_merge(
+				$new_active_plugins,
+				WP_Plugin_Dependencies::get_dependencies( $this->plugin_basename )
+			);
 
-		$new_active_plugins = array_merge(
-			$new_active_plugins,
-			// Include any dependents of Plugin Check, but only if they were already active.
-			array_filter(
-				WP_Plugin_Dependencies::get_dependents( dirname( $plugin_check_basename ) ),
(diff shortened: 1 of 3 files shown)


ernilambar:
Currently `Requires at least:` is set to `6.3`. We should set to `6.5` as `WP_Plugin_Dependencies` was introduced in the 6.5 release.

swissspidy:
Thanks for pointing that out. Alternatively, I could also just add some `class_exists()` checks.

Also, we should really be running PHP unit tests against our minimum required WP version...

Edit: #610

ernilambar (approved):

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -74,39 +75,67 @@ public function prepare() {
...
+		if ( ! in_array( $this->plugin_basename, $active_plugins, true ) ) {
+			return $active_plugins;
+		}
 
-			$plugin_base_file = plugin_basename( $plugin_check_file );
+		if ( defined( 'WP_PLUGIN_CHECK_MAIN_FILE' ) ) {
+			$plugin_check_file = WP_PLUGIN_CHECK_MAIN_FILE;
+		} else {
+			$plugins_dir       = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : WP_CONTENT_DIR . '/plugins';
+			$plugin_check_file = $plugins_dir . '/plugin-check/plugin.php';
```
felixarntz: Not strictly related to this PR since the code was already here before, but: Rather than doing this guesswork, wouldn't it be more reliable to look from here?
```suggestion
			$plugin_check_file = dirname( __DIR__, 3 ) . '/plugin.php';
```

Looks ugly, but should definitely be more reliable.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -74,39 +75,67 @@ public function prepare() {
...
+			);
+
+			$new_active_plugins = array_merge(
+				$new_active_plugins,
+				// Include any dependents of Plugin Check, but only if they were already active.
+				array_filter(
+					WP_Plugin_Dependencies::get_dependents( dirname( $plugin_check_basename ) ),
+					static function ( $dependent ) use ( $active_plugins, $new_active_plugins ) {
+						return in_array( $dependent, $active_plugins, true ) ||
+								in_array( $dependent, $new_active_plugins, true );
```
felixarntz: Why do we need the second condition here? That seems pointless to me, since we're checking for existence in `$new_active_plugins` which is the same array we want to potentially add these plugins to.
```suggestion
						return in_array( $dependent, $active_plugins, true );
```

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -74,39 +75,67 @@ public function prepare() {
...
+			$new_active_plugins = array_merge(
+				$new_active_plugins,
+				WP_Plugin_Dependencies::get_dependencies( $this->plugin_basename )
+			);
+
+			$new_active_plugins = array_merge(
+				$new_active_plugins,
+				// Include any dependents of Plugin Check, but only if they were already active.
+				array_filter(
+					WP_Plugin_Dependencies::get_dependents( dirname( $plugin_check_basename ) ),
```
felixarntz, line 129: What a confusing API, requiring a plugin _file_ for `get_dependencies()` but a plugin _slug_ for `get_dependents()`... 🙄

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
```diff
@@ -74,39 +75,67 @@ public function prepare() {
...
+	 * Filters active plugins to only include required ones.
+	 *
+	 * This means:
+	 *
+	 * * The plugin being tested
+	 * * All dependencies of the plugin being tested
+	 * * Plugin Check itself
+	 * * All plugins depending on Plugin Check (they could be adding new checks)
 	 *
 	 * @since 1.0.0
```
felixarntz, line 87: Can you add a `@since 1.2.0|n.e.x.t` here to cover this enhancement?

felixarntz (approved):
@swissspidy Thanks, looks great! Just a few minor comments.

felixarntz:
@swissspidy Are you planning to merge #518 first and then see which uncommented Behat tests this here will fix?

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
swissspidy: Good point. Downside is that this needs to be kept in mind in case files are moved.

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
swissspidy: The dependent could be the plugin being checked, so it would already be in `$new_active_plugins`.

But I'm also doing a `array_unique()` below, so it shouldn't matter.

swissspidy:
> @swissspidy Are you planning to merge #518 first and then see which uncommented Behat tests this here will fix?

Yep!

includes/Checker/Preparations/Force_Single_Plugin_Preparation.php
felixarntz: Yeah, that's what I was thinking. I was initially hesitant about using `array_unique()` (because before this was just one single string comparison instead), but since you use it, no need to prevent duplicates here either.

swissspidy:
Right now I can't enable any of the commented out Behat tests because they're still not passing, because we still need to address (2) and (3) from https://github.com/WordPress/plugin-check/issues/597#issuecomment-2327483452

felixarntz:
> Right now I can't enable any of the commented out Behat tests because they're still not passing, because we still need to address (2) and (3) from [#597 (comment)](https://github.com/WordPress/plugin-check/issues/597#issuecomment-2327483452)

I think it would only be point (3) that would help with the remaining Behat tests, so maybe we can work on that one first.

Could you give it a try to implement a Behat test that runs two runtime checks, one from the addon and one built-in one? For instance amend/duplicate the very last Behat test in `plugin-check.feature` that way. I'm curious what happens then, since part of the problem that (3) would solve is only relevant if the only runtime check comes from an addon. But if we also have a runtime check from the built-in ones, that's a different implication because it'll already know we need to initialize runtime check relevant logic.

felixarntz:
Following up on my previous comment, based on what I wrote in https://github.com/WordPress/plugin-check/issues/597#issuecomment-2332289958, I would actually assume that if you do what I recommended before, you'll run into the exception that the slug is invalid - which will prevent the rest from proceeding even if then the plugin checker technically would know to initialize runtime environment preparations.

swissspidy:
> > Right now I can't enable any of the commented out Behat tests because they're still not passing, because we still need to address (2) and (3) from [#597 (comment)](https://github.com/WordPress/plugin-check/issues/597#issuecomment-2327483452)
> 
> I think it would only be point (3) that would help with the remaining Behat tests, so maybe we can work on that one first.
> 
> Could you give it a try to implement a Behat test that runs two runtime checks, one from the addon and one built-in one? For instance amend/duplicate the very last Behat test in `plugin-check.feature` that way. I'm curious what happens then, since part of the problem that (3) would solve is only relevant if the only runtime check comes from an addon. But if we also have a runtime check from the built-in ones, that's a different implication because it'll already know we need to initialize runtime check relevant logic.

Argh, sorry I missed this suggestion. I think we can do this in #612 though

PR #612, opened 2024-09-05
Allow runtime initialization even when only addon checks are requested

felixarntz:
This should fix the 3rd problem outlined in https://github.com/WordPress/plugin-check/issues/597#issuecomment-2327483452. Also see [explanation of the approach](https://github.com/WordPress/plugin-check/issues/597#issuecomment-2332289958).

Note that this will only make sense to be tested in combination with #608.

### DIFF
--- a/includes/Checker/Abstract_Check_Runner.php
+++ b/includes/Checker/Abstract_Check_Runner.php
@@ -8,6 +8,7 @@
 namespace WordPress\Plugin_Check\Checker;
 
 use Exception;
+use WordPress\Plugin_Check\Checker\Exception\Invalid_Check_Slug_Exception;
 use WordPress\Plugin_Check\Checker\Preparations\Universal_Runtime_Preparation;
 use WordPress\Plugin_Check\Utilities\Plugin_Request_Utility;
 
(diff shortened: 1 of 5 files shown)


felixarntz:
@swissspidy Once #608 is merged, let's give this one a try. At the very least, it should fix part of the remaining problem with addon runtime checks.

swissspidy:
> Could you give it a try to implement a Behat test that runs two runtime checks, one from the addon and one built-in one?

Doesn't work. When I use e.g. ` --checks=non_blocking_scripts,example_runtime` I still get `Error: Check with the slug "example_runtime" does not exist.`.

swissspidy (approved):

felixarntz:
@swissspidy You approved this, but I wonder whether we should keep it open until we figured out what else is missing. Just want to make sure we're not introducing something here that's wrong. WDYT?

Maybe we should reactivate those failing Behat tests in this PR and work towards making them pass.

swissspidy:
Whether we do smaller PRs or try to fix everything in this PR, either works for me

felixarntz:
That's fair. I guess this fixes at least part of the problem, so we could already merge it.

ernilambar (approved):

PR #613, opened 2024-09-06
Warning for invalid Tested up to value

### LINKED ISSUE #609: Warn if Tested up to value in readme exceeds the released version of WP
Lets say, current stable version is `6.6`. If the value is `6.8` or above, there should be a warning.

```
ERROR: Tested up to: 6.8 - This version of WordPress does not exist (yet).
```

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/609


davidperezgar (approved):

swissspidy:
Could we add tests for this? :-)

* Latest WP version
* Latest +1
* Latest +2

ernilambar:
@swissspidy I was thinking about using similar test like `test_run_with_errors_parser_warnings_with_custom_set_transient_version()`. Should we add three separate test plugins for this? Or is there any better approach?

swissspidy:
Mocking the transient makes sense to me. We have way too many test plugins already 😄

ernilambar:
Unit test added.

swissspidy (approved):

PR #622, opened 2024-09-10
Fix apostrophe issue in error messages

ernilambar:
* Remove incorrect apostrophe from error messages
* Messages were being rendered incorrectly in HTML format. Screenshot:
Screenshot 2024-09-08 at 2 53 14 PM


swissspidy (approved):

davidperezgar (approved):
Thanks for polishing these messages!

PR #626, opened 2024-09-12
Add validation for Contributors in readme

### LINKED ISSUE #611: Custom validation of `Contributors` in readme
**The Contributors field should only contain WordPress.org usernames.**

Following are the frequent occurrence in readme.

```
Contributors: https://profiles.wordpress.org/myuserid/
Contributors: John Doe
Contributors: (this should be a list of wordpress.org userid's)
```

In the plugin, we cant validate whether user exists or not but at least we can validate if it is comma separated WP.org ID.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/611

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
@@ -682,7 +682,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		$valid = true;
 
 		foreach ( $usernames as $username ) {
-			if ( 1 !== preg_match( '/^[a-zA-Z0-9\s_.\-@]+$/', $username ) ) {
+			if ( 1 !== preg_match( '/^[a-z0-9\s_.\-@]+$/', $username ) ) {
 				$valid = false;
 				break;
 			}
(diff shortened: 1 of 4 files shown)


includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -652,6 +655,57 @@ private function check_for_warnings( Check_Result $result, string $readme_file,
...
+				break;
+			}
+		}
+
+		if ( ! $valid ) {
+			$this->add_result_warning_for_file(
+				$result,
+				sprintf(
+					/* translators: %s: plugin header field */
+					__( 'The "%s" header in the readme file must be a comma-separated list of WordPress.org-formatted userid\'s.', 'plugin-check' ),
```
swissspidy:
```suggestion
					__( 'The "%s" header in the readme file must be a comma-separated list of WordPress.org-formatted usernames.', 'plugin-check' ),
```

davidperezgar (commented):
I think that we can complent this check if they are valid usernames with the response code of website of each contributor as:
https://profiles.wordpress.com/username/

swissspidy:
Seems overkill to me.

The readme parser on dotorg already does a more accurate username validation by checking whether the user actually exists. That should suffice IMO.

davidperezgar:
Ok then.

davidperezgar (approved):

swissspidy (approved):

PR #629, opened 2024-09-12
Look for badly named files

### LINKED ISSUE #628: Check: Badly named files (spaces etc)
This check is based in this sentence that we use reviewing plugins:

> ## There are poorly named files in your plugin
> 
> Due to the different types of servers people use for WordPress, we do not permit the use of spaces or certain characters in file names, as they would cause conflicts. Related to this, we do not permit filenames with the same names, but using uppercase (or lowercase) to differentiate them, as they are not universally supported.

We have to make a check to see this problem.

davidperezgar:
Fixes #628


includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -244,6 +247,63 @@ protected function look_for_application_files( Check_Result $result, array $file
...
+			if ( preg_match('/\s/', $file ) ) {
+				$badly_name = true;
+			}
+
+			foreach ( $conflict_chars as $char ) {
+				if ( strpos( basename( $file ) , $char ) !== false ) {
+					$badly_name = true;
+					break;
+				}
+			}
```
swissspidy: Instead of a loop you could just use a regex as well.

Something like this (untested):

```suggestion
			if ( preg_match( '/' . preg_quote( implode( $conflict_chars, '|' ), '/' ) . '/' ) ) {
				$badly_name = true;
			}
```

davidperezgar: Good point! I'll change it then.

composer.lock
```diff
@@ -954,26 +954,26 @@
         },
         {
             "name": "composer/pcre",
-            "version": "2.2.0",
+            "version": "2.3.1",
```
ernilambar: These changes in lock file are unrelated to this PR.

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
```diff
@@ -98,4 +98,23 @@ public function test_run_without_any_file_type_errors() {
 		$this->assertEmpty( $errors );
 		$this->assertSame( 0, $check_result->get_error_count() );
 	}
+
+	public function test_run_with_badly_named_errors() {
```
ernilambar, line 102: May be we should add more tests for other types of errors like directory name with space, file name with restricted characters, etc.

composer.lock
davidperezgar: yes, I've missed and used composer update.

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
davidperezgar: Ok, I agree. I'll go for it.

davidperezgar:
Stills need duplicated filenames

davidperezgar:
I believe that is done!

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
```diff
@@ -98,4 +98,38 @@ public function test_run_without_any_file_type_errors() {
...
+
+		// Badly named file with special chars.
+		$this->assertArrayHasKey( 0, $errors['badly|file%name!@#$%^&*()+=[]{};:"\'<>,?|`~.php'] );
+		$this->assertArrayHasKey( 0, $errors['badly|file%name!@#$%^&*()+=[]{};:"\'<>,?|`~.php'][0] );
+		$this->assertCount( 1, wp_list_filter( $errors['badly|file%name!@#$%^&*()+=[]{};:"\'<>,?|`~.php'][0][0], array( 'code' => 'badly_named_files' ) ) );
+
+		// Duplicated filenames.
+		$this->assertArrayHasKey( 0, $errors['class-filename.php'] );
+		$this->assertArrayHasKey( 0, $errors['class-filename.php'][0] );
+		$this->assertCount( 1, wp_list_filter( $errors['class-filename.php'][0][0], array( 'code' => 'duplicated_files' ) ) );
```
ernilambar: Having same file name in separate folders should not be error I believe. From what I understand, duplicate files refers to like this:
`filename.php` and `Filename.php`. These two files in same location are treated differently based on operating system.

davidperezgar: ok!

davidperezgar: But I have to force, I cannot create in my system two files like that.

davidperezgar: How could I do it?

ernilambar: I cannot create in my MacOS also. May be we need to rewrite test from an array of files names rather than actual filename.

swissspidy: Yes that will be easiest

davidperezgar: How can achieve that? some guide to make it?

swissspidy: You should be able to achieve a positive outcome using things like mocking/overriding the `::get_files` method to return a fake list of bad files.

davidperezgar: Any ideas @felixarntz ?

davidperezgar:
This PR is stucked, so I'm going to remove the check of duplicated names to other issue/PR.

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -90,6 +90,9 @@ protected function check_files( Check_Result $result, array $files ) {
 		if ( $this->flags & self::TYPE_APPLICATION ) {
 			$this->look_for_application_files( $result, $files );
 		}
+
+		// Check for badly named files.
+		$this->look_for_badly_named_files( $result, $files );
```
felixarntz: Can you please wrap this check in a flag check, similar to all other parts of this check above? You could add a new `TYPE_BADLY_NAMED` class constant for that.

Please keep in mind to also update the `TYPE_ALL` constant accordingly.

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
```diff
@@ -98,4 +98,33 @@ public function test_run_without_any_file_type_errors() {
 		$this->assertEmpty( $errors );
 		$this->assertSame( 0, $check_result->get_error_count() );
 	}
+
+	public function test_run_with_badly_named_errors() {
+		// Test plugin without any forbidden file types.
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-file-type-badly-named-files-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check = new File_Type_Check();
```
felixarntz: See above: Here you would then pass `File_Type_Check::TYPE_BADLY_NAMED`, so that the test only tests _this_ specific part.

felixarntz (changes requested):
@davidperezgar The new logic looks good, but can you please use a conditional for this like all other parts of this check do, so that it can be controlled whether or not to include this?

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
davidperezgar: ok thanks Felix. I'll do that.

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
davidperezgar: nice!

felixarntz (approved):
Thank you for the update @davidperezgar, this looks great!

davidperezgar:
Awesome!

davidperezgar:
I'll add it for 1.2.0.

swissspidy (approved):

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -244,6 +249,42 @@ protected function look_for_application_files( Check_Result $result, array $file
 		}
 	}
 
+	/**
+	 * Looks for application files and amends the given result with an error if found.
+	 *
+	 * @since 1.0.0
```
ernilambar:
```suggestion
	 * @since 1.2.0
```

composer.lock
```diff
@@ -5848,5 +5848,5 @@
     "platform-overrides": {
         "php": "7.2.24"
     },
```
ernilambar: This lock file should not be in this PR.

PR #633, opened 2024-09-14
Add i18n check to plugin repo category

### LINKED ISSUE #632: Make available the i18 checks for Plugin Repo
As we says while reviewing a plugin and we see the issue:
 
> Please do not use variable names for the text domain portion of a gettext function.  
> 
> For example, this is a bad shortcut: __( ‘Translate me.’ , $text_domain );
> 
> The text domain also needs to be added to the plugin header. WordPress uses it to internationalize your plugin meta-data even when the plugin is disabled. The text domain should be same as the one used when loading the text domain.
> 
> If your plugin permalink is wp-penfold-macrame, then your text domain should be the same.
> 
> You can read https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#text-domains for more information.

We do not allow to use variables, constants or empty values while using translation functions.

I believe that we could use i18n Sniff from WordPress Coding Standards.

davidperezgar:
Fixes #632


davidperezgar:
I've updated this to make it available this check in Repo as well. It's necessary for us.

swissspidy:
Looks like the Behat tests need updating as they‘re failing.

ernilambar (approved):

davidperezgar:
Waiting for approval!

swissspidy (approved):

PR #641, opened 2024-09-18
Exclude test-content from the built zip.

dd32:


ernilambar:
This folder is needed to run Runtime Checks.

ernilambar:
Related https://github.com/WordPress/plugin-check/pull/427

dd32:
Huh, very weirdly named for that purpose :)

swissspidy:
Yes we should really rename it

ernilambar:
👍 on renaming. `runtime-content` or something?

swissspidy:
Yeah that works

davidperezgar:
I thought the same, that it was a weird name. +1 to runtime-content.

PR #642, opened 2024-09-18
Issue with chars in plugin name

ernilambar:
Fixes the encoding issue with `&`.

It seems get_plugin_data() gives `&amp;` for `&` character.


ernilambar:
@swissspidy I am not finding another approach here. Any better way?

swissspidy:
I would need to take a closer look, but if it works it works :)

davidperezgar (approved):

swissspidy (approved):

PR #645, opened 2024-09-18
Allow parameter force-slug

### LINKED ISSUE #644: New parameter in CLI to specify the correct slug
As @dd32 said in [#441 comment](https://github.com/WordPress/plugin-check/issues/441#issuecomment-2357911990), we need to specify the correct slug in the wp cli command, so we could compare with the ones that we have in the plugin.

It's related with PR #441

davidperezgar:
Fixes #644 and it's related with PR [#480](https://github.com/WordPress/plugin-check/pull/480)

This is a first approach to solve this issue. It works.

What I see is that phpcs allows to force the text_domain with a slug. It can be done like [explained here](https://github.com/WordPress/WordPress-Coding-Standards/blob/7f766304b0654cee7c1dfa8e548f65fce197e05c/WordPress/Sniffs/WP/I18nSniff.php#L30C46-L31C21) with the parameters:
`--runtime-set text_domain desired-slug`
So, we need to allow a new argument in PHPCS, and set the slug to check.

### DIFF
--- a/includes/Checker/Checks/General/I18n_Usage_Check.php
+++ b/includes/Checker/Checks/General/I18n_Usage_Check.php
@@ -51,7 +51,7 @@ class I18n_Usage_Check extends Abstract_PHP_CodeSniffer_Check {
 		$slug_prefix = '--force-slug=';
 		$result      = array_filter(
 			$argv,
-			function( $element ) use ( $slug_prefix ) {
+			function ( $element ) use ( $slug_prefix ) {
 				return strpos( $element, $slug_prefix ) === 0;
 			}
 		);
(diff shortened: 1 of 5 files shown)


includes/Checker/Checks/General/I18n_Usage_Check.php
```diff
@@ -41,11 +41,27 @@ public function get_categories() {
...
-			'extensions' => 'php',
+		$sniff_args = array(
 			'standard'   => 'WordPress',
+			'extensions' => 'php',
 			'sniffs'     => 'WordPress.WP.I18n',
 		);
+
+		global $argv;
+		$slug_prefix = '--force-slug=';
+		$result      = array_filter(
```
ernilambar, line 52: I not liking this approach of fetching global arguments inside the check. May be we should save this new slug globally with which we can access from any checks. We need this value in other checks also.

davidperezgar: Ok, How could I make it?

davidperezgar:
I see that @ernilambar approach is better than mine in the PR #650 . So Let's close it!

PR #649, opened 2024-09-19
WIP: Global slug value

ernilambar:


swissspidy:
@ernilambar What's the difference between this and #650?

davidperezgar:
I think this is duplicated.

PR #651, opened 2024-09-20
Detect possible updater file inside vendor folder

### LINKED ISSUE #640: Update checker inside vendor folder is not detected
For update checker, currently we look for **plugin-update-checker.php**. 

But we ignore folders "node_modules" and "vendor" folder by default when we scan files.

Those who have `plugin-update-checker` inside the vendor folder as a composer package are not being detected. Interestingly that is the case for more than 90% of the plugins who bundle update checker.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/640

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Updater_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Updater_Check.php
@@ -134,8 +134,22 @@ class Plugin_Updater_Check extends Abstract_File_Check {
 	 * @param array        $php_files List of absolute PHP file paths.
 	 */
 	protected function look_for_updater_file( Check_Result $result, array $php_files ) {
+		// Possible extra files which are not included in default files list.
+		$updater_files = array(
+			'vendor/yahnis-elsts/plugin-update-checker/plugin-update-checker.php',
+			'vendor/plugin-update-checker/plugin-update-checker.php',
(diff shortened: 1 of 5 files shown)


davidperezgar (approved):

includes/Checker/Checks/Plugin_Repo/Plugin_Updater_Check.php
```diff
@@ -134,8 +134,22 @@ protected function look_for_update_uri_header( Check_Result $result ) {
 	 * @param array        $php_files List of absolute PHP file paths.
 	 */
 	protected function look_for_updater_file( Check_Result $result, array $php_files ) {
+		// Possible extra files which are not included in default files list.
+		$updater_files = array(
+			'vendor/yahnis-elsts/plugin-update-checker/plugin-update-checker.php',
+			'vendor/plugin-update-checker/plugin-update-checker.php',
+			'vendor/kernl/kernl-update-checker/kernl-update-checker.php',
+		);
```
mukeshpanchal27, line 142: Have we received any reports of similar issues?

In my opinion, we don't need to address plugin-specific issues at this time. Instead, I suggest we wait and see if more similar bugs are reported. If we observe a pattern, we can explore possible solutions then.

mukeshpanchal27 (changes requested):

PR #658, opened 2024-09-22
Add ruleset for error_log check

### LINKED ISSUE #657: Check: Debug code has been left behind error_log
We warn the developers to not use logs functions in final plugins.

As we say while reviewing, we warn them:
> During development it is common to use functions such as error_log or var_dump to check the value or state of variables. 
> Sometimes we leave this code by mistake in our final version, please take a look at it. 

I see that is already implemented in WPCS with [DevelopmentFunctionsSniff](https://github.com/WordPress/WordPress-Coding-Standards/blob/7f766304b0654cee7c1dfa8e548f65fce197e05c/WordPress/Sniffs/PHP/DevelopmentFunctionsSniff.php#L20). So, let's use it. It should have a 6 severity as it's not a blocker for wordpress.org submission.

davidperezgar:
Fixes #657

### DIFF
--- a/phpcs-rulesets/plugin-review.xml
+++ b/phpcs-rulesets/plugin-review.xml
@@ -60,6 +60,11 @@
 		<message>The "goto" language construct should not be used.</message>
 	</rule>
 
+	<!-- Check for error logs in plugin -->
+	<rule ref="WordPress.PHP.DevelopmentFunctionsSniff">
+		<type>warning</type>
+	</rule>
+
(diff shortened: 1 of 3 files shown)


davidperezgar:
Thanks @ernilambar !

davidperezgar:
It's ready to be approved!

swissspidy (approved):

ernilambar (approved):

PR #664, opened 2024-09-25
Introduce namespace prefixing

ernilambar:


ernilambar:
We need to investigate alternate way to fix the issue.

davidperezgar:
ok

PR #670, opened 2024-09-26
Check restricted plugin header fields

### LINKED ISSUE #669: Restrict custom plugin headers
Restrict plugin headers used for updaters. 

Example:

```
GitHub Plugin URI: johndoe/package
Bitbucket Plugin URI: johndoe/package
...
etc
```

Those are not supposed to be in the plugins hosted in WordPress.org plugin directory.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/669

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
@@ -229,6 +229,41 @@ class Plugin_Header_Fields_Check implements Static_Check {
 			}
 		}
 
+		$restricted_headers = array(
+			'BitbucketPluginURI' => 'Bitbucket Plugin URI',
+			'GistPluginURI'      => 'Gist Plugin URI',
+			'GiteaPluginURI'     => 'Gitea Plugin URI',
+			'GitHubPluginURI'    => 'GitHub Plugin URI',
+			'GitLabPluginURI'    => 'GitLab Plugin URI',
+		);
+
+		$plugin_data = get_file_data( $plugin_main_file, $restricted_headers, 'plugin' );
+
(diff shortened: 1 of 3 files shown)


ernilambar:
@swissspidy Since we are now using get_file_data() for fetching headers, we could also validate `Network` field which we postponed earlier. I will do it in follow up PR.

davidperezgar (approved):

davidperezgar:
Thanks Nilambar for being fast to solve this meta's request.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -74,6 +69,16 @@ public function run( Check_Result $result ) {
 			'RequiresPlugins' => 'Requires Plugins',
 		);
 
+		$restricted_labels = array(
+			'BitbucketPluginURI' => 'Bitbucket Plugin URI',
+			'GistPluginURI'      => 'Gist Plugin URI',
+			'GiteaPluginURI'     => 'Gitea Plugin URI',
+			'GitHubPluginURI'    => 'GitHub Plugin URI',
+			'GitLabPluginURI'    => 'GitLab Plugin URI',
+		);
```
swissspidy, line 78: Do we want to block just these fields, or should we just block anything that is not supported by `get_plugin_data`?

ernilambar: We should not block like that. Currently this list contains fields that are used for updater which we don't allow. Other non-core fields are frequently seen like `Elementor tested up to`, `WC tested up to`, etc which should be allowed.

swissspidy (approved):

PR #681, opened 2024-09-29
Allow `runtime-set` in Abstract_PHP_CodeSniffer_Check class and implement it in I18n_Usage_Check

### LINKED ISSUE #679: Allow `Abstract_PHP_CodeSniffer_Check` based check to add `runtime-set` arguments
Currently only following args are allowed:

* standard
* extensions
* sniffs
* exclude

**Use case:**

What I am trying to do is running PHPCS with `--runtime-set text_domain plugin-slug`. As of now, this does not seem to be possible.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/679

* Accepts **runtime-set** arguments in an array form( key value pair)

Example implementation:

```php

return array(
  'extensions'  => 'php',
  'standard'    => 'WordPress',
  'sniffs'      => 'WordPress.WP.I18n',
  'runtime-set' => array(
    'text_domain' => 'sample-domain',
  ),
);

```

**Edit:**
* Implement `runtime-set text_domain` in `I18n_Usage_Check` class
* Update i18n related tests as now errors will be increased
* Update other test files where `i18n_usage` check was used for the errors

### DIFF
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -43,7 +43,6 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 	 * @since 1.0.0
 	 *
 	 * @param Check_Result $result The check result to amend, including the plugin context to check.
-	 *
 	 * @return array {
 	 *    An associative array of PHPCS CLI arguments. Can include one or more of the following options.
 	 *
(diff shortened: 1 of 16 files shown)


swissspidy:
> Example implementation:

Are you gonna do that in a follow-up PR?

swissspidy (approved):

ernilambar:
> > Example implementation:
> 
> Are you gonna do that in a follow-up PR?

I could not find idea to get slug value in that `protected function get_args()`. We don't have context or result object passed to there.

swissspidy:
We can make `get_args()` accept a `Check_Result` as a param and then pass `$result` to `get_args()` here:

https://github.com/WordPress/plugin-check/blob/6418d8d1ac11ba5bc29659667c698ad934dbb15a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php#L82-L86

ernilambar:
@swissspidy ok. That will afect lots of files. Separate pr makes sense?

swissspidy:
Up to you :-)

swissspidy (approved):
Niiice!

davidperezgar (approved):
Wow! Awesome!

PR #684, opened 2024-10-01
Make `Contributors` regex check less strict

ernilambar:
After feedback from Meta team, we are going to make regex check for `Contributors` less strict.

* Make case insensitive
* Replace `\s` with actual space


dd32 (approved):
This matches the WordPress.org user format a bit more clearly.

WordPress.org usernames today at signup can't contain things like `@` or uppercase characters, but historically you could even use your email as your username or `Joe Bloggs`.

For 99% of users, having uppercase or @ in the field is incorrect, but for some it's expected.

On WordPress.org we will probably ignore this check and/or extend it with a `Check the user table`.

swissspidy (approved):

PR #686, opened 2024-10-01
Description in admin page

### LINKED ISSUE #140: Add copy to the page to make it more user friendly
> I would like to see some copy on the page, providing context on what the plugin does. That said, if its mainly plugin developers and engineers at Automattic using it, it's probably not needed. Worth adding this as a discussion point though. 

I think adding a couple of sentences on the Admin Page to guide the user on what to do and what to expect would be enough? We don't want to provide a detailed guide and we have documentation alongside this in the future? Worth discussing.

davidperezgar:
Fixes #140

### DIFF
--- a/templates/admin-page.php
+++ b/templates/admin-page.php
@@ -23,7 +23,7 @@
 				</h2>
 
 				<p>
-					<?php esc_html_e( 'Select a plugin to check best practices in several categories and security issues.', 'plugin-check' ); ?>
+					<?php esc_html_e( 'Select a plugin to check it for best practices in several categories.', 'plugin-check' ); ?>
 				</p>
 
 				<select id="plugin-check__plugins-dropdown" name="plugin_check_plugins">
--- a/templates/admin-page.php
+++ b/templates/admin-page.php
@@ -22,6 +22,10 @@
 					</label>
 				</h2>
 
+				<p>
+					<?php esc_html_e( 'Select a plugin to check best practices in several categories and security issues.', 'plugin-check' ); ?>
+				</p>
+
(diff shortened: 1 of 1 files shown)


templates/admin-page.php
```diff
@@ -22,6 +22,10 @@
 					</label>
 				</h2>
 
+				<p>
```
felixarntz: Maybe we should give this an `id` attribute and then in the `select` provide `aria-describedby` that points to this description? That would be better for accessibility.

templates/admin-page.php
```diff
@@ -22,6 +22,10 @@
...
+					<?php esc_html_e( 'Select a plugin to check best practices in several categories and security issues.', 'plugin-check' ); ?>
```
felixarntz: This reads a bit confusing IMO. I wouldn't call out "security issues" explicitly since it's one of the "several categories" anyway. Maybe better something like this?
```suggestion
					<?php esc_html_e( 'Select a plugin to check it for best practices in several categories.', 'plugin-check' ); ?>
```

davidperezgar: Yeah, why not.

felixarntz: One follow up:
```suggestion
					<?php esc_html_e( 'Select a plugin to check it for best practices in several categories and security issues.', 'plugin-check' ); ?>
```

swissspidy:
Worth mentioning the screen help tab for additional information? That's where we added the check/categories descriptions.

davidperezgar:
Now I think is ready with all suggestions.

ernilambar (approved):

felixarntz (approved):
Thanks @davidperezgar!

PR #689, opened 2024-10-02
Output a success message if no errors or warnings are found.

### LINKED ISSUE #685: CLI: if "wp plugin check <my-plugin>" finds no errors/warnings, then output a message saying such
thanx for this useful plugin.  I've used it several times from the WP Dashboard (`Tools > Plugin Check`).  Today I tried the WP_CLI `wp plugin check` command for the first time.

It would be helpful if the `check` command output a `WP_CLI::success()` message if no errors/warnings are found.  The way things are it is not possible to know whether there were no errors/warnings or whether there was a run-time fatal while the `check` command was running (without consulting the PHP error log).

Not sure how such a message might affect the GitHub Action, as I don't use GH actions ;-)

If there is consensus that this would be a good idea, I'm happy to do a PR.

pbiron:
fixes #685

When no errors or warnings are found, output a success message with that same text that is used in `Tools > Plugin Check`.

Before merging this, please verify that producing output on success doesn't cause problems with [plugin-check-action](https://github.com/WordPress/plugin-check-action).

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -244,7 +244,7 @@ final class Plugin_Check_Command {
 		}
 
 		if ( empty( $errors ) && empty( $warnings ) ) {
-			WP_CLI::success( 'Checks complete. No errors found.' );
+			WP_CLI::success( __( 'Checks complete. No errors found.', 'plugin-check' ) );
 
 			return;
 		}
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -243,6 +243,12 @@ final class Plugin_Check_Command {
 			$warnings = $result->get_warnings();
 		}
 
+		if ( empty( $errors ) && empty( $warnings ) ) {
+			WP_CLI::success( 'Checks complete. No errors found.' );
+
+			return;
+		}
+
 		// Default fields.
 		$default_fields = $this->get_check_default_fields( $assoc_args );
 
--- a/tests/behat/features/plugin-check.feature
+++ b/tests/behat/features/plugin-check.feature
@@ -601,3 +601,33 @@ Feature: Test that the WP-CLI command works.
 #      """
 #      ExampleRuntimeCheck.ForbiddenScript,WARNING
 #      """
+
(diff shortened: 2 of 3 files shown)


swissspidy:
Would you mind updating the Behat tests to cover this new scenario? :-)

pbiron:
@swissspidy I'm just looking at that now.  I'm not great with Behat, but I'll try.

pbiron:
@swissspidy OK, the test with `--ignore-errors` is now succeeding and the code linting is now passing as well.

There is no existing test for a plugin with no (non-ignored) errors or warnings (i.e., that expects STDOUT to be empty).  Would it be helpful to add one?

Not sure why some of the tests were "canceled" after the last commit.  Can you please restart the tests?

pbiron:
All tests are passing now ;-)

swissspidy (approved):
Looks great to me, thanks!

pbiron:
will the [plugin-check-action](https://github.com/WordPress/plugin-check-action) need to be updated?  Since I don't use GHA, I don't know if it will fail when the success message is output.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -243,6 +243,12 @@ static function ( $dirs ) use ( $excluded_files ) {
 			$warnings = $result->get_warnings();
 		}
 
+		if ( empty( $errors ) && empty( $warnings ) ) {
+			WP_CLI::success( 'Checks complete. No errors found.' );
```
davidperezgar: I don't see the translation function...
__( '', 'plugin-check' )

davidperezgar (changes requested):

davidperezgar:
Thanks Paul for the PR! I've requested one change.

includes/CLI/Plugin_Check_Command.php
pbiron: sorry, I hadn't noticed that other strings passed to `WP_CLI::line()`, etc were translated...I'm so used to strings in standard WP_CLI commands not being translated.

Another commit coming shortly.

davidperezgar (approved):

davidperezgar:
Nice!

PR #690, opened 2024-10-02
Added support to multisite

### LINKED ISSUE #64: Add multisite support for admin page

The admin screen from #32 and the plugin action link from #36 are not fully working in multisite yet, or rather, they are limited to the individual site's WP Admin UI. We should expand that so that the same admin screen works in the Network Admin UI as well.

This is particularly critical, since on some multisite configurations regular administrators are not allowed to control plugins.

## Acceptance Criteria

* The admin page class from #32 should be expanded to also add the same admin screen on the `network_admin_menu` action hook.
    * The admin page added can behave the same way for now, but since the "Tools" menu does not exist in the Network Admin, we should add the screen in the "Settings" menu there (i.e. use `add_submenu_page()` with `settings.php`).
* The plugin action links from #36 should be expanded to also show up in the Network Admin via the `network_admin_plugin_action_links` filter.
    * Particular attention has to be paid to the URL, since the page lives under `settings.php` here (instead of `options-general.php` in the regular WP Admin).

## Test Coverage

* Ensure the admin page is added propertly when using the Network Admin specific action.
* Ensure the plugin link is added properly when using the Network Admin specific filter.

davidperezgar:
Fixes #64

### DIFF
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -54,7 +54,7 @@ final class Admin_Page {
 	 * @since 1.0.0
 	 */
 	public function add_hooks() {
-		$admin_menu_hook = is_multisite() ? 'network_admin_menu' : 'admin_menu';
+		$admin_menu_hook         = is_multisite() ? 'network_admin_menu' : 'admin_menu';
 		$plugin_action_link_hook = is_multisite() ? 'network_admin_plugin_action_links' : 'plugin_action_links';
 
 		add_action( $admin_menu_hook, array( $this, 'add_and_initialize_page' ) );
@@ -322,7 +322,13 @@ final class Admin_Page {
 			return $actions;
 		}
 
-		if ( ( is_multisite() && current_user_can( 'manage_network_plugins' ) ) || current_user_can( 'activate_plugins' ) ) {
+		if ( is_multisite() && current_user_can( 'manage_network_plugins' ) ) {
+			$actions[] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( network_admin_url( "settings.php?page=plugin-check&plugin={$plugin_file}" ) ),
+				esc_html__( 'Check this plugin', 'plugin-check' )
+			);
+		} elseif ( current_user_can( 'activate_plugins' ) ) {
 			$actions[] = sprintf(
 				'<a href="%1$s">%2$s</a>',
 				esc_url( admin_url( "tools.php?page=plugin-check&plugin={$plugin_file}" ) ),
--- a/includes/Admin/Admin_Page.php
+++ b/includes/Admin/Admin_Page.php
@@ -54,8 +54,11 @@ final class Admin_Page {
 	 * @since 1.0.0
 	 */
 	public function add_hooks() {
-		add_action( 'admin_menu', array( $this, 'add_and_initialize_page' ) );
-		add_filter( 'plugin_action_links', array( $this, 'filter_plugin_action_links' ), 10, 4 );
+		$admin_menu_hook = is_multisite() ? 'network_admin_menu' : 'admin_menu';
+		$plugin_action_link_hook = is_multisite() ? 'network_admin_plugin_action_links' : 'plugin_action_links';
+
+		add_action( $admin_menu_hook, array( $this, 'add_and_initialize_page' ) );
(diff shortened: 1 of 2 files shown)


davidperezgar:
The tests are not working in GitHub actions, but when I run on local, works smoothly.

includes/Admin/Admin_Page.php
```diff
@@ -54,8 +54,11 @@ public function __construct( Admin_AJAX $admin_ajax ) {
 	 * @since 1.0.0
 	 */
 	public function add_hooks() {
-		add_action( 'admin_menu', array( $this, 'add_and_initialize_page' ) );
-		add_filter( 'plugin_action_links', array( $this, 'filter_plugin_action_links' ), 10, 4 );
+		$admin_menu_hook         = is_multisite() ? 'network_admin_menu' : 'admin_menu';
+		$plugin_action_link_hook = is_multisite() ? 'network_admin_plugin_action_links' : 'plugin_action_links';
+
+		add_action( $admin_menu_hook, array( $this, 'add_and_initialize_page' ) );
+		add_filter( $plugin_action_link_hook, array( $this, 'filter_plugin_action_links' ), 10, 4 );
```
felixarntz, line 61: I'm not sure we would want to only show the Plugin Check UI in the Network Admin just because it's a Multisite network. Maybe it should be _in addition_ instead? I think that's worth discussing further before we make a decision and implement it.

includes/Admin/Admin_Page.php
```diff
@@ -64,13 +67,18 @@ public function add_hooks() {
 	/**
 	 * Adds the admin page under the tools menu.
 	 *
+	 * @since n.e.x.t Added multisite support.
 	 * @since 1.0.0
 	 */
 	public function add_page() {
-		$this->hook_suffix = add_management_page(
+		$admin_parent_page = is_multisite() ? 'settings.php' : 'tools.php';
+		$capabilities      = is_multisite() ? 'manage_network_plugins' : 'activate_plugins';
```
felixarntz, line 75: See above, maybe this should support both regular WP Admin _and_ Network Admin?

Instead of checking `is_multisite()` (which forces it to only work for one of the two), you could check for `is_network_admin()` to make the decision.

felixarntz, line 75: This variable should probably be called `$capability` since it only stores _one_ capability.

includes/Admin/Admin_Page.php
```diff
@@ -307,7 +315,13 @@ public function filter_plugin_action_links( $actions, $plugin_file, $plugin_data
...
 		}
 
-		if ( current_user_can( 'activate_plugins' ) ) {
+		if ( is_multisite() && current_user_can( 'manage_network_plugins' ) ) {
+			$actions[] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( network_admin_url( "settings.php?page=plugin-check&plugin={$plugin_file}" ) ),
+				esc_html__( 'Check this plugin', 'plugin-check' )
+			);
+		} elseif ( current_user_can( 'activate_plugins' ) ) {
```
felixarntz, line 324: See above, this could also support both WP Admin and Network Admin UIs.
```suggestion
		if ( is_network_admin() && current_user_can( 'manage_network_plugins' ) ) {
			$actions[] = sprintf(
				'<a href="%1$s">%2$s</a>',
				esc_url( network_admin_url( "settings.php?page=plugin-check&plugin={$plugin_file}" ) ),
				esc_html__( 'Check this plugin', 'plugin-check' )
			);
		} elseif ( ! is_network_admin() && current_user_can( 'activate_plugins' ) ) {
```

felixarntz (commented):
@davidperezgar I think we need to discuss a bit more what we want Multisite support of Plugin Check to look like. It's probably a bit too early to implement this PR without thinking about this more.

For example, what would be the benefit of having the admin page in the Network Admin instead of regular WP Admin if the functionality is exactly the same? Should we maybe show it under both? Also, what about the behavior in terms of runtime checks? Should the Network Admin UI only allow covering network-active plugins?

There are several questions to discuss here, so I suggest we move this conversation to the issue before we work on the code, to make decisions on these blocking questions.

davidperezgar:
Thanks for your feedback, yes, maybe we could discuss in the issue, what would expect to have in MU, and then this PR. I agree what you mentioned.

davidperezgar:
Closed until we have the discussion about Multisite.

PR #696, opened 2024-10-03
CLI: Treat errors below severity threshold as warnings

### LINKED ISSUE #691: Show errors with severity less than a number as a warning
In WP-CLI, when we use severity, such as 7 for errors, we lose the results of those errors and they are not displayed. My proposal would be to show them as warnings, so the user can still address them since they are important, but they do not block the process.

The severity level 7 is used in the [WordPress Plugin Submission](https://meta.trac.wordpress.org/changeset/14074), and it filters errors with a severity level lower than 7. That's correct, but we are losing other issues that, if the user received those results, they could work on them. However, it wouldn't block the submission. Therefore, we could display them as warnings.

davidperezgar:
Fixes #691


ernilambar:
Current approach of severity is aligned with the PHPCS which could be considered as an industry standard. I would not want to break that approach and expectation. 
IMO, we could add some flags to get the required results with errors converted to warnings. 
Anyway lets wait for other opinions also.

davidperezgar:
Ok @ernilambar I agree with you as you tell that is a industry standard. We all want the same results. Which name of the flag do you recommend? something like `--show-error-warning-low-severity` ?

felixarntz (commented):
@davidperezgar I'm not sure I fully understand the problem that this is trying to solve. Is the problem that some `$results` are removed because of the filtering? That seems reasonable to fix.

However, even with this PR implementation, some results would still be missing, which would be warnings with a lower severity.

I'm not sure that we should "migrate" low-severity errors to warnings. That seems like an odd choice. Would be great to get other people's feedback on this.

davidperezgar:
Thanks @felixarntz . The thing is that we have errors that could have false positives. That's why we use the severity 7 in plugin submission, so the errors with severity less than 7, won't be a blocker. In our team we try to developers to learn about the errors, and it would be very helpful for the team, show that errors, and they could be a warning. I think is the best scenario.  

Otherwise, they would find that issues does not a appear and they could not manage to solve them.

I agree with @ernilambar to use a new flag to do that.

davidperezgar:
I've commited my proposal. To have less severity errors as warnings with the flag: `--include-error-severity`. That will include less severity errors as warnings.

swissspidy:
A Behat test covering this change would be beneficial.

davidperezgar:
Ok, I'm going to make it then.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -687,4 +696,32 @@ function ( $item ) use ( $warning_severity ) {
...
+	/**
+	 * Returns check results filtered by severity level.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param array $results          Check results.
+	 * @param int   $error_severity   Error severity level.
+	 * @return array Filtered results.
+	 */
+	private function convert_filtered_errors_by_severity( $results, $error_severity ) {
```
swissspidy: Implementation-wise, this could be much simplified and integrated into the `get_filtered_results_by_severity()` method. There we can do a simple `foreach` loop so we only loop through the messages once and avoid unnecessary `array_filter` and `array_merge` callse.

tests/behat/features/plugin-check-severity.feature
```diff
@@ -192,5 +192,19 @@ Feature: Test that the severity level in plugin check works.
...
+      default_readme_text,WARNING,7
+      """
+    And STDOUT should contain:
+      """
+      upgrade_notice_limit,WARNING,5
+      """
+
     When I run the WP-CLI command `plugin check foo-bar-wp --format=csv --fields=code,type,severity --severity=10`
-    Then STDOUT should be empty
+    Then STDOUT should be empty
```
swissspidy: Nit: Missing newline at end of file

swissspidy:
I agree with @felixarntz that it is odd to downgrade errors below the severity threshold to warnings. Plus, the same isn't implemented for warnings, so it's still an open question whether they should have the same treatment or not.

Maybe we could find another way of including/naming the messages below the threshold? Maybe instead of "WARNING" use "IGNORED" or so.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -140,15 +143,16 @@ public function check( $args, $assoc_args ) {
...
-				'slug'                 => '',
+				'checks'                 => '',
+				'format'                 => 'table',
+				'ignore-warnings'        => false,
+				'ignore-errors'          => false,
+				'include-experimental'   => false,
+				'severity'               => '',
+				'error-severity'         => '',
+				'warning-severity'       => '',
+				'include-error-severity' => '',
```
felixarntz: I'm not sure this is a suitable name for what the argument controls. Isn't this responsible for "downgrading" certain errors to warnings if they are low severity? The current name does not clarify that, it sounds like it would lead to showing the severity in the results table, or include additional elements based on certain criteria.

Maybe something like `include-low-severity-errors`?

includes/CLI/Plugin_Check_Command.php
felixarntz: +1 to combining this, it's overly complicated to have a separate method for it, both in terms of implementation and understanding the code. You could make the new boolean argument a parameter on `get_filtered_results_by_severity()`.

I'm not sure the current implementation works actually: As far as I can tell, the `ERROR` condition below is mutually exclusive with the `ERROR` condition in `get_filtered_results_by_severity()`, which runs on `$file_results` before this method. So the array would always be empty here? 🤔 

Either way, I think it makes more sense to combine this in one place with shared logic.

felixarntz (changes requested):
Thanks for the PR @davidperezgar, a few points of feedback to improve the way this currently works.

tests/behat/features/plugin-check-severity.feature
```diff
@@ -192,5 +192,19 @@ Feature: Test that the severity level in plugin check works.
       upgrade_notice_limit,WARNING,5
       """
 
+    When I run the WP-CLI command `plugin check foo-bar-wp --format=csv --fields=code,type,severity --error-severity=7 --include-error-severity`
```
ernilambar: May be I am missing something but I don't see this scenario testing the converted ERROR to WARNING. Which ERROR with less than 7 severity is converted to WARNING here?

ernilambar:
In WP.org we run the plugin with arguments `--error-severity=7 --warning-severity=6`
As I understand from the issue details, we want ERRORs less that 7 severity converted to WARNINGs so that messages are displayed for the plugin author. If we just convert ERROR to WARNING only (without converting severity) some errors will be still hidden. ERROR with 5 severity is converted to WARNING with 5 severity and it will still be hidden from the output. Please correct me if my understanding is wrong.

includes/CLI/Plugin_Check_Command.php
ernilambar: Agree. Argument name is pretty confusing.

ernilambar:
For me I would not touch `get_filtered_results_by_severity()` method in this case. Its responsibility is clear from its name and defined perfectly. It just filters the result based on given error and warning severity. If we need any manipulation like converting ERROR to WARNING or converting WARNING to ERROR, it should be done beforehand and passed modified results to this function.

davidperezgar:
Ok, thanks Nilambar for your comments. I'll check it out.

davidperezgar:
I tried to combine the function but PHPMD give me errors about the responsibility of the function, using as an argument a boolean.

includes/CLI/Plugin_Check_Command.php
davidperezgar: Ok, I wasn't sure for the name, we all clarified that. Changed.

tests/behat/features/plugin-check-severity.feature
davidperezgar: yes, it's on the results.

swissspidy:
> I tried to combine the function but PHPMD give me errors about the responsibility of the function, using as an argument a boolean.

they can always be suppressed :-)

davidperezgar:
Ok, if we remove that check, it would be ok to review.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -668,15 +673,22 @@ private function has_runtime_check( array $checks ) {
 	 * @param array $results          Check results.
 	 * @param int   $error_severity   Error severity level.
 	 * @param int   $warning_severity Warning severity level.
+	 * @param bool  $include_low_severity_errors Include less level of severity issues as warning.
```
felixarntz: Can you fix the indentation so that these docs are aligned?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -113,6 +113,9 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * [--warning-severity=<warning-severity>]
 	 * : Warning severity level.
 	 *
+	 * [--include-low-severity-errors]
+	 * : Include less level of severity issues as warning.
```
```suggestion
	 * : Include errors with lower severity than the threshold as warnings.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -140,15 +143,16 @@ public function check( $args, $assoc_args ) {
...
-				'slug'                 => '',
+				'checks'                      => '',
+				'format'                      => 'table',
+				'ignore-warnings'             => false,
+				'ignore-errors'               => false,
+				'include-experimental'        => false,
+				'severity'                    => '',
+				'error-severity'              => '',
+				'warning-severity'            => '',
+				'include-low-severity-errors' => '',
```
felixarntz: Shouldn't this use `false` as default instead of `''`?

includes/CLI/Plugin_Check_Command.php
```diff
@@ -668,15 +673,22 @@ private function has_runtime_check( array $checks ) {
...
-				return ( 'ERROR' === $item['type'] && $item['severity'] >= $error_severity );
+	private function get_filtered_results_by_severity( $results, $error_severity, $warning_severity, $include_low_severity_errors = false ) {
+		$errors = array();
+		foreach ( $results as $item ) {
+			if ( 'ERROR' === $item['type'] && $item['severity'] >= $error_severity ) {
+				$errors[] = $item;
+			} elseif ( $include_low_severity_errors && $item['severity'] < $error_severity ) {
+				$item['type']     = 'WARNING';
+				$item['severity'] = 10;
+				$errors[]         = $item;
```
felixarntz: This doesn't look right, warnings shouldn't be added to the `$errors` list, but the `$warnings` list.

felixarntz (commented):
@davidperezgar This looks better, a few points of follow up feedback.

felixarntz:
@ernilambar 
> For me I would not touch `get_filtered_results_by_severity()` method in this case. Its responsibility is clear from its name and defined perfectly. It just filters the result based on given error and warning severity. If we need any manipulation like converting ERROR to WARNING or converting WARNING to ERROR, it should be done beforehand and passed modified results to this function.

This is a fair point, I would agree with it. This is also confirmed by PHPMD, because boolean parameters to change what a method does are code smell indeed. Still, implementing this as a single method is more performant, so there's arguments for either side. Moving this logic to a separate `convert_low_severity_errors_to_warnings()` method would make a lot of sense to me - but it would need to be called _before_ `get_filtered_results_by_severity()` like you're saying, not how this was initially implemented. But either way works for me.

> As I understand from the issue details, we want ERRORs less that 7 severity converted to WARNINGs so that messages are displayed for the plugin author. If we just convert ERROR to WARNING only (without converting severity) some errors will be still hidden. ERROR with 5 severity is converted to WARNING with 5 severity and it will still be hidden from the output. Please correct me if my understanding is wrong.

Thanks for clarifying, that kind of makes sense to me. Still, related to what @swissspidy pointed out earlier, I'm curious what's the purpose of that. If we want to see the low severity errors, why do we even limit the error severity? We could just omit that argument and only pass warning severity, and that way we would get all errors, plus high-severity warnings. The idea of migrating low-severity errors to warnings is still questionable as a benefit to me: For example, how can we reliably assume that a low-severity error is a high-severity warning? We don't know that, so it feels wrong to me to just assume warning severity `10`, like this PR currently does. So I would like to better understand why this is needed.

davidperezgar:
I give you the main explanation for this PR. We want to avoid false positives in Plugin submission. This means that if you have errors, you won't be able to upload your plugin to submission, as it will be blocked. For errors with less severity, I aim to show the user how to solve them while the Team is reviewing but not blocking the update.

Is that make sense?

includes/CLI/Plugin_Check_Command.php
davidperezgar: ok

includes/CLI/Plugin_Check_Command.php
davidperezgar: yes!

swissspidy:
> For example, how can we reliably assume that a low-severity error is a high-severity warning? We don't know that, so it feels wrong to me to just assume warning severity 10, like this PR currently does. So I would like to better understand why this is needed.

I think it would help if we didn't label them was warnings, but as "ignored" or "info" or "ignored error" or so.

An error might be below the threshold, but it's still an error, not a warning.

davidperezgar:
But all the system is based in errors and warnings. I don't see make another type of error.

swissspidy:
It's not really a new type, it's just a different label that is displayed to the user. Converting an error to a warning is more confusing than just adding an "(ignored)" to it IMO

ernilambar:
> It's not really a new type, it's just a different label that is displayed to the user. Converting an error to a warning is more confusing than just adding an "(ignored)" to it IMO

Yah, this approach would be easier I think. Rather than conversion, we could just append the ignored items with ERROR replaced with ERROR_EXTRA (or something better name).

ernilambar:
> If we want to see the low severity errors, why do we even limit the error severity? We could just omit that argument and only pass warning severity, and that way we would get all errors, plus high-severity warnings. 

Because not all ERRORs are 100% full proof. Some errors could potentially have false positives and those should be reviewed manually. Eg, `WordPress.Security.EscapeOutput` could show several false positives.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -665,25 +670,29 @@ private function has_runtime_check( array $checks ) {
...
-	 * @param array $results          Check results.
-	 * @param int   $error_severity   Error severity level.
-	 * @param int   $warning_severity Warning severity level.
+	 * @param array $results                     Check results.
+	 * @param int   $error_severity              Error severity level.
+	 * @param int   $warning_severity            Warning severity level.
+	 * @param bool  $include_low_severity_errors Include less level of severity issues as warning.
+	 *
+	 * @SuppressWarnings(PHPMD.BooleanArgumentFlag)
 	 * @return array Filtered results.
```
```suggestion
	 * @param array $results                     Check results.
	 * @param int   $error_severity              Error severity level.
	 * @param int   $warning_severity            Warning severity level.
	 * @param bool  $include_low_severity_errors Include less level of severity issues as warning. Default false.
	 * @return array Filtered results.
	 *
	 * @SuppressWarnings(PHPMD.BooleanArgumentFlag)
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -665,25 +670,29 @@ private function has_runtime_check( array $checks ) {
...
+		$warnings = array();
 
-		$warnings = array_filter(
-			$results,
-			function ( $item ) use ( $warning_severity ) {
-				return ( 'WARNING' === $item['type'] && $item['severity'] >= $warning_severity );
+		foreach ( $results as $item ) {
+			if ( 'ERROR' === $item['type'] && $item['severity'] >= $error_severity ) {
+				$errors[] = $item;
+			} elseif ( $include_low_severity_errors && $item['severity'] < $error_severity ) {
```
mukeshpanchal27: Do we needs to add `'ERROR' === $item['type']` check here?

davidperezgar: yes.

davidperezgar:
Last thing that we need to discuss how to name the low severity errors.  I believe they are Warnings.

felixarntz:
I like the approach using a different type.

How about this?
* Use custom type `ERRORS_LOW_SEVERITY` for low severity errors.
* Do the same for warnings, i.e. add a flag `--include-low-severity-warnings`, and if that is set, include low severity warnings under `WARNINGS_LOW_SEVERITY`.

WDYT?

davidperezgar:
Hello, 
I see that is a good approach and all will be happy with that solution.

davidperezgar:
Done! and ready to review!

davidperezgar:
It's strange as it works BEHAT tests in my local installation, but is failing.

includes/CLI/Plugin_Check_Command.php
```diff
@@ -111,6 +111,12 @@ public function __construct( Plugin_Context $plugin_context ) {
 	 * [--warning-severity=<warning-severity>]
 	 * : Warning severity level.
 	 *
+	 * [--include-low-severity-errors]
+	 * : Include errors with lower severity than the threshold as other type.
+	 *
+	 * [--include-low-severity-warnings]
+	 * : Include errors with lower severity than the threshold as other type.
```
ernilambar:
```suggestion
	 * : Include warnings with lower severity than the threshold as other type.
```

includes/CLI/Plugin_Check_Command.php
```diff
@@ -249,7 +259,7 @@ static function ( $dirs ) use ( $excluded_files ) {
 			$file_results = $this->flatten_file_results( $file_errors, $file_warnings );
 
 			if ( '' !== $error_severity || '' !== $warning_severity ) {
-				$file_results = $this->get_filtered_results_by_severity( $file_results, intval( $error_severity ), intval( $warning_severity ) );
+				$file_results = $this->get_filtered_results_by_severity( $file_results, intval( $error_severity ), intval( $warning_severity ), $include_low_severity_errors, $include_low_severity_warnings );
```
ernilambar, line 262: This new implementation should be done for warnings also. Around line 275.

tests/behat/features/plugin-check-severity.feature
```diff
@@ -192,5 +192,33 @@ Feature: Test that the severity level in plugin check works.
...
+      """
+      allow_unfiltered_uploads_detected,ERROR,7
+      """
+    And STDOUT should contain:
+      """
+      upgrade_notice_limit,WARNINGS_LOW_SEVERITY,5
+      """
+    And STDOUT should contain:
+      """
+      default_readme_text,WARNING,7
```
ernilambar: This error code `default_readme_text` has been converted to ERROR few weeks back.

ernilambar:
> It's strange as it works BEHAT tests in my local installation, but is failing.

`composer run behat -- tests/behat/features/plugin-check-severity.feature` is not passing locally in my local setup.

includes/CLI/Plugin_Check_Command.php
davidperezgar: ok

davidperezgar:
Ok Thanks Nilambar

davidperezgar:
Finished!

includes/CLI/Plugin_Check_Command.php
ernilambar: @davidperezgar This is not fixed in the PR.

davidperezgar: you're right.

ernilambar (approved):
Looks good.

ernilambar:
CC @felixarntz for re-review. 😊

includes/CLI/Plugin_Check_Command.php
```diff
@@ -626,25 +636,32 @@ private function display_results( $formatter, $file_name, $file_results ) {
...
 	 * @since 1.1.0
 	 *
-	 * @param array $results          Check results.
-	 * @param int   $error_severity   Error severity level.
-	 * @param int   $warning_severity Warning severity level.
+	 * @param array $results                     Check results.
+	 * @param int   $error_severity              Error severity level.
+	 * @param int   $warning_severity            Warning severity level.
+	 * @param bool  $include_low_severity_errors Include less level of severity issues as warning.
+	 * @param bool  $include_low_severity_warnings Include less level of severity issues as warning.
```
```suggestion
	 * @param array $results                       Check results.
	 * @param int   $error_severity                Error severity level.
	 * @param int   $warning_severity              Warning severity level.
	 * @param bool  $include_low_severity_errors   Include less level of severity issues as warning.
	 * @param bool  $include_low_severity_warnings Include less level of severity issues as warning.
```

felixarntz (approved):
Just one note on WPCS, otherwise LGTM!

davidperezgar:
Thanks!

dd32:
Any reason these output types are pluralised?

ie. The types are now:
 -  ERROR
 - ERRORS_LOW_SEVERITY
 - WARNING
 - WARNING_LOW_SEVERITY

ernilambar:
Followup PR - https://github.com/WordPress/plugin-check/pull/818

PR #701, opened 2024-10-06
Fixes SPDX warning

### LINKED ISSUE #698: Checker seems to be returning invalid license warning
As an user says in the forum, it giving an error with the License. Let's investigate that.

> The checker seems to be returning an invalid_license warning for valid licenses. E.g., this line triggers a warning:
> 
> Main plugin file
>  * License: GNU General Public License (GPL) version 3
> 
> readme.txt
> License: GNU General Public License (GPL) version 3
> There is a suggestion to use an SPDX license identifier from [here](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses), but the above is valid according to the link.
> 
> Interestingly, this passes the checks though it is not present in the link:
> 
> Main plugin file
>  * License: GPLv3 or later
> 
> readme.txt
> License: GPLv3 or later

davidperezgar:
Fixes #698

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
@@ -419,6 +419,7 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 		$license = str_replace( '-', '', $license );
 		$license = str_replace( 'GNU General Public License (GPL)', 'GPL', $license );
 		$license = str_replace( 'GNU General Public License', 'GPL', $license );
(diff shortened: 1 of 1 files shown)


ernilambar (approved):

felixarntz (approved):
Thanks @davidperezgar!

One additional suggestion: Further up in the `Plugin_Readme_Check.php` file, maybe we should add another string `(or later)`? I've seen several plugins use that, i.e. including parentheses.

davidperezgar:
Ok, I didn't see it. Yes. Let's do it.

PR #708, opened 2024-10-07
Changelog and changes for version 1.2.0

davidperezgar:
Updated Changelog.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -70,7 +70,6 @@ In any case, passing the checks in this tool likely helps to achieve a smooth pl
 
 = 1.2.0 =
 
-- The Plugin Check is now part of the automatic checks for plugin submissions on WordPress.org.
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -68,6 +68,29 @@ In any case, passing the checks in this tool likely helps to achieve a smooth pl
 
 == Changelog ==
 
+= 1.2.0 =
+
+- The Plugin Check is now part of the automatic checks for plugin submissions on WordPress.org.
```
felixarntz:
```suggestion
* The Plugin Check is now part of the automatic checks for plugin submissions on WordPress.org.
```

felixarntz (approved):

readme.txt
swissspidy: This line is not reallly a change in the plugin itself though, so maybe it could also be removed entirely?

mukeshpanchal27: Agree with @swissspidy

davidperezgar: ok

ernilambar (approved):

PR #711, opened 2024-10-09
Added check for case sensitive duplicated file names

### LINKED ISSUE #710: Check: Look for case sensitive duplicated filenames
This is a splited issue from #628 . We need to check if the user is using same name but with different capital letters. That causes a conflict as some servers don't differentiate that letters.

davidperezgar:
Fixes #710


davidperezgar:
@swissspidy this is the part of badly names that suggested @ernilambar to move it away, so we can work on reflection tests.

ernilambar:
@davidperezgar I have updated the test for the check. But I noticed current check does not account for the duplicated folder name. Example:

```
UNIT_TESTS_PLUGIN_DIR . 'test-plugin-file-type-badly-named-files-errors/sub directory/file-1.php'
UNIT_TESTS_PLUGIN_DIR . 'test-plugin-file-type-badly-named-files-errors/Sub Directory/file-2.php'
```

I think we should also check for folder. We should refine the check along with error code - "duplicated_files"

davidperezgar:
Hey awesome! It's ready then to review!

davidperezgar:
Updated check for duplicated folders.

ernilambar (approved):

swissspidy:
This check and its error messages are a bit confusing. It says „duplicated file names are not permitted“, which could led one to believe that „foo/bar.php“ und „baz/bar.php“ is not OK. Bit this is only about mixed case like „bar.php“ and „BAR.php“. I recommend clarifying the error messages here, maybe even include an example in it.

davidperezgar:
Ok!

davidperezgar:
Yes, you're right. The error message wasn't good enough. I've clarified the Case-Sensitive idea.

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -302,6 +302,49 @@ function ( $file ) use ( $plugin_path ) {
...
+			$folders[] = $folder;
+		}
+		$folders                = array_unique( $folders );
+		$folders_lowercase      = array_map( 'strtolower', $folders );
+		$case_sensitive_folders = array_unique( array_diff_assoc( $folders_lowercase, array_unique( $folders_lowercase ) ) );
+
+		if ( ! empty( $case_sensitive_folders ) ) {
+			$this->add_result_error_for_file(
+				$result,
+				__( 'Case-Sensitive duplicated folders were found. They\'ll be problematic in different systems.', 'plugin-check' ),
```
swissspidy:
```suggestion
				__( 'Multiple folders with the same name but different case were found. This can be problematic on certain file systems.', 'plugin-check' ),
```

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -302,6 +302,49 @@ function ( $file ) use ( $plugin_path ) {
...
+		}
+
+		// Duplicated names in same folder.
+		$files_lowercase      = array_map( 'strtolower', $files );
+		$case_sensitive_files = array_unique( array_diff_assoc( $files_lowercase, array_unique( $files_lowercase ) ) );
+
+		if ( ! empty( $case_sensitive_files ) ) {
+			$this->add_result_error_for_file(
+				$result,
+				__( 'Case-Sensitive duplicated files were found. They\'ll be problematic in different systems.', 'plugin-check' ),
```
swissspidy:
```suggestion
				__( 'Multiple files with the same name but different case were found. This can be problematic on certain file systems.' ),
```

swissspidy (approved):

PR #715, opened 2024-10-11
Check including libraries already in wp core including jquery

### LINKED ISSUE #702: Check: Including Libraries Already In WP Core (including jquery)
We warn developer to not reinvent the wheel. They should use the external libraries that WordPress has already.

As we say in the review:
> ## Including Libraries Already In Core
> 
> Your plugin has included a copy of a library that WordPress already includes.
> 
> WordPress includes a number of useful libraries, such as jQuery, Atom Lib, SimplePie, PHPMailer, PHPass, and more. For security and stability reasons, plugins may not include those libraries in their own code, but instead must use the versions of those libraries packaged with WordPress.
> 
> You can see the list of JS Libraries here:
> 
> https://developer.wordpress.org/reference/functions/wp_enqueue_script/#default-scripts-and-js-libraries-included-and-registered-by-wordpress
> 
> While we do not YET have a decent public facing page to list all these libraries, we do have a list here: 
> 
> https://meta.trac.wordpress.org/browser/sites/trunk/api.wordpress.org/public_html/core/credits/wp-59.php#L739
> 
> It’s fine to locally include add-ons not in core, but please ONLY add those additional files. For example, you do not need the entire jquery UI library for one file. If your code doesn't work with the built-in versions of jQuery, it's most likely a noConflict issue.

davidperezgar:
Fixes #702


davidperezgar:
How can i disable the spell check in one line?

ernilambar:
> How can i disable the spell check in one line?

This seems to false positive and issue seems to be filed https://github.com/crate-ci/typos/issues/1103

Not sure when that will be fixed. For current fix we could add toml file.

Documentation: https://github.com/crate-ci/typos?tab=readme-ov-file#false-positives

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
```diff
@@ -127,4 +127,22 @@ public function test_run_with_badly_named_errors() {
...
+
+		$errors = $check_result->get_errors();
+
+		$this->assertNotEmpty( $errors );
+		$this->assertEquals( 1, $check_result->get_error_count() );
+
+		// Check for core PHPMailer.
+		$this->assertArrayHasKey( 0, $errors['PHPMailer.php'] );
+		$this->assertArrayHasKey( 0, $errors['PHPMailer.php'][0] );
+		$this->assertCount( 1, wp_list_filter( $errors['PHPMailer.php'][0][0], array( 'code' => 'library_core_files' ) ) );
```
ernilambar, line 146: We could also add test for jquery library file inclusion as it is most recurring issue.

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -304,7 +382,7 @@ function ( $file ) use ( $plugin_path ) {
 	 * @return string Description.
 	 */
 	public function get_description(): string {
-		return __( 'Detects the usage of hidden and compressed files, VCS directories, application files and badly named files.', 'plugin-check' );
+		return __( 'Detects the usage of hidden and compressed files, VCS directories, application files,badly named files and library core files.', 'plugin-check' );
```
ernilambar: Space missing before `badly`.

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -294,6 +298,80 @@ function ( $file ) use ( $plugin_path ) {
 		}
 	}
 
+	/**
+	 * Looks for library core files and amends the given result with an error if found.
+	 *
+	 * @since 1.2.0
```
ernilambar:
```suggestion
	 * @since 1.3.0
```

tests/phpunit/tests/Checker/Checks/File_Type_Check_Tests.php
davidperezgar: yes, that's important one.

davidperezgar:
Thanks @ernilambar

ernilambar (approved):

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
```diff
@@ -304,7 +382,7 @@ function ( $file ) use ( $plugin_path ) {
 	 * @return string Description.
 	 */
 	public function get_description(): string {
-		return __( 'Detects the usage of hidden and compressed files, VCS directories, application files and badly named files.', 'plugin-check' );
+		return __( 'Detects the usage of hidden and compressed files, VCS directories, application files, badly named files and library core files.', 'plugin-check' );
```
swissspidy:
```suggestion
		return __( 'Detects the usage of hidden and compressed files, VCS directories, application files, badly named files and core library files.', 'plugin-check' );
```

Not a native speaker but this sounds more correct

swissspidy (approved):

includes/Checker/Checks/Plugin_Repo/File_Type_Check.php
ernilambar: All naming is based on that: `Library Core Files`. Should we rename those accordingly?

PR #716, opened 2024-10-11
Don't flag Apache license

swissspidy:
Apache 2.0 is a GPL compatible license and there are many Apache 2.0 licensed plugins in the repository. It should not be flagged by PCP.


felixarntz (approved):
Good catch!

ernilambar:
I have found few instances of MPL 2.0 also in few plugins. https://www.gnu.org/licenses/license-list.en.html#MPL-2.0

davidperezgar (approved):
Ok. Apache is allowed.

davidperezgar:
@ernilambar we would need to ask in the Team if it's allowed. We're not lawyers, so it would be fine if someone confirms that.

PR #720, opened 2024-10-14
Restricted Headers: remove ones that are allowed, but leave the mech for checking in place for future use

### LINKED ISSUE #669: Restrict custom plugin headers

### LINKED ISSUE #718: Git Updater headers now listed as an ERROR for `plugin_header_restricted_fields`
I'm not sure when this check was added but I think it should be removed or at least made much less stringent. These are plugin headers or comments. I do understand a check for `Updates URI` for a repository plugin but I don't understand why it would list `GitHub Plugin URI` as an error.

Image

I would like some clarification and guidance. Thanks.

kraftbj:
Partly reverts the functionality of #670.
Resolves #718 

Per the conversation in #669, the presence of these headers—in and of themselves—do not create a third-party upgrader situation. The UpdateURI, which is already checked elsewhere, does prevent the updates from WordPress.org, but the headers originally included in this code are for information only.

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
@@ -69,7 +69,9 @@ class Plugin_Header_Fields_Check implements Static_Check {
 			'RequiresPlugins' => 'Requires Plugins',
 		);
 
-		$restricted_labels = array(); // Reserved for future use.
+		$restricted_labels = array(
+			'RestrictedLabel' => 'Restricted Label',
+		); // Reserved for future use.
 
 		$plugin_header = $this->get_plugin_data( $plugin_main_file, array_merge( $labels, $restricted_labels ) );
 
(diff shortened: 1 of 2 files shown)


afragen:
Thank you Brandon.

sc0ttkclark (approved):
This makes a lot of sense and keeps the functionality available for problematic comments that could be moved in here like the one which lets you override the `Update URI` (https://make.wordpress.org/core/2021/06/29/introducing-update-uri-plugin-header-in-wordpress-5-8/)

danieliser:
I approve this! ☝️

xwolfde:
Thank you!

dd32:
I've chatted with @chriscct7 and I agree with reverting this change for now, assigning it to him to handle.

dd32 (approved):

chriscct7 (approved):

lucprincen:
Thank you!

PR #721, opened 2024-10-15
Add new image functions check

### LINKED ISSUE #26: Create Image_Functions_Usage_Check
## Description

Checks and warns against the usage of image HTML in templates. Developers are encouraged to use image functions such as wp_get_attachment_image that generate image HTML in templates. The HTML output by these functions is often better as it generates the tag with more performant attributes such as srcset and sizes. No existing sniff exists but the existing WordPress.WP.EnqueuedResources sniff in the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards/) can be used as a starting point.

## Acceptance Criteria

- Class `Image_Functions_Usage_Check` should be created and exists at `includes/Checker/Checks`
- Class `Image_Functions_Usage_Check` should extend the `PHPCodesniffer_Check` abstract class which extends `Check` interface
- Method `run` should be implemented from the `PHPCodesniffer_Check` abstract class if needed
- `$_SERVER['argv']` arguments should be passed to parent `run` method to provide PHPCS information about `standard`, `report` and `sniff`
- Custom sniff to detect the non-usage of wp image functions 

## Tests Coverage
- `Image_Functions_Usage_Check` class expected to extend `PHPCodesniffer_Check` abstract class
- `Image_Functions_Usage_Check` class expected to be instance of the `PHPCodesniffer_Check` class
- `run` method expected to append results of check into `$check_result`

swissspidy:
Checks and warns against the usage of image HTML in templates. Developers are encouraged to use image functions such as wp_get_attachment_image that generate image HTML in templates

Fixes #26

### DIFF
--- a/includes/Checker/Checks/Performance/Image_Functions_Check.php
+++ b/includes/Checker/Checks/Performance/Image_Functions_Check.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Class Enqueued_Resources_Check.
+ * Class Image_Functions_Check.
  *
  * @package plugin-check
  */
@@ -13,11 +13,11 @@ use WordPress\Plugin_Check\Checker\Checks\Abstract_PHP_CodeSniffer_Check;
 use WordPress\Plugin_Check\Traits\Stable_Check;
 
 /**
- * Check for running WordPress enqueued resources sniffs.
+ * Check for running WordPress image functions sniffs.
  *
- * @since 1.0.2
+ * @since 1.3.0
  */
-class Enqueued_Resources_Check extends Abstract_PHP_CodeSniffer_Check {
+class Image_Functions_Check extends Abstract_PHP_CodeSniffer_Check {
 
 	use Stable_Check;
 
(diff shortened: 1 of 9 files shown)


includes/Checker/Checks/Performance/Image_Functions_Check.php
```diff
@@ -0,0 +1,78 @@
...
+	 * Gets the documentation URL for the check.
+	 *
+	 * Every check must have a URL with further information about the check.
+	 *
+	 * @since 1.3.0
+	 *
+	 * @return string The documentation URL.
+	 */
+	public function get_documentation_url(): string {
+		return __( 'https://developer.wordpress.org/plugins/', 'plugin-check' );
```
felixarntz, line 76: Is there no more precise documentation URL about this? Probably would be worth covering this topic somewhere.

felixarntz (approved):
@swissspidy Thanks for implementing this, looks great!

includes/Checker/Checks/Performance/Image_Functions_Check.php
swissspidy: Couldn‘t find one at first glance. A lot of our checks need more detailed docs, IIRC we previously discussed having a dedicated page in that handbook for Plugin Check or so. Once we have a better page we can update the link here.

davidperezgar: Yes, you can ask for a section in the Developer Handbook.

davidperezgar (approved):

PR #722, opened 2024-10-15
Modifies i18n_usage check to consider explicit default textdomain usage a warning

### LINKED ISSUE #713: Plugin i18n check complains about `default` text domain
I just noticed that as of 1.2.0, the Plugin Check complains about usage of the `default` text domain in plugins. I believe that this is not correct. It is explicitly allowed to use WordPress Core's text domain if it is for strings that are taken directly from Core. Not doing so creates unnecessary work for translators.

It looks like the relevant change was made in #681. I believe we need to update that code so that in addition to the plugin's own text domain we pass `default` there so that it is allowed too.

Here's the example where I noticed this check suddenly failing: https://github.com/felixarntz/ai-services/actions/runs/11282317040/job/31379501989?pr=11

It is worth noting that this was also automatically flagged during the review of that plugin when I submitted it to the directory, however after clarification it was allowed. So this confirms to me that this is simply a bug in the tooling, and we should allow `default`.

cc @davidperezgar @ernilambar

felixarntz:
Fixes #713.

Note: There's no way to do this via WPCS flags, so we need to rely on the error message to detect the relevant violation and downgrade it to a warning accordingly. This can be done with minimal additional code by overriding the relevant parent method.

### DIFF
--- a/includes/Checker/Checks/General/I18n_Usage_Check.php
+++ b/includes/Checker/Checks/General/I18n_Usage_Check.php
@@ -81,4 +81,28 @@ class I18n_Usage_Check extends Abstract_PHP_CodeSniffer_Check {
 	public function get_documentation_url(): string {
 		return __( 'https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/', 'plugin-check' );
 	}
+
+	/**
+	 * Amends the given result with a message for the specified file, including error information.
+	 *
+	 * @since 1.3.0
+	 *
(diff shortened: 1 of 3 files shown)


swissspidy (approved):
Still not a fan of projects doing this but this is better than the current experience I suppose.

ernilambar (approved):

ernilambar:
> Still not a fan of projects doing this but this is better than the current experience I suppose.

Agree. Personally I would want my plugin's string independent from core string and have full control to the plugin itself.

PR #725, opened 2024-10-18
WIP: Low severity threshold

### LINKED ISSUE #691: Show errors with severity less than a number as a warning

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/691

- Introduce flag in CLI for including low severity errors.
- Behat test added for testing new flag

### DIFF
--- a/includes/CLI/Plugin_Check_Command.php
+++ b/includes/CLI/Plugin_Check_Command.php
@@ -703,23 +703,29 @@ final class Plugin_Check_Command {
 		return array_merge( $errors, $warnings );
 	}
 
-	private function get_low_severity_errors( $file_results, $error_severity ) {
+	/**
+	 * Returns low severity errors.
+	 *
+	 * @since 1.3.0
+	 *
+	 * @param array $results        Check results.
+	 * @param int   $error_severity Error severity level.
+	 * @return array Filtered low severity errors.
+	 */
+	private function get_low_severity_errors( $results, $error_severity ) {
 		$low_severity_errors = array_filter(
-			$file_results,
+			$results,
 			function ( $item ) use ( $error_severity ) {
 				return ( 'ERROR' === $item['type'] && $item['severity'] < $error_severity );
 			}
 		);
 
-		$low_severity_errors = array_map(
+		return array_map(
 			function ( $item ) {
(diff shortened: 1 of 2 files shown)


ernilambar:
Example:
Before with no severity argument
```
FILE: pc-sample.php
type,code,severity
ERROR,allow_unfiltered_uploads_detected,7
ERROR,plugin_header_non_functional_plugin_uri,7
ERROR,localhost_code_detected,8
ERROR,obfuscated_code_detected,6
ERROR,WordPress.WP.I18n.MissingTranslatorsComment,5
ERROR,WordPress.WP.I18n.TextDomainMismatch,5
ERROR,WordPress.WP.I18n.NonSingularStringLiteralDomain,5
ERROR,WordPress.Security.EscapeOutput.OutputNotEscaped,5
```

With `--error-severity=7`:

```
FILE: pc-sample.php
type,code,severity
ERROR,allow_unfiltered_uploads_detected,7
ERROR,plugin_header_non_functional_plugin_uri,7
ERROR,localhost_code_detected,8
```

With `--error-severity=7 --include-low-severity-errors`:

```
FILE: pc-sample.php
type,code,severity
ERROR,allow_unfiltered_uploads_detected,7
ERROR,plugin_header_non_functional_plugin_uri,7
ERROR,localhost_code_detected,8
ERROR_EXTRA,obfuscated_code_detected,6
ERROR_EXTRA,WordPress.WP.I18n.MissingTranslatorsComment,5
ERROR_EXTRA,WordPress.WP.I18n.TextDomainMismatch,5
ERROR_EXTRA,WordPress.WP.I18n.NonSingularStringLiteralDomain,5
ERROR_EXTRA,WordPress.Security.EscapeOutput.OutputNotEscaped,5

```

PR #743, opened 2024-10-24
Add localhost sniff

### LINKED ISSUE #267: Do not raise error "Do not use Localhost in your code" when "localhost" in comment

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/267

- Add sniff `PluginCheck.CodeAnalysis.Localhost`
- Update `Localhost_Check` to use new sniff
- Localhost in comments will be ignored now

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Localhost_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Localhost_Check.php
@@ -9,7 +9,7 @@ namespace WordPress\Plugin_Check\Checker\Checks\Plugin_Repo;
 
 use WordPress\Plugin_Check\Checker\Check_Categories;
 use WordPress\Plugin_Check\Checker\Check_Result;
-use WordPress\Plugin_Check\Checker\Checks\Abstract_File_Check;
+use WordPress\Plugin_Check\Checker\Checks\Abstract_PHP_CodeSniffer_Check;
 use WordPress\Plugin_Check\Traits\Amend_Check_Result;
 use WordPress\Plugin_Check\Traits\Stable_Check;
 
@@ -18,7 +18,7 @@ use WordPress\Plugin_Check\Traits\Stable_Check;
  *
  * @since 1.0.0
  */
-class Localhost_Check extends Abstract_File_Check {
+class Localhost_Check extends Abstract_PHP_CodeSniffer_Check {
 
 	use Amend_Check_Result;
 	use Stable_Check;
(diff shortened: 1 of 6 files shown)


phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/LocalhostSniff.php
```diff
@@ -0,0 +1,89 @@
...
+
+use PHP_CodeSniffer\Util\Tokens;
+use WordPressCS\WordPress\Sniff;
+
+/**
+ * Detect localhost.
+ *
+ * @link https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
+ *
+ * @since 1.1.0
```
swissspidy:
```suggestion
 * @since 1.3.0
```

swissspidy (approved):

includes/Checker/Checks/Plugin_Repo/Localhost_Check.php
```diff
@@ -93,4 +81,26 @@ public function get_description(): string {
...
+	 * @param bool         $error    Whether it is an error or notice.
+	 * @param string       $message  Error message.
+	 * @param string       $code     Error code.
+	 * @param string       $file     Absolute path to the file where the issue was found.
+	 * @param int          $line     The line on which the message occurred. Default is 0 (unknown line).
+	 * @param int          $column   The column on which the message occurred. Default is 0 (unknown column).
+	 * @param string       $docs     URL for further information about the message.
+	 * @param int          $severity Severity level. Default is 5.
+	 */
+	protected function add_result_message_for_file( Check_Result $result, $error, $message, $code, $file, $line = 0, $column = 0, string $docs = '', $severity = 5 ) {
```
mukeshpanchal27, line 100: Why the `$severity` parameter needed if we want to override it by `8`?

ernilambar: Kept same arguments and doc for consistency. We could omit that if needed as it is optional argument.

davidperezgar (approved):

PR #744, opened 2024-10-24
Allow custom checks to provide `installed_paths`

### LINKED ISSUE #740: Addon should be able to use sniff from external repo
I am developing PCP addon which will use external sniff. Those are setup properly in the addon. But PHPCS runner does not detect such coding standard installed from the addon.

This is a minimal PCP addon for example: https://github.com/ernilambar/tiny-check

Coding standard I am trying to use: https://github.com/ernilambar/coding-standard/

`class Todo_Check extends Abstract_PHP_CodeSniffer_Check`

```php
return array(
	'extensions'  => 'php',
	'standard'    => 'NilambarCodingStandard',
	'sniffs'      => 'NilambarCodingStandard.Commenting.TodoComment',
);
```

This rule is not detected by our runner. Gives sniff not registered error.

For debugging, I did composer require this standard in PCP, and then this sniff is detected and check is working properly.

Is there a way to make addon's sniffs recognizable to the main PCP plugin?

swissspidy:
And then in your addon you do:

```php
	protected function get_args( Check_Result $result ) {
		return array(
			'extensions'  => 'php',
			'standard'    => 'NilambarCodingStandard',
			'sniffs'      => 'NilambarCodingStandard.Commenting.TodoComment',
			'installed_paths' => [
				__DIR__ . '/vendor/ernilambar/coding-standard',
				__DIR__ . '/vendor/wp-coding-standards/wpcs',
			],
		);
	}
```

cc @ernilambar 

Fixes #740

### DIFF
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -58,6 +58,8 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 	/**
 	 * Amends the given result by running the check on the associated plugin.
 	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
+	 *
 	 * @since 1.0.0
 	 *
 	 * @param Check_Result $result The check result to amend, including the plugin context to check.
--- a/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
+++ b/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
@@ -79,6 +79,12 @@ abstract class Abstract_PHP_CodeSniffer_Check implements Static_Check {
 			);
 		}
 
+		if ( ! class_exists( '\PHP_CodeSniffer\Config' ) ) {
+			throw new Exception(
+				__( 'Unable to find PHPCS Config class.', 'plugin-check' )
+			);
+		}
+
 		// Backup the original command line arguments.
 		$orig_cmd_args = $_SERVER['argv'] ?? '';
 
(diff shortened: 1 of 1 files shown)


includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -81,14 +82,24 @@ final public function run( Check_Result $result ) {
 		// Backup the original command line arguments.
 		$orig_cmd_args = $_SERVER['argv'] ?? '';
 
+		$args = $this->get_args( $result );
+
+		// Reset PHP_CodeSniffer config.
+		$this->reset_php_codesniffer_config();
+
+		// Get current installed_paths config.
+		$installed_paths = Config::getConfigData( 'installed_paths' );
```
swissspidy, line 99: Why does PHPStan complain about `Config` not being found, but not about the `Runner` class? 🤔

ernilambar:
Great. It is working as expected. No idea about the PHPStan issue.

swissspidy:
Missed the `class_exists` check for the `Runner` class, added one for `Config` now too.

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
```diff
@@ -78,17 +81,33 @@ final public function run( Check_Result $result ) {
 			);
 		}
 
+		if ( ! class_exists( '\PHP_CodeSniffer\Config' ) ) {
```
felixarntz: Nit-pick:
```suggestion
		if ( ! class_exists( Config::class ) ) {
```

felixarntz (approved):
@swissspidy Thanks, looks great! Just one non-blocking suggestion.

ernilambar (approved):

includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php
swissspidy: FWIW it's the exact same check as for the Runner class just a few lines above. I'd need to change it for both.

felixarntz: It would be cleaner to do so, rather than hard-coding the fully qualified class name again. Not critical obviously, but also there's no good reason not to :)

I saw it's also hard-coded in the reset method further down in the class.

PR #746, opened 2024-10-25
License check improvements

### LINKED ISSUE #737: Error `no_license` not triggered with missing readme file
When there is no readme file, "no_license" is not triggered even when there is no "License" header in plugin main file.

This is happening because license check is under `Plugin_Readme_Check`. May be we should separate license check for plugin header and readme file.

Check `Plugin_Header_Fields_Check` should be better place for License plugin header check.

Also same error code `no_license` is used for missing license in readme and plugin header which is confusing.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/737

* Add `License_Utils` trait
* New trait is used in `Plugin_Readme_Check` and `Plugin_Header_Fields_Check`
* Add new error codes: `plugin_header_no_license` and `plugin_header_invalid_license` (Earlier same error code `no_license` was used for missing license in readme and plugin header which is confusing)
* `license_mismatch` error code is moved from `Plugin_Readme_Check` to `Plugin_Header_Fields_Check`
* Update unit tests to reflect changes
* Add unit test for invalid license
* Add scenario for missing license in single file plugin

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
@@ -10,19 +10,24 @@ namespace WordPress\Plugin_Check\Checker\Checks\Plugin_Repo;
 use Exception;
 use WordPress\Plugin_Check\Checker\Check_Categories;
 use WordPress\Plugin_Check\Checker\Check_Result;
-use WordPress\Plugin_Check\Checker\Static_Check;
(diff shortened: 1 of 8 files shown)


ernilambar:
This PR is getting bigger and bigger. So I have created separate PR for license check infrastructure. Will work on this PR after rebasing after https://github.com/WordPress/plugin-check/pull/761 is merged and https://github.com/WordPress/plugin-check/issues/762 is finalized.

davidperezgar:
Ok!

PR #747, opened 2024-10-25
Fix slug for single-file plugins in both admin and CLI

### LINKED ISSUE #745: Single file plugin text domain check yields mixed and incorrect expected strings
For a single file plugin with a text domain set, the `WordPress.WP.I18n.TextDomainMismatch` check relies on a plugin directory slug (non-existent in this context) and shows different results between the UI and CLI. It has been suggested that this check [be bypassed for single-file plugins](https://github.com/WordPress/plugin-check/issues/713#issuecomment-2434285443) (hat tip @ernilambar).

To test this, I added the `hello-dolly` text domain to the `hello.php` plugin that ships with WP core (required to avoid an error with the `WordPress.WP.I18n.MissingArgDomain` check):

```php
__( 'Quote from Hello Dolly song, by Jerry Herman:', 'hello-dolly' ),
```

Running PCP, the admin shows
  `Mismatched text domain. Expected 'plugins' but got 'hello-dolly'.`
whereas the CLI returns
  `Mismatched text domain. Expected 'hello.php' but got 'hello-dolly'.`

Originally reported at https://github.com/WordPress/plugin-check/issues/713#issuecomment-2408257109.

swissspidy:
Do not try to guess slug in `Abstract_Check_Runner`. This can be handled by `Plugin_Context`.

This way both admin and CLI will determine the same slug for the same plugin.

In `Plugin_Context`, use correct slug for single-file plugins. For `hello.php` it would be `hello`.

Fixes #745


swissspidy:
OK the Behat test require some work when WP isn't fully loaded yet 😄

ernilambar:
> OK the Behat test require some work when WP isn't fully loaded yet 😄

@swissspidy Should we add polyfill function for `plugin_dir_path()` ?

swissspidy:
It's just `trailingslashit( dirname() )`, nothing magic

ernilambar (approved):

davidperezgar (approved):

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -627,7 +627,7 @@ final protected function get_slug() {
 	private function get_check_context() {
 		$plugin_basename = $this->get_plugin_basename();
 		$plugin_path     = is_dir( $plugin_basename ) ? $plugin_basename : WP_PLUGIN_DIR . '/' . $plugin_basename;
-		return new Check_Context( $plugin_path, $this->get_slug() );
+		return new Check_Context( $plugin_path, $this->get_slug() ? $this->get_slug() : basename( $plugin_path, '.php' ) );
```
felixarntz, line 630: I think this should happen within `Plugin_Context` (which `Check_Context` extends).

That's where there's already logic to determine the slug by default, if none was specified. It doesn't make sense to me to have that duplicated in this class, neither here nor in the `set_slug()` method where it was so far.

I think it makes most sense to have `get_slug()` pass through the empty string that `get_slug_param()` may return (for AJAX always, for CLI only if no slug was explicitly specified). And then because that string is empty, `Plugin_Context` would define the default slug as the source of truth.

Alternatively, if there's value to have the logic here in `Abstract_Check_Runner`, we should remove the logic from `Plugin_Context` and make the constructor `$slug` parameter required (and throw an exception if it's empty). It just doesn't make sense to have the logic split/duplicated between two places.

felixarntz (changes requested):
Conceptually the fix looks good, but I think we should consolidate some duplicate / fragmented logic here.

includes/Checker/Abstract_Check_Runner.php
swissspidy: I just moved this there in 16f8e0998a9d882481464709148dab7380b750c3 because otherwise it didn't work when using WP-CLI "shortinit" with `--require`. We don't pass enough information to `Check_Context`/`Plugin_Context` on its own to determine whether it's a single-file plugin or not.

Feel free to make edits to the PR with what you have in mind.


----

Aside: whenever we have to add workarounds like this I'm wondering whether we actually really want to support single-file plugins or not. They aren't allowed in the repo anyway.

felixarntz: How does WP-CLI "shortinit" work? Not familiar with it... How does it prevent that logic in the constructor from working correctly?

swissspidy: I meant the `--require` hack we use with the object-cache drop-in. Then a lot of the WP constants and functions won't be available that early.  See https://github.com/WordPress/plugin-check/actions/runs/11660800826/job/32463868033?pr=747 for example.

felixarntz: Thanks for clarifying. I can have a look later this week to see if I can come up with a solution for that.

ernilambar:
> Aside: whenever we have to add workarounds like this I'm wondering whether we actually really want to support single-file plugins or not. They aren't allowed in the repo anyway.

Agree on this one. There are lots of workarounds for the single file plugin.

includes/Checker/Abstract_Check_Runner.php
davidperezgar: @felixarntz could you see it?

PR #749, opened 2024-10-25
CLI: Fix confusing runtime environment setup order

felixarntz:
Related to #597.

As of today, in WP-CLI the `Runtime_Environment_Setup` sets up the separate database tables _after_ all the regular preparations have run (i.e. `Universal_Runtime_Preparation`). This is confusing and inconsistent with how it works in the AJAX based workflow, where the `Runtime_Environment_Setup` is executed in a previous request. This is how it was intended, as ideally setting up the separate database tables to run runtime checks against needs to be the earliest step. However, in WP-CLI this has not been happening so far, because it was invoked by the command implementation itself, which however is executed _after_ the early initialization process.

For example, today the `option_active_plugins` filter is applied on the _real_ database tables for WP-CLI, which means the separate database tables will only contain the filtered plugins. In the AJAX based flow on the other hand, all active plugins from the _real_ database tables will be added to the separate database tables, and the `option_active_plugins` filter is only applied then. This is, to my knowledge, not causing an actual problem today, but it's just strange and error-prone to have to consider both execution orders.

I had mentioned in [this comment](https://docs.google.com/document/d/1wDGZBwWB2WAxfbHE3lygIzQFK8IssCa5apOyaBolukQ/edit?disco=AAABUy8TaSI) that it's "just a technical nuance", but it's still weird and inconsistent, and it doesn't have to be.

This PR fixes it:
* Abstract relevant runtime initialization in the `Abstract_Check_Runner` class to a new protected method `initialize_runtime()`.
* In `CLI_Runner`, override that method to first execute `Runtime_Environment_Setup`, and of course also add its cleanup callback to the list of callbacks.
* This furthermore allows to simplify the `Plugin_Check_Command` class:
    * Remove the separate use of `Runtime_Environment_Setup` since it's now handled by `CLI_Runner`.
    * By doing that, we can also remove the private (and somewhat duplicated) `has_runtime_check()` method, as well as an extra call to the `get_checks_to_run()` method which only existed for that purpose.

It now all works through the `CLI_Runner`, i.e. less technical and cognitive overhead for the `Plugin_Check_Command` implementation.


felixarntz:
It took me a while to get the `wp_install()` call to work early, but I got it resolved now. So this PR is good for review!

includes/Checker/Abstract_Check_Runner.php
```diff
@@ -489,6 +487,18 @@ final public function get_checks_to_run() {
 		return $collection->to_map();
 	}
 
+	/**
+	 * Initializes the runtime environment so that runtime checks can be run against a separate set of database tables.
+	 *
+	 * @since n.e.x.t
```
swissspidy: Note that we don't use this pattern in this plugin, there's no script to replace them. Can just manually set it to 1.3.0.

```suggestion
	 * @since 1.3.0
```

includes/Checker/CLI_Runner.php
```diff
@@ -178,6 +178,31 @@ protected function get_slug_param() {
 		return $slug;
 	}
 
+	/**
+	 * Initializes the runtime environment so that runtime checks can be run against a separate set of database tables.
+	 *
+	 * @since n.e.x.t
```
swissspidy:
```suggestion
	 * @since 1.3.0
```

swissspidy (approved):
Appreciate all the digging into this! 👏

includes/Checker/CLI_Runner.php
felixarntz: Good point! Updated in https://github.com/WordPress/plugin-check/pull/749/commits/3dcd99c78e95636d79e07f6e59c991977c22e8f9

ernilambar (approved):

davidperezgar:
Awesome!

PR #751, opened 2024-10-25
Fix addon checks not being executed when running runtime checks

### LINKED ISSUE #607: Add end-to-end (e2e) tests
We already have some e2e tests for the WP-CLI entry point, but not for the admin entry point. Having e2e coverage for the Ajax flow would increase confidence in our setup.

Should be simple to set up, with 1 dummy plugin for testing and 1 addon installed.

Can use Playwright and wp-env or maybe even Playground

Scenarios to test:

* Running all checks and their output
* Running only runtime checks
* Running checks from an add-on
* Maybe more?

felixarntz:
**This is a child PR of #749. Please merge that one first.**

Fixes #607.

As noted prior, despite our various fixes related to e.g. #597, addon runtime checks still don't work as of today. This PR aims to fix that.

It looks like the underlying problem is that the `Default_Check_Repository` runs the `wp_plugin_check_checks` filter too early when any runtime checks are being run. Now that I noticed it, it seems like an obvious oversight, but was kind of hard to spot 😅 

To address this problem, we simply need to ensure the filter is run again after plugins have been loaded.

For the scenario where the `Default_Check_Repository` is instantiated when plugins are already loaded, no change is needed, in this case we can set a flag that the checks are "fully initialized" from the start.

### DIFF
--- a/tests/behat/features/plugin-check.feature
+++ b/tests/behat/features/plugin-check.feature
@@ -556,14 +556,14 @@ Feature: Test that the WP-CLI command works.
       ExampleRuntimeCheck.ForbiddenScript,WARNING
       """
 
-    # Same again, to verify object-cache.php was properly cleared again
-    When I run the WP-CLI command `plugin check foo-sample --fields=code,type --format=csv --require=./wp-content/plugins/plugin-check/cli.php`
+    # Same again but without requiring the cli.php file, to verify object-cache.php was properly cleared again
+    When I run the WP-CLI command `plugin check foo-sample --fields=code,type --format=csv`
     Then STDOUT should contain:
       """
       WordPress.WP.EnqueuedResourceParameters.NotInFooter,WARNING
       """
 
-    And STDOUT should contain:
+    And STDOUT should not contain:
       """
       ExampleRuntimeCheck.ForbiddenScript,WARNING
       """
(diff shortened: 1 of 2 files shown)


swissspidy:
> **This is a child PR of #749. Please merge that one first.**

Wanna rebase to make that more clear?

felixarntz:
> > **This is a child PR of #749. Please merge that one first.**
> 
> Wanna rebase to make that more clear?

Yes, I just had this open against `trunk` so that all the tests run and I can verify they pass. Just changed it back to the other branch.

Here you can see the Behat tests passing with the latest state of this PR, including all the ones that used to be commented out: https://github.com/WordPress/plugin-check/actions/runs/11525465349

tests/behat/features/plugin-check.feature
```diff
@@ -551,40 +551,36 @@ Feature: Test that the WP-CLI command works.
...
       WordPress.WP.EnqueuedResourceParameters.NotInFooter,WARNING
       """
-# This doesn't currently work, because we are not actually loading any other plugins, including pcp-addon.
-#    And STDOUT should contain:
-#      """
-#      ExampleRuntimeCheck.ForbiddenScript,WARNING
-#      """
+    And STDOUT should contain:
+      """
+      ExampleRuntimeCheck.ForbiddenScript,WARNING
```
swissspidy, line 556: Nice!

swissspidy (approved):

ernilambar (approved):

PR #755, opened 2024-10-27
Added support for MPL-2.0 SPDX license. Fixes #719

### LINKED ISSUE #719: Allow MPL 2.0 License
The MPL 2.0 is compatible with GPL, and there are few instances of MPL 2.0 also in few plugins.

More information about the license:
https://www.gnu.org/licenses/license-list.en.html#MPL-2.0

Not other versions are compatible, so we should develop only for that version.

Links:
https://www.tldrlegal.com/license/mozilla-public-license-2-0-mpl-2#:~:text=MPL%20is%20a%20copyleft%20license,GPL%20version%202%20or%20greater.

https://www.mozilla.org/en-US/MPL/2.0/FAQ/

syrusakbary:
This PR fixes #719, adding support for the MPL-2.0 [SPDX license](https://spdx.org/licenses/MPL-2.0.html).
This PR also includes an unit test to verify that the MPL-2.0 license is properly supported and doesn't trigger any errors.

> Note: This work was done as part of the WordCamp Granada 2024. Thanks @davidperezgar for the guidance!


syrusakbary:
Just linked my Github account into the [WordPress profile](https://profiles.wordpress.org/syruswasmer/). @davidperezgar it may be a good idea to add the `props-bot` label to the issue so the Github warning disappears :)

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -180,6 +180,36 @@ public function test_run_with_errors_no_license() {
...
+	public function test_run_without_error_mpl2_license() {
+		$readme_check  = new Plugin_Readme_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-plugin-readme-mpl2-license-without-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$readme_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		$this->assertEmpty( $errors );
```
ernilambar, line 192: This test is pretty fragile. This will fail when WP 6.8 is released even though MPL license is good. In that case `$errors` will not be empty. Alternative way is to make sure there is at least 1 error in the plugin and check that `invalid_license` is not there.

ernilambar: On second thought, lets keep like this in this PR. In separate PR I am working on separating license check for readme file and plugin header. In that PR, tests will be rewritten and few would be moved to another test file.

ernilambar (approved):

davidperezgar (approved):

davidperezgar:
Congrats Syrus for your first PR!

frantorres:
Congrats and thanks @syrusakbary ! 👍

PR #761, opened 2024-10-30
Add License trait

ernilambar:
* Add `License_Utils` trait
* New trait is used in `Plugin_Readme_Check`

This trait will be useful in other checks like `Plugin_Header_Fields_Check` in the future.


ernilambar:
@swissspidy 
I want to use private method `get_wordpress_stable_version()` which is currently in `Plugin_Readme_Check` in plugin header check `Plugin_Header_Fields_Check`. In the future it could be used in other checks also. Do you suggest adding single Trait for all utility functions or separate Traits?

swissspidy:
I'd probably start with separate ones

davidperezgar (approved):

swissspidy (approved):

PR #768, opened 2024-11-05
Always use prefixed tables during runtime check requests

felixarntz:
Related to #597 (see 2nd point in the description, also see [this doc comment](https://docs.google.com/document/d/1wDGZBwWB2WAxfbHE3lygIzQFK8IssCa5apOyaBolukQ/edit?disco=AAABUzHEE-g)).

While for the checks itself it's fine to only use the custom database prefix right when the checks are run, this poses some risks because certain parts of the environment (e.g. active plugins, active theme) are already filtered. This could lead to problematic side effects on the actual database tables depending on what plugins do - potentially disastrous, if it was a live site.

Therefore this PR sets the custom database prefix as part of overall runtime preparations (via `Universal_Runtime_Preparation`) so that it remains for the duration of more or less the entire request:
* The relevant process of setting the prefix in the `Abstract_Check_Runner::run()` method has been removed as it is now unnecessary there.
* This change led to the PHPUnit test `AJAX_Runner_Tests::test_prepare_with_runtime_check()` to fail, which however was exactly _because_ of the problem this PR is fixing: The `Use_Minimal_Theme_Preparation` may modify the database with the custom theme root, and this would so far have updated the _actual_ database. So the test now initially sets up the runtime environment, which is needed to execute runtime checks in any case, and the runtime preparations then operate on the "fake tables".
* A similar fix needed to be added to `Runtime_Check_UnitTestCase`, which was also missing the runtime environment setup bit.

Additionally, it makes a few minor code improvements:
* Move check whether runtime environment is set up into its own method on `Runtime_Environment_Setup`.
* Improve PHPUnit tests to only clean up what is actually needed, rather than a "catch-all" cleanup, as the latter could lead to potential cleanup related problems to go unnoticed.

### DIFF
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -159,20 +159,14 @@ final class Admin_AJAX {
 			wp_send_json_error( $valid_request, 403 );
 		}
 
-		// Set the new prefix.
-		$old_prefix = $wpdb->set_prefix( $table_prefix . 'pc_' );
-
 		$message = __( 'Runtime environment was not prepared, cleanup was not run.', 'plugin-check' );
 
-		// Test if the runtime environment tables exist.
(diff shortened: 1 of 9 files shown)


ernilambar (approved):
Nice.

swissspidy (approved):

davidperezgar (commented):
Awesome!

PR #775, opened 2024-11-07
Add separate license error codes for Plugin Header

ernilambar:
This PR implements second part of https://github.com/WordPress/plugin-check/issues/762

Add two error codes:

- plugin_header_no_license
- plugin_header_invalid_license

Now, error codes `no_license` and `invalid_license` refers only to the license check in readme.

### DIFF
--- a/tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
+++ b/tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
@@ -192,24 +192,6 @@ class Plugin_Readme_Check_Tests extends WP_UnitTestCase {
 		$this->assertEmpty( $errors );
 	}
 
-	public function test_run_with_errors_mpl1_license() {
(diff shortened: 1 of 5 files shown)


tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -192,24 +192,6 @@ public function test_run_without_error_mpl2_license() {
 		$this->assertEmpty( $errors );
 	}
 
-	public function test_run_with_errors_mpl1_license() {
```
ernilambar, line 195: This has been removed as it was checking license value from plugin header. The same plugin is used in `test_run_with_invalid_mpl1_license` (in Plugin_Header_Fields_Check_Tests) for checking for invalid license.

davidperezgar (approved):

frantorres (approved):

PR #780, opened 2024-11-10
Force slug in wp cli command to ensure that it get's the correct one by default

### LINKED ISSUE #774: [Bug] `TextDomainMismatch` if checking a plugin via path
**Expected behavior**

When running `wp plugin check path/to/my-plugin`, the checked text domain should be `my-plugin`

**Actual behavior**

When running `wp plugin check path/to/my-plugin`, the checked text domain actually is the full path: `wp plugin check path/to/my-plugin`

**Context**

After the explanation in #773 that a local zip file is not supported yet, I now tried to check my plugin via path: 

1. I am creating a scoped package of my plugin using [php-scoper](https://github.com/humbug/php-scoper). The scoped package is created in a subfolder of my plugin with the same name as my plugin. Seen from the plugins folder it looks like this: `./my-plugin/my-plugin`
3. I make sure plugin-checker is available in wp-env
4. I check the subfolder in wp-env
```
{
    "composer scoped:create",
    "npx wp-env run cli wp plugin install plugin-check --activate",
    "npx wp-env run cli wp plugin check wp-content/plugins/my-plugin/my-plugin"
}
```

The plugin check works this time, except for one issue: The expected text domain contains the full path `wp-content/plugins/my-plugin/my-plugin` instead of simply `my-plugin`, which produces a lot of `WordPress.WP.I18n.TextDomainMismatch` errors.

davidperezgar:
Fixes #774


ernilambar:
@davidperezgar https://github.com/WordPress/plugin-check/issues/774 This issue is about incorrect plugin slug calculation when checked like `/wp-content/plugins/my-plugin/my-plugin`. I am not getting how this solves the mentioned issue. Am I missing something?

ernilambar:
Removing from `1.3` milestone. We can push this for `1.4` after detailed discussion.

davidperezgar:
My proposal is to calculate the default slug before launching the checks. If slug is empty, then it should calc the slug.

ernilambar (approved):

frantorres (approved):

PR #786, opened 2024-11-12
Add new checks for Plugin Header fields

### LINKED ISSUE #783: Add new checks in `Plugin_Header_Fields_Check`
While going through the Plugin Uploader in Meta, I found that following issues are considered blockers:

- Empty `Description`
- Empty `Version`
- Invalid `Version` - (Regex: `|[^\d\.]|` )
- `Plugin URI` and `Author URI` are same

It would make sense to incorporate these in PCP also so that plugin author will be aware of the issues before uploading zip for the directory.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/783

New error codes added:

- plugin_header_missing_plugin_description
- plugin_header_missing_plugin_version
- plugin_header_invalid_plugin_version


includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -174,6 +223,29 @@ public function run( Check_Result $result ) {
...
+						__( 'The "%1$s" and "%2$s" header in the plugin file must be different. It is not required to provide both, so pick the one that best applies to your situation.', 'plugin-check' ),
+						esc_html( $labels['PluginURI'] ),
+						esc_html( $labels['AuthorURI'] )
+					),
+					'plugin_header_same_plugin_author_uri',
+					$plugin_main_file,
+					0,
+					0,
+					__( 'https://developer.wordpress.org/plugins/plugin-basics/header-requirements/', 'plugin-check' ),
+					7
```
davidperezgar: I wouldn't make 7 either.

davidperezgar: @ernilambar could you comment this?

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -155,6 +170,40 @@ public function run( Check_Result $result ) {
...
+				__( 'https://developer.wordpress.org/plugins/plugin-basics/header-requirements/', 'plugin-check' ),
+				7
+			);
+		} else {
+			if ( preg_match( '|[^\d\.]|', $plugin_header['Version'] ) ) {
+				$this->add_result_error_for_file(
+					$result,
+					sprintf(
+						/* translators: %s: plugin header field */
+						__( 'The "%s" header in the plugin file should only contain numeric and period characters.', 'plugin-check' ),
```
swissspidy: Why? 🤔 That sounds way too strict. Many plugins use versions such as 14.0.0-RC1 or so, and semver is not that restrictive.

ernilambar: Yah, I have also thought about that. Even Akismet has beta is some versions. Did I misread the code? I took reference from here https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-upload-handler.php#L294

ernilambar: Ah, it makes sense to block this in uploader as initial submission is not expected to have alphabets in the version.

ernilambar: Lets make regex less strict by this: `/^[0-9a-z.-]+$/i`

chriscct7 (approved):

davidperezgar (approved):

PR #789, opened 2024-11-14
Use latest readme parser

ernilambar:
This working fine in local.

Tests are failing in GitHub actions because Docker in which wp-env run does not have `patch` command which is required by `cweagans/composer-patches`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ local.properties
 .cproject
 .buildpath
 nbproject/
+patches.lock.json
 
 ############
 ## Build
--- a/composer.json
+++ b/composer.json
@@ -6,12 +6,13 @@
   "require": {
     "php": ">=7.2.24",
     "ext-json": "*",
-    "afragen/wordpress-plugin-readme-parser": "dev-master#67fba498d0b112acf84386b95e4905c539a33f0b",
+    "afragen/wordpress-plugin-readme-parser": "dev-master#c3758599348148be684b3c4ad1105d09b6230d51",
     "automattic/vipwpcs": "^3.0.0",
     "composer/installers": "^2.2",
     "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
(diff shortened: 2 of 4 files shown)


afragen:
Or just don't specify a particular commit. The `master` branch of `afragen/wordpress-plugin-readme-parser` will always have the most recent version. It's set to auto-update when there's an update on .org and GA creates an automatic PR to my repository occurs. All I do is merge.

ernilambar:
@afragen Yes, I understand that. But I dont think removing version from there solves the actual issue at hand. Parser class is not PHP 7.2 compatible but PCP is. Patching I am trying to do here is that will make parser class compatible to PHP 7.2.

afragen:
OK, you could try asking Dion to make the changes necessary.

ernilambar:
I have already made the ticket https://meta.trac.wordpress.org/ticket/7749 but looks like it wont be updated in the meta side.

afragen:
Is there something else that won't run in PHP7.4?

ernilambar:
> Is there something else that won't run in PHP7.4?

`$this->warnings['contributor_ignored'] ??= [];` Only this one line is incompatible with PHP 7.2.

afragen:
My question was why not raise the requirement to 7.4?

ernilambar:
> My question was why not raise the requirement to 7.4?

I have not seen the discussion about minimum PHP version support in PCP particularly. May be trying to follow approach which core uses (5% threshold).

Current usage (from [Stats](https://wordpress.org/about/stats/)):

- PHP 7.2: 2.78%
- PHP 7.3: 3.31%

Combined would be greater than 5% if bumped to 7.4.
Saying that, I would be personally happy to get version bumped to 7.4 or even 8.0, since PCP is targeted mainly for devs. Will need discussion among senior maintainers about the issue.

ernilambar:
Closing this as we wont be use patching.

PR #800, opened 2024-11-19
Disallow usage of HEREDOC and NOWDOC

### LINKED ISSUE #792: Do not use HEREDOC or NOWDOC syntax in your plugins
### Do not use HEREDOC or NOWDOC syntax in your plugins

> While both are totally valid, and in many ways desirable features of PHP that allow you to output content, it comes with a cost that is too high for most plugins.

> The primary issue is that most (if not all) codesniffers won't detect lack of escaping in code when you use HEREDOC or NOWDOC. While there are ways around this they have the end result of dashing all that readability to the rubbish pile and leaving you with a jumbled mess that won't properly be scanned.

> We feel the risk here is much higher than the benefits, which is why we don't permit their use.

We could use this rule - https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Sniffs/PHP/HeredocSniff.php

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/792

### DIFF
--- a/phpcs-rulesets/plugin-review.xml
+++ b/phpcs-rulesets/plugin-review.xml
@@ -56,6 +56,11 @@
 		<severity>7</severity>
 	</rule>
 
+	<!-- Prohibit the use of HEREDOC or NOWDOC. -->
+	<rule ref="Squiz.PHP.Heredoc">
+		<severity>7</severity>
+	</rule>
+
 	<!-- Prohibit the use of the `goto` PHP language construct. -->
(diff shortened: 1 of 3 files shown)


davidperezgar (approved):

davidperezgar:
I'd suggest to update docs/checks.md

ernilambar:
> I'd suggest to update docs/checks.md

No new check has been added. Rule is added under plugin_review_phpcs check.

frantorres (approved):

PR #810, opened 2024-11-26
Make Requires Plugins header check stricter

### LINKED ISSUE #806: Make error code `plugin_header_invalid_requires_plugins` an ERROR with severity 7
In the directory when `Requires Plugins` value is invalid, even new release of plugin is blocked. Its value should be comma separated plugin slug and those plugins should exist in the directory.

May be it is not feasible to check existence of the plugin in the PCP, we could at least raise an ERROR if the value is not proper comma separated slugs.

Error:

```
Requires Plugins: WooCommerce
```

Expected:
```
Requires Plugins: woocommerce
```

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/806

### DIFF
--- a/tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
+++ b/tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
@@ -4,7 +4,7 @@
  * Plugin URI: https://github.com/WordPress/plugin-check
  * Description: Test plugin for the Localhost check.
  * Requires at least: 6.0
- * Requires PHP: 5.6
+ * Requires PHP: latest
  * Version: 1.0.0
  * Author: WordPress Performance Team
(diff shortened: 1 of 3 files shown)


tests/phpunit/testdata/plugins/test-plugin-unfiltered-uploads-with-errors/load.php
```diff
@@ -4,7 +4,7 @@
  * Plugin URI: https://github.com/WordPress/plugin-check
  * Description: Test plugin for the Localhost check.
  * Requires at least: 6.0
- * Requires PHP: 5.6
+ * Requires PHP: latest
```
ernilambar, line 7: We have one test where we check for lack of error `plugin_header_invalid_requires_plugins`. This change has been done so that there is errors in `$errors` variable in the test and our targeted code is not there.

davidperezgar (approved):

swissspidy (approved):

PR #813, opened 2024-12-02
Fix PHPCS checks on unwritable filesystems.

dd32:
In #744 a call to setConfigData() was added which has the `$temp` parameter set to the default, which is `true`.

This causes PHPCS to attempt to write the given config to disk.
The default config location is.. `wp-content/plugins/plugin-check/vendor/squizlabs/php_codesniffer/CodeSniffer.conf`

If WordPress is installed on an immutable filesystem (Such as WordPress.org) this causes the entire plugin check to abort.


dd32:
cc @swissspidy 

I've not exactly tested this, I have tested that it allows an unwritable config file not to abort, but I haven't actually tested that it doesn't break the functionality.

This has been deployed to WordPress.org.

swissspidy:
@ernilambar Does your custom addon still work with that?

I thought the config needs to be written so that an addon with custom PHPCS checks works as expected.

ernilambar:
Yes. Addon is working like before with third parameter set to true.

swissspidy (approved):

ernilambar (approved):

felixarntz (approved):

PR #817, opened 2024-12-03
Add version utilities

ernilambar:
Add version utilities:

- Add `get_wordpress_stable_version()` similar to that helper method in `Plugin_Readme_Check` class
- New methods to fetch version informations
   * get_wordpress_latest_version()
- Added `Version_Utils_Tests` unit test class

### DIFF
--- a/includes/Traits/Version_Utils.php
+++ b/includes/Traits/Version_Utils.php
@@ -80,7 +80,7 @@ trait Version_Utils {
 	 *
 	 * @return string Recommended PHP version.
 	 */
-	protected function get_recommended_php_version(): string {
+	protected function get_wordpress_recommended_php_version(): string {
 		$recommended_version = '7.4'; // Default fallback recommended version.
 
 		$version_details = wp_check_php_version();
--- a/includes/Traits/Version_Utils.php
+++ b/includes/Traits/Version_Utils.php
@@ -48,51 +48,6 @@ trait Version_Utils {
 	}
 
 	/**
-	 * Returns required PHP version.
-	 *
-	 * @since 1.4.0
-	 *
-	 * @return string Required PHP version.
-	 */
-	protected function get_php_required_version(): string {
-		$version = $this->get_latest_version_info( 'php_version' );
-
-		return $version ?? $this->get_required_php_version();
-	}
-
-	/**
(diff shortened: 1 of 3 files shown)


ernilambar:
On second thought, this could be alternative naming convention:

- get_wordpress_stable_version()
- get_wordpress_latest_version()
- get_php_required_version()
- get_php_recommended_version()
- get_mysql_required_version()

includes/Traits/Version_Utils.php
```diff
@@ -0,0 +1,155 @@
...
+ * Trait for version utilities.
+ *
+ * @since 1.4.0
+ */
+trait Version_Utils {
+
+	/**
+	 * Returns current major WordPress version.
+	 *
+	 * @since 1.0.0
```
mukeshpanchal27:
```suggestion
	 * @since 1.4.0
```

ernilambar:
PR updated with new naming pattern.

Also added `get_wordpress_relative_major_version()` method to find relative WP major version.

includes/Traits/Version_Utils.php
```diff
@@ -0,0 +1,174 @@
...
+					set_transient( 'wp_plugin_check_latest_version_info', $info, DAY_IN_SECONDS );
+				}
+			}
+		}
+
+		return array_key_exists( $key, $info ) ? $info[ $key ] : null;
+	}
+
+	/**
+	 * Returns the current WordPress' required PHP version.
```
swissspidy: Why do we need that? Same for the current site's required MySQL version

ernilambar: In our internal addon we are trying to use this like checking this value vs PHP version mentioned in the plugin header. We can remove get_mysql_required_version() as I dont see its implementation soon in the near future.

ernilambar: We could also use PHP version values in message `The "Requires PHP" header in the plugin file should only contain a PHP version such as "5.2.4" or "7.0".` where currently those versions are hardcoded.

swissspidy: Those being hardcoded seems fine to me

swissspidy (approved):

davidperezgar (approved):

PR #818, opened 2024-12-05
Preferred error type to be singular

ernilambar:
Followup https://github.com/WordPress/plugin-check/pull/696

Renamed error type to `ERROR_LOW_SEVERITY` and `WARNING_LOW_SEVERITY`.


dd32 (approved):
For consistency with `ERROR` and `WARNING`, this makes sense to me.

Ref: https://github.com/WordPress/plugin-check/pull/696#issuecomment-2518991667

(Note: the WordPress.org plugin-check implementation will handle this change fine)

swissspidy (approved):

PR #819, opened 2024-12-06
Add check for restricted contributors

ernilambar:
Add check for restricted contributors.

These are few frequent encounters where readme is generated from some third party readme generator and plugin author forget to update the Contributors value in readme:

```
$restricted_contributors = array(
	'username',
	'your-username',
	'your_wordpress_username',
	'yourusername',
);

```

Edit: Add sample output

Screen Shot 2024-12-09 at 1 16 29 PM

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
@@ -840,9 +840,12 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 	private function get_restricted_contributors() {
 		$restricted_contributors = array(
 			'username'                => 'error',
+			'your-name'               => 'error',
 			'your-username'           => 'error',
+			'your-wordpress-username' => 'error',
 			'your_wordpress_username' => 'error',
 			'yourusername'            => 'error',
+			'yourwordpressusername'   => 'error',
 			'wordpressdotorg'         => 'warning',
 		);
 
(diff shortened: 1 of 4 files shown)


includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -692,6 +692,35 @@ private function check_for_contributors( Check_Result $result, string $readme_fi
 				'',
 				6
 			);
+
+			return;
+		}
+
+		$restricted_contributors = array(
```
swissspidy: Could we add a _warning_ for wordpressorg? Sometimes people mistakenly add this user as a contributor, even though it‘s not a feature/community plugin

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -762,6 +830,34 @@ private function get_ignored_warnings( Parser $parser ) {
...
+			'yourusername'            => true,
+			'wordpressdotorg'         => false,
+		);
+
+		/**
+		 * Filter the list of restricted contributors.
+		 *
+		 * @since 1.4.0
+		 *
+		 * @param array $restricted_contributors Array of restricted contributors with boolean value to indicate whether username should be error or warning. Value 'true' for error and 'false' for warning.
```
swissspidy: What do you think about using 'error' and 'warning' instead of true/false?

swissspidy (approved):

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -762,6 +830,34 @@ private function get_ignored_warnings( Parser $parser ) {
...
+	 *
+	 * @return array Restricted contributors.
+	 */
+	private function get_restricted_contributors() {
+		$restricted_contributors = array(
+			'username'                => 'error',
+			'your-username'           => 'error',
+			'your_wordpress_username' => 'error',
+			'yourusername'            => 'error',
+			'wordpressdotorg'         => 'warning',
```
davidperezgar, line 849: Should we add woocommerce?

davidperezgar (approved):

PR #825, opened 2024-12-10
Add check for unsupported plugin name in plugin header field

### LINKED ISSUE #805: Restrict plugin name with all unsupported characters
In meta, slug generation fails and uploading is blocked if the plugin name has complete unsupported characters. We could also add check for that.

Example:

```
Plugin Name: अद्भुत प्लगइन
```

We could add this check as error with severity 7.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/805

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
@@ -97,6 +97,25 @@ class Plugin_Header_Fields_Check implements Static_Check {
 					'',
 					6
 				);
+			} else {
+				$potential_slug = sanitize_title_with_dashes( str_replace( '_', '-', preg_replace( '/[^a-z0-9 _.-]/i', '', remove_accents( $plugin_header['Name'] ) ) ) );
+
+				if ( empty( $potential_slug ) ) {
+					$this->add_result_error_for_file(
+						$result,
+						sprintf(
+							/* translators: %s: plugin header field */
+							__( 'The "%s" header in the plugin file is not valid. It may only contain latin letters (A-z), numbers, spaces, and hyphens.', 'plugin-check' ),
+							esc_html( $labels['Name'] )
+						),
+						'plugin_header_unsupported_plugin_name',
+						$plugin_main_file,
+						0,
+						0,
+						'https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields',
+						7
+					);
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ernilambar:
Tests are failing due to unrelated issue with this PR.

Edit: Tests are passing now.

davidperezgar (approved):

swissspidy (approved):

ernilambar:
@frantorres  has suggested to modify the conditional and make less strict. Like:

Check if plugin name "has at least 5 latin letters (A-z), numbers, spaces, and hyphens" and not worry if there are other non valid characters around.

davidperezgar:
I think that you meant @frantorres

ernilambar:
I have updated the conditional to make less strict. Current message looks off after new conditional:

> The "Plugin Name" header in the plugin file is not valid. It may only contain latin letters (A-z), numbers, spaces, and hyphens.

Can you please suggest updated message?

frantorres:
Maybe
> The "Plugin Name" header in the plugin file is not valid. It needs to contain at least 5 latin letters (a-Z) and/or numbers. This is necessary because the initial plugin slug is generated from the name.

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -97,6 +97,25 @@ public function run( Check_Result $result ) {
 					'',
 					6
 				);
+			} else {
+				$valid_chars_count = preg_match_all( '/[a-z0-9-]/i', $plugin_header['Name'] );
```
frantorres: The regex might be` /[a-zA-Z0-9]/i`

includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
```diff
@@ -97,6 +97,25 @@ public function run( Check_Result $result ) {
...
+
+				if ( intval( $valid_chars_count ) < 5 ) {
+					$this->add_result_error_for_file(
+						$result,
+						sprintf(
+							/* translators: %s: plugin header field */
+							__( 'The "%s" header in the plugin file is not valid. It may only contain latin letters (A-z), numbers, spaces, and hyphens.', 'plugin-check' ),
```
frantorres: The "%s" header in the plugin file is not valid. It needs to contain at least 5 latin letters (a-Z) and/or numbers. This is necessary because the initial plugin slug is generated from the name.

frantorres (changes requested):

frantorres (approved):

PR #832, opened 2024-12-14
Check Readme Third Parties must have Privacy Notes in the readme

### LINKED ISSUE #475: Check: README Third parties MUST have privacy notes in the readme
In this check, we find third parties and ask the users to add it in the readme. So, we need to:
- Detect external calls
- Search in the readme to see that is added

And we make this reply to our users

> ## Undocumented use of a 3rd Party or external service
>  
> We permit plugins to require the use of 3rd party (i.e. external) services, provided they are properly documented in a clear manner.
> 
> We require plugins that reach out to other services to disclose this, in clear and plain language, so users are aware of where data is being sent. This allows them to ensure that any legal issues with data transmissions are covered. This is true even if you are the 3rd party service.
> 
> In order to do so, you must update your readme to do the following:
> 
> clearly explain that your plugin is relying on a 3rd party as a service and under what circumstances
> provide a link to the service
> provide a link to the services’ a terms of use and/or privacy policies
> 
> Remember, this is for your own legal protection. Use of services must be upfront and well documented.
> 

davidperezgar:
Fixes #475


includes/Traits/External_Utils.php
```diff
@@ -0,0 +1,289 @@
...
+	/**
+	 * Filter the given array of files for php,js,css files.
+	 *
+	 * @since 1.4.0
+	 *
+	 * @param array  $files                Array of file files to be filtered.
+	 * @param string $plugin_relative_path Plugin relative path.
+	 * @return array An array containing php,js.css files, or an empty array if none are found.
+	 */
+	protected function filter_files_for_external( array $files, $plugin_relative_path ) {
```
ernilambar: We already have `filter_files_by_extensions()` method to filter files by extensions.

PR #852, opened 2025-01-09
Check valid WP version for `Requires at least` plugin header

### LINKED ISSUE #851: Check `Requires at least` is a correct value
In the plugin headers:

**Requires at least**: The lowest WordPress version that the plugin will work on.

Some author confuses this with PHP version and keep `Requires at least: 8.0`. This is not correct and seems to be problematic. We should inform user if nonexistent version is kept in the header.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/851

- Check added for `Requires at least` plugin header for validity of WordPress version
- Version check logic is similar to `Tested up to`.
- At `6.7` WP version, error will be generated for more than or equal to WP `6.9`
- Unit tests added for the updated check

### DIFF
--- a/tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
+++ b/tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
@@ -110,4 +110,61 @@ class Plugin_Header_Fields_Check_Tests extends WP_UnitTestCase {
 		$this->assertCount( 1, wp_list_filter( $errors['load.php'][0][0], array( 'code' => 'plugin_header_missing_plugin_description' ) ) );
 		$this->assertCount( 1, wp_list_filter( $errors['load.php'][0][0], array( 'code' => 'plugin_header_invalid_plugin_version' ) ) );
 	}
+
+	public function test_run_with_errors_requires_at_least_latest_plus_two_version() {
+		// Target plugin has "6.0" in plugin header.
+		set_transient( 'wp_plugin_check_latest_version_info', array( 'current' => '5.8.1' ) );
+
+		$readme_check  = new Plugin_Header_Fields_Check();
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-localhost-with-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$readme_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		$this->assertNotEmpty( $errors );
+
+		$filtered_items = wp_list_filter( $errors['load.php'][0][0], array( 'code' => 'plugin_header_nonexistent_requires_wp' ) );
+
+		$this->assertCount( 1, $filtered_items );
+		$this->assertStringContainsString( 'Requires at least: 6.0', $filtered_items[0]['message'] );
+		$this->assertStringContainsString( 'This version of WordPress does not exist (yet).', $filtered_items[0]['message'] );
+
+		delete_transient( 'wp_plugin_check_latest_version_info' );
+	}
+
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
```diff
@@ -110,4 +110,61 @@ public function test_run_with_invalid_header_fields() {
...
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-localhost-with-errors/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$readme_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		$this->assertEmpty( $errors );
+
+		delete_transient( 'wp_plugin_check_latest_version_info' );
```
swissspidy, line 166: Nit: cleanup should happen before doing any assertions. Otherwise, if the assertion fails, this code is never executed.

So either move this a few lines above, or move this to a `tear_down()` method.

Same for the other tests above.

tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
```diff
@@ -110,4 +110,61 @@ public function test_run_with_invalid_header_fields() {
...
+		$filtered_items = wp_list_filter( $errors['load.php'][0][0], array( 'code' => 'plugin_header_nonexistent_requires_wp' ) );
+
+		$this->assertCount( 1, $filtered_items );
+		$this->assertStringContainsString( 'Requires at least: 6.0', $filtered_items[0]['message'] );
+		$this->assertStringContainsString( 'This version of WordPress does not exist (yet).', $filtered_items[0]['message'] );
+
+		delete_transient( 'wp_plugin_check_latest_version_info' );
+	}
+
+	public function test_run_with_errors_requires_at_least_latest_plus_one_version() {
```
swissspidy: Why name it `run_with_errors` when this tests the scenario without any errors? 🤔

tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
```diff
@@ -110,4 +110,61 @@ public function test_run_with_invalid_header_fields() {
...
+		$readme_check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		$this->assertEmpty( $errors );
+
+		delete_transient( 'wp_plugin_check_latest_version_info' );
+	}
+
+	public function test_run_with_errors_requires_at_least_latest_version() {
```
swissspidy: Ditto here about `run_with_errors` in the name

tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
ernilambar: Copy/pasting effect. :-) I will rename method names.

tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
ernilambar: Hmm. Looks like this issue is in other tests also. May be similar should apply for add_filter/remove_filter approach in our tests?

swissspidy: I haven't looked at other tests, but sounds like it :)

swissspidy (approved):

davidperezgar (approved):

PR #854, opened 2025-01-10
Add check for setting sanitization

### LINKED ISSUE #833: Check: Sanitization for register_setting
When the fields are registered through register_setting(), should be sanitized properly.

https://developer.wordpress.org/reference/functions/register_setting/

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/833

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Setting_Sanitization_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Setting_Sanitization_Check.php
@@ -1,0 +1,113 @@
+<?php
+/**
+ * Class Setting_Sanitization_Check.
+ *
+ * @package plugin-check
+ */
+
+namespace WordPress\Plugin_Check\Checker\Checks\Plugin_Repo;
+
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Abstract_PHP_CodeSniffer_Check;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+
+/**
+ * Check to detect sanitization in register_setting().
+ *
+ * @since 1.4.0
+ */
+class Setting_Sanitization_Check extends Abstract_PHP_CodeSniffer_Check {
+
+	use Amend_Check_Result;
+	use Stable_Check;
+
+	/**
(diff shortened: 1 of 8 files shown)


frantorres (approved):

davidperezgar (approved):
I see severity 7. We'll have to care about it in logs.

nightwalker89:
The SettingSanitization sniff is incorrectly flagging valid WordPress code patterns. Here's a minimal example demonstrating the issue:

```php
<?php
/**
 * Example demonstrating valid WordPress register_setting() usage
 * that is incorrectly flagged by PluginCheck.CodeAnalysis.SettingSanitization
 */
class Example_Plugin {
    public function register_settings() {
        // This is valid WordPress code but gets flagged:
        register_setting(
            'example_options',
            'example_options',
            array(
                'type' => 'array',
                'sanitize_callback' => array($this, 'sanitize_options'),
                'default' => array(
                    'setting_1' => true,
                    'setting_2' => 'default'
                )
            )
        );
    }

    public function sanitize_options($input) {
        $sanitized = array();
        $sanitized['setting_1'] = !empty($input['setting_1']);
        $sanitized['setting_2'] = sanitize_text_field($input['setting_2']);
        return $sanitized;
    }
}
```

### Issues with Current Sniffer:

1. The sniffer flags `array($this, 'method_name')` as a dynamic argument, but this is standard WordPress practice
2. WordPress core itself uses this exact pattern extensively
3. Static arrays with 'sanitize_callback' are valid and secure
4. The current implementation forces less maintainable code patterns

### Example from WordPress Core

Here's how WordPress core itself uses this pattern (from `wp-admin/includes/options.php`):

```php
register_setting('general', 'blogname', array(
    'sanitize_callback' => array($this, 'sanitize_option_blogname'),
    'show_in_rest'     => true,
    'type'             => 'string',
));
```

### Suggested Improvements

The sniffer should be updated to:
1. Recognize static arrays with 'sanitize_callback' as valid
2. Allow class method callbacks using `array($this, 'method_name')`
3. Consider WordPress core's own usage patterns as valid
4. Focus on actual dynamic/unsafe arguments rather than standard WordPress patterns

This would help prevent false positives while still catching genuinely unsafe code.

digitaldonkey:
Anyone from Plugin Team can comment on the @nightwalker89 issue or do we need to open a bug ticket for it? @frantorres ?

davidperezgar:
Hello @digitaldonkey and @nightwalker89 . The best way is to create a new issue, where we can contribute in a new PR to solve the issue.

PR #856, opened 2025-01-15
Introduce `--ignore-codes` in CLI

### LINKED ISSUE #769: Introduce mechanism to exclude certain error codes only
Currently we have following argument for excluding check:

- **--exclude-checks** - Excluding certain checks.

We could introduce a way for excluding certain error codes from a check. There could be several error codes in a check. Currently there is no way we could ignore some specific error codes only.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/769

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -16,6 +16,9 @@ Runs plugin check.
 : Exclude checks provided as an argument in comma-separated values, e.g. i18n_usage, late_escaping.
 Applies after evaluating `--checks`.
 
+[--ignore-codes=<checks>]
+: Ignore error codes provided as an argument in comma-separated values.
+
 [--format=<format>]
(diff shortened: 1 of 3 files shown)


swissspidy:
`--exclude-checks` makes sense because it excludes some checks from running completely.

Here, I would rather label this `--ignore-codes`, `--ignore-messages` or simply `--ignore`

swissspidy (approved):
Logic OK, but still not 100% happy with the name. Curious what others think.

davidperezgar:
I'm not sure if it's an standard, but I'd vote for ignore-codes. I think it's fine.

davidperezgar (approved):

felixarntz (approved):
LGTM!

PR #865, opened 2025-01-28
Fix incorrect database tables being referenced on subsites in Multisite

### LINKED ISSUE #859: Missing database tables on Multisite
I have Plugin Check installed on a multisite installation and activated it on site 19. I see many errors in debug log about non existing tables in the database.

> WordPress database error Table 'database.wp_19_pc_blogs' doesn't exist for query 
WordPress database error Table 'database.wp_19_pc_sitemeta' doesn't exist for query 
WordPress database error Table 'database.wp_19_pc_usermeta' doesn't exist for query 
WordPress database error Table 'database.wp_19_pc_users' doesn't exist for query 

Apparently these tables are not created.

felixarntz:
Fixes #859

* A new `Amend_DB_Base_Prefix` trait is introduced to abstract away this complication in a single place.
* That trait is now used anywhere the database prefix is being altered.


felixarntz:
@swissspidy Any advice here on how we can avoid the PHPStan failures in https://github.com/WordPress/plugin-check/actions/runs/13018635484/job/36313903361?pr=865? It doesn't seem to recognize `wpdb`. But what's particularly strange to me is: Why does it fail now but didn't fail before this PR?

swissspidy:
> @swissspidy[](https://profiles.wordpress.org/swissspidy) Any advice here on how we can avoid the PHPStan failures in [WordPress/plugin-check/actions/runs/13018635484/job/36313903361?pr=865](https://github.com/WordPress/plugin-check/actions/runs/13018635484/job/36313903361?pr=865)? It doesn't seem to recognize `wpdb`. But what's particularly strange to me is: Why does it fail now but didn't fail before this PR?

Good question. Maybe because this is in a trait and not a class. I'd need to check this in more detail. In the meantime I pushed a small fix.

felixarntz:
@swissspidy Thanks! Seeing the fix, I'm thinking it would be cool if PHPStan supported this via `@global` annotations alone. Those already include the type, so I wonder why they are not being relied on when present.

swissspidy:
I think the WordPress stubs/extension actually recognizes any global `$wpdb` variable, without any need for PHPDoc. But again, there might be something different about traits and this would need to be further investigated.

davidperezgar (approved):

swissspidy (approved):

PR #867, opened 2025-01-30
Improve license check for Apache

ernilambar:
License:

License "Apache 2.0" is allowed as it is GPL compatible. This PR improves the license check and accepts few variations of the license text which as found in the directory.

Eg:
- Apache License, Version 2.0
- Apache License 2.0
- Apache 2.0
- Apache-2.0
- Apache 2
- Apache License 2

When normalizing, 2 is also kept as if somehow we encounter Apache 1.0, it should be blocked. So, I have kept normalized license like "Apache2".


davidperezgar (approved):

davidperezgar:
Be aware that you have it in Draft.

swissspidy (approved):

PR #874, opened 2025-02-17
Add include experimental option in admin

### LINKED ISSUE #873: Add Include Experimental option in admin UI
Proposal:

- Add checkbox or toggle to enable/disable `Include Experimental` feature for the check

Note: This feature has been already implemented in the CLI.

Why?

This feature would provide some key benefits:

- It allows us to gather crucial feedback from plugin authors before making the check mandatory, ensuring a smoother transition and a more robust final implementation.
- By providing an opt-in period, we avoid potential disruption and negative feedback that might arise from suddenly enforcing the check. 
- It allows for thorough testing and incremental improvements to the check based on real-world usage and feedback from participating plugin authors.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/873

Screenshot:
Screenshot 2025-02-17 at 4 19 08 PM

### DIFF
--- a/includes/Admin/Admin_AJAX.php
+++ b/includes/Admin/Admin_AJAX.php
@@ -112,7 +112,11 @@ final class Admin_AJAX {
 		$checks = is_null( $checks ) ? array() : $checks;
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
+		$include_experimental = filter_input( INPUT_POST, 'include-experimental', FILTER_VALIDATE_INT );
+		$include_experimental = ( 1 === absint( $include_experimental ) );
+
 		try {
+			$runner->set_experimental_flag( $include_experimental );
 			$runner->set_check_slugs( $checks );
 			$runner->set_plugin( $plugin );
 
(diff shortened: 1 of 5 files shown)


includes/Admin/Admin_Page.php
```diff
@@ -284,6 +284,13 @@ public function render_page() {
 		$user_enabled_categories = get_user_setting( 'plugin_check_category_preferences', implode( '__', $this->get_default_check_categories_to_be_selected() ) );
 		$user_enabled_categories = explode( '__', $user_enabled_categories );
 
+		$check_repo = new Default_Check_Repository();
+
+		$non_experimental_collection = $check_repo->get_checks( Check_Repository::TYPE_ALL );
+		$complete_collection         = $check_repo->get_checks( Check_Repository::TYPE_ALL + Check_Repository::INCLUDE_EXPERIMENTAL );
+
+		$has_experimental_checks = ( count( $non_experimental_collection ) !== count( $complete_collection ) );
```
swissspidy: Could this be simplified?

```php
$experimental_checks = $check_repo->get_checks( Check_Repository::INCLUDE_EXPERIMENTAL );

$has_experimental_checks = count( $experimental_checks ) > 0;

File: includes/Admin/Admin_AJAX.php [hunk 2]
```diff
@@ -112,7 +112,11 @@ public function set_up_environment() {
 		$checks = is_null( $checks ) ? array() : $checks;
 		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
+		$include_experimental = filter_input( INPUT_POST, 'include-experimental', FILTER_VALIDATE_INT );
+		$include_experimental = ( 1 === absint( $include_experimental ) );
```
mukeshpanchal27:
```suggestion
		$include_experimental = 1 === filter_input( INPUT_POST, 'include-experimental', FILTER_VALIDATE_INT );
```

Simplified?

includes/Admin/Admin_Page.php
ernilambar: `$experimental_checks = $check_repo->get_checks( Check_Repository::INCLUDE_EXPERIMENTAL );` <- This is not giving experimental checks.

ernilambar:
Should we save the state of new checkbox like we are doing for categories?

includes/Admin/Admin_Page.php
swissspidy: Ok, strange 🤔

swissspidy:
@ernilambar still a draft?

templates/admin-page.php
```diff
@@ -60,6 +60,14 @@
 				<?php } ?>
 				</table>
 				<?php } ?>
+
+				<?php if ( $has_experimental_checks ) { ?>
+					<h4><?php esc_attr_e( 'Other Options', 'plugin-check' ); ?></h4>
+					<p>
+						<input type="checkbox" value="include-experimental" id="plugin-check__include-experimental" /> <?php esc_html_e( 'Include Experimental Checks', 'plugin-check' ); ?>
```
swissspidy: This should be wrapped in a `<label>` so that you can click on the text. Like you can with the other checkboxes.

Note: the `for` attribute is not needed if the label wraps the input.

includes/Admin/Admin_Page.php
swissspidy: I think I would do this instead:

```php
$collection = $check_repo->get_checks( Check_Repository::TYPE_ALL | Check_Repository::INCLUDE_EXPERIMENTAL )->filter(
	static function ( Check $check ) {
		return $check->get_stability() === Check::STABILITY_EXPERIMENTAL;
	}
);

$has_experimental_checks = count( $collection ) > 0;
```

swissspidy (commented):
Looks ok, just a couple of suggestions.

Not really an urgent change as we don't actually have any experimental checks right now.

swissspidy (approved):

frantorres (approved):

PR #878, opened 2025-02-20
Make sure headers are not empty in the requires header check

ernilambar:
When there is `Requires at least` is plugin header but not in readme, this check should not have been triggered. It generated message like this.

```
Mismatched Requires at least:  != 4.9. "Requires at least" needs to be exactly the same with that in your main plugin file's header.
```

So I believe it makes sense to check for not empty for header value before actually checking the comparison.


davidperezgar (approved):

swissspidy:
Is there a test or at least bug description that explains whag this solves?

ernilambar:
Description updated.

swissspidy (approved):

PR #879, opened 2025-02-25
Prefix dependencies with PHP-Scoper

### LINKED ISSUE #328: Prefix Composer dependencies
PHPCS is installed via Composer and we use the Composer autoloader for loading all these classes.

To avoid conflicts with any other plugins bundling PHPCS, we should prefix the dependencies' classes with something like PHP-Scoper. See https://github.com/googleforcreators/web-stories-wp/ and https://github.com/google/site-kit-wp for some prior art where this is done.

swissspidy:
All the production dependencies required by Plugin Check are prefixed and stored in `vendor-prefixed`. This includes:

* `afragen/wordpress-plugin-readme-parser`
* `plugin-check/phpcs-sniffs`
* `squizlabs/php_codesniffer`

This avoids any autoloading conflicts in case another plugin uses the same dependencies.

The plugin's own classes are autoloaded with a new slim autoloader in `includes/vendor`. No actual dependencies are in there, it's just a classmap.

For `afragen/wordpress-plugin-readme-parser` there is a special case: if the plugin is running on WordPress.org, the original and unprefixed `WordPressdotorg\Plugin_Directory\Readme\Parser` is available, so there is a `class_exists` check to use that one if available.

Unfortunately right now requires dependencies to be prefixed manually by installing PHP-Scoper in that other directory & running `composer prefix-dependencies`

Fixes #328

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -9,6 +9,7 @@ patches
 # Distributed via vendor/plugin-check/phpcs-sniffs instead
 /phpcs-sniffs/
 tests
+/vendor
 
 *.DS_Store
 .DS_Store
@@ -40,3 +41,4 @@ phpunit.xml.dist
 plugin-check.iml
 README.md
 SECURITY.md
+scoper.inc.php
--- a/.github/workflows/behat-test.yml
+++ b/.github/workflows/behat-test.yml
@@ -81,18 +81,36 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          tools: composer
+          php-version: ${{ matrix.php }}
+
+      - name: Install PHP dependencies
+        uses: ramsey/composer-install@57532f8be5bda426838819c5ee9afb8af389d51a
+        with:
+          composer-options: '--prefer-dist'
+
+      - name: Make Composer packages available globally
+        run: |
+          echo "${PWD}/vendor/bin" >> $GITHUB_PATH
+
       - name: Setup PHP for PHP-Scoper
         uses: shivammathur/setup-php@v2
         with:
           tools: composer
-          php-version: 8.4
-          coverage: none
+          php-version: latest
 
       - name: Install PHP-Scoper
         uses: ramsey/composer-install@57532f8be5bda426838819c5ee9afb8af389d51a
         with:
-          working-directory: "vendor-bin/php-scoper"
+          working-directory: 'vendor-bin/php-scoper'
           composer-options: '--prefer-dist'
+
+      - name: Prefix Composer dependencies
+        run: |
+          composer prefix-dependencies
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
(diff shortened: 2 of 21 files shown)


swissspidy:
@ernilambar FWIW the issue is that the latest PHP-Scoper requires PHP 8.2. This could be mitigated by installing that dependency with PHP 8.2 separate from the others, run the prefixing, then downgrade again.

swissspidy:
@ernilambar I think it's very close now, after having fixed most of the namespace/autoloader issues. The last one seems to be specifically `Fatal error: Cannot declare class WordPress\Plugin_Check\Vendor\WordPressVIPMinimum\Sniffs\Sniff, because the name is already in use in /var/www/html/wp-content/plugins/plugin-check/vendor-prefixed/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Sniff.php` in unit tests. Probably the same thing that also causes some Behat failures.

Don't yet know why because it doesn't really happen to me locally.

BrianHenryIE:
I haven't got this working yet, but...

I maintain [brianhenryie/strauss](https://github.com/BrianHenryIE/strauss/), a fork of [coenjacobs/mozart](http://github.com/coenjacobs/mozart), and it runs on PHP 7.4. I tried it on this branch and after [a small bugfix](https://github.com/BrianHenryIE/strauss/pull/325) the prefixing is working correctly

That bug was that when Strauss prefixed `vendor-prefixed/composer/*.php` it didn't update classes that were referencing those.

Config in `composer`.`extra` is:
```json
    "strauss": {
      "namespace_prefix": "WordPress\\Plugin_Check\\Vendor\\",
      "include_root_autoload": true,
      "override_autoload": {
        "squizlabs/php_codesniffer": {
          "classmap": [
            "."
          ]
        }
      },
      "exclude_from_copy": {
        "file_patterns": [
          "/\\/\\..*/"
        ]
      }
    }
```
and the Composer script becomes:
```json
   "prefix-namespaces": [
      "sh -c 'test -f ./bin/strauss.phar || mkdir bin && curl -o bin/strauss.phar -L -C - https://github.com/BrianHenryIE/strauss/releases/download/0.29.1/strauss.phar'",
      "@php bin/strauss.phar --info"
    ]
```

* `include_root_autoload` respects the `composer`.`autoload` key in the project, without it it would just generate autoload files for the prefixed packages (this is a change/simplification from this branch)
* `override_autoload` config is because `squizlabs/php_codesniffer` doesn't have a Composer `autoload` key. Strauss copies all files but only prefixes autoloaded ones. The intention there is that packages with their own namespace collision avoidance strategy in their custom autoloader, e.g. Action Scheduler, are not prefixed.
* `exclude_from_copy` in this case is excluding hidden files because there was an error shown. Generally files that shouldn't be included are dealt with in the upstream `.gitattributes` or in `.distignore`

I rebased this branch on `trunk` which should be the same as this (probably some mistakes): https://github.com/WordPress/plugin-check/compare/trunk...BrianHenryIE:0409de7fc1d79297f962bde683a2c86d09dda10b

Unneeded:
* `scoper.inc.php`
* `bamarni/composer-bin-plugin`
* `vendor-bin/php-scoper/composer.json`
* `vendor-bin/php-scoper/composer.lock`

Otherwise it's mostly the same.

Unfortunately it's not working yet. It appears ok in the web browser but when I run the unit tests I get a lot of:

> Unable to find PHPCS Runner class.

https://github.com/WordPress/plugin-check/blob/93b777d79f4dbb7f6ffef25cbf8955c705217de5/includes/Checker/Checks/Abstract_PHP_CodeSniffer_Check.php#L70-L80

Although I have the path updated ( to `WP_PLUGIN_CHECK_PLUGIN_DIR_PATH . 'vendor-prefixed/squizlabs/php_codesniffer/autoload.php`) and all the `PHP_CodeSniffer` namespace references in that file seem to be correctly updated, and `Runner`'s namespace is correctly updated too.

I don't have time to continue right now.

Here are the changes: https://github.com/WordPress/plugin-check/compare/0409de7fc1d79297f962bde683a2c86d09dda10b...BrianHenryIE:plugin-check:fix/328-php-scoper

I'll get a 0.29.1 [release of Strauss](https://github.com/BrianHenryIE/strauss/releases) when CI finishes: https://github.com/BrianHenryIE/strauss/pull/325

When I have the chance I'll get step debugging on those unit tests and hopefully it will be straightforward.

PR #886, opened 2025-03-05
Check minor version tested up

### LINKED ISSUE #885: Check: Don't allow to use minor versions in Tested Up
In our reviews we do not allow to use minor versions in Tested Up header. We ask the autor to change te version like this:

ERROR: Tested up to: "6.7.1" needs to be just the major version number. Example: Tested up: 6.7
ℹ️ Please, include only the major WordPress version, as the minor version is ignored.

So we will have to make the check to accomplish this guideline.

davidperezgar:
Fixes #885

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
@@ -214,12 +214,18 @@ class Plugin_Readme_Check extends Abstract_File_Check {
 				if ( ! empty( $parser->{$field_key} ) && 'tested' === $field_key ) {
 					list( $tested_upto, ) = explode( '-', $parser->{$field_key} );
 
+					$tested_upto_major = $tested_upto;
+					if ( preg_match( '#^\d.\d#', $tested_upto, $matches ) ) {
+						$tested_upto_major = $matches[0];
+					}
+
 					if ( preg_match( '/^\d+\.\d+\.\d+/', $tested_upto ) ) {
 						$this->add_result_error_for_file(
 							$result,
 							sprintf(
 								/* translators: %s: currently used version */
-								__( '<strong>Tested up to: %s</strong><br>The version number should only include major versions (e.g. 6.7), not minor versions (e.g. 6.7.1).', 'plugin-check' ),
+								__( '<strong>Tested up to: </strong><br>The version number should only include major versions %1$s, not minor versions %2$s.', 'plugin-check' ),
+								$tested_upto_major,
 								$tested_upto
 							),
 							'invalid_tested_upto_minor',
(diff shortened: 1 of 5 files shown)


includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -214,6 +214,23 @@ private function check_headers( Check_Result $result, string $readme_file, Parse
 				if ( ! empty( $parser->{$field_key} ) && 'tested' === $field_key ) {
 					list( $tested_upto, ) = explode( '-', $parser->{$field_key} );
 
+					if ( preg_match( '/^\d+\.\d+\.\d+/', $tested_upto ) ) {
+						$this->add_result_error_for_file(
+							$result,
+							sprintf(
+								/* translators: %s: currently used version */
+								__( '<strong>Tested up to: %s</strong><br>The version number should only include major versions (e.g. 6.7), not minor versions (e.g. 6.7.1).', 'plugin-check' ),
```
ernilambar: I would like version here be dynamic rather than hardcoded 6.7.

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
```diff
@@ -491,8 +509,8 @@ public function test_run_with_errors_tested_up_to_latest_plus_two_version() {
 		$filtered_items = wp_list_filter( $errors['readme.md'][0][0], array( 'code' => 'nonexistent_tested_upto_header' ) );
 
 		$this->assertCount( 1, $filtered_items );
-		$this->assertStringContainsString( 'Tested up to: 6.1', $filtered_items[0]['message'] );
-		$this->assertStringContainsString( 'This version of WordPress does not exist (yet).', $filtered_items[0]['message'] );
+		$this->assertStringContainsString( 'Tested up to: 6.1', $filtered_items[1]['message'] );
```
ernilambar: Not sure why this test was affected by the PR.

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
davidperezgar: That's true. I'll change it. thanks

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
davidperezgar: Yeap. I think that is because of the array of elements.

davidperezgar:
@ernilambar I'm having this problem:
```
/home/runner/work/plugin-check/plugin-check/includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php:195  ExcessiveMethodLength  The method check_headers() has 100 lines of code. Current threshold is set to 100. Avoid really long methods.
Script phpmd . text phpmd.xml handling the phpmd event returned with error code 2
```
How could avoid that?

davidperezgar:
I've solved. I saw where to avoid that directive.

tests/phpunit/tests/Checker/Checks/Plugin_Readme_Check_Tests.php
ernilambar: `$filtered_items = array_values( wp_list_filter( $errors['readme.md'][0][0], array( 'code' => 'nonexistent_tested_upto_header' ) ) );` If we use `array_values()` here we could make test more robust and it will whatever position this exists in the $errors array.

davidperezgar: yes, you're right. thanks

ernilambar:
Screenshot 2025-03-14 at 9 56 04 AM
I think message could be little improved. Like:

```
Tested up to: 6.6.1. The version number should only include major versions 6.6.
```

What do you think?

ernilambar (approved):

swissspidy (approved):

PR #905, opened 2025-04-01
Update i18n error codes severity

ernilambar:
PRT team has decided to increase severity for several i18n error codes which are frequently hit in plugin submissions.

This PR:

- Raises the severity for i18n error codes for `WordPress.WP.I18n.InterpolatedVariableXXX` and `WordPress.WP.I18n.NonSingularStringLiteralXXX`. Earlier only selective errors codes were assigned severity 7.
- Unit tests have been added.

### DIFF
--- a/includes/Checker/Checks/General/I18n_Usage_Check.php
+++ b/includes/Checker/Checks/General/I18n_Usage_Check.php
@@ -97,6 +97,7 @@ class I18n_Usage_Check extends Abstract_PHP_CodeSniffer_Check {
 	 * @param string       $docs     URL for further information about the message.
 	 * @param int          $severity Severity level. Default is 5.
 	 *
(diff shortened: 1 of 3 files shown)


swissspidy:
Looking at the diff, it seems these were already at severity 7, no? 🤔

ernilambar:
> Looking at the diff, it seems these were already at severity 7, no? 🤔

Earlier only three were severity 7, `WordPress.WP.I18n.InterpolatedVariableDomain`, `WordPress.WP.I18n.NonSingularStringLiteralContext` and `WordPress.WP.I18n.NonSingularStringLiteralDomain`.

Now other codes will also be updated like `WordPress.WP.I18n.InterpolatedVariableXXX`. `XXX` could contain `Single`, `Plural`, etc.

swissspidy (approved):

davidperezgar (approved):

PR #941, opened 2025-05-13
Check for discouraged functions

### LINKED ISSUE #909: Check: Using load_plugin_textdomain() for loading the plugin translations is not needed for WordPress.org directory since WordPress 4.6
We're asking users to not use load_plugin_textdomain in the plugin, as it's not needed in the WordPress.org directory since WordPress 4.6.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/909

### DIFF
--- a/phpcs-rulesets/plugin-review.xml
+++ b/phpcs-rulesets/plugin-review.xml
@@ -173,4 +173,9 @@
 		<severity>7</severity>
 	</rule>
 
+	<!-- Check for discouraged load_plugin_textdomain() call. -->
+	<rule ref="PluginCheck.CodeAnalysis.DiscouragedFunctions.load_plugin_textdomain_load_plugin_textdomain">
+		<severity>6</severity>
+	</rule>
+
 </ruleset>
--- a/phpcs-rulesets/plugin-review.xml
+++ b/phpcs-rulesets/plugin-review.xml
@@ -175,7 +175,8 @@
 
 	<!-- Check for discouraged load_plugin_textdomain() call. -->
 	<rule ref="PluginCheck.CodeAnalysis.DiscouragedFunctions.load_plugin_textdomainFound">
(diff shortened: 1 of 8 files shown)


phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/DiscouragedFunctionsSniff.php
```diff
@@ -0,0 +1,50 @@
...
+	 *  )
+	 * )
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		return array(
+			'load_plugin_textdomain' => array(
+				'type'      => 'warning',
+				'message'   => 'Using %s() for loading the plugin translations is not needed for WordPress.org directory since WordPress 4.6.',
```
swissspidy: Can this sniff take into account the plugin's `Requires at least` value accordingly? If it still supports older versions, then the call can still be there.

ernilambar: PR has been updated.

phpcs-rulesets/plugin-review.xml
```diff
@@ -173,4 +173,9 @@
 		<severity>7</severity>
 	</rule>
 
+	<!-- Check for discouraged load_plugin_textdomain() call. -->
+	<rule ref="PluginCheck.CodeAnalysis.DiscouragedFunctions.load_plugin_textdomainFound">
+		<severity>6</severity>
```
davidperezgar: Why not 7 for wordpress.org submission?

swissspidy (approved):
If y'all think the error message will be clear enough for developers / there is solid documentation, then this looks good.

davidperezgar (approved):

PR #943, opened 2025-05-15
Add Readme Parser PHP class

ernilambar:
In this PR:

- Adds `Parser.php` class
- Parser class in direct copy of the class obtained from WordPress.org. Only thing changed is the namespace.
- Ref: [Parser in WordPress.org](https://github.com/WordPress/wordpress.org/blob/0da845425e77183308f99f6a4e5c095be5d91233/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php)


dd32:
Thank you for changing the namespace 🙇

swissspidy:
Why still put it in a separate folder? We could just drop the single file somewhere in our existing codebase.
If we still want to use Composer, then I‘d jeep the dependency but use composer-patches to change the namespace.

swissspidy (approved):

davidperezgar (approved):

PR #947, opened 2025-05-23
Use dotorg readme parser if available

ernilambar:
This PR:

- If `WordPressdotorg\Plugin_Directory\Readme\Parser` class is available, use it for parsing readme. Otherwise use `WordPress\Plugin_Check\Lib\Readme\Parser`


dd32 (approved):
Haven't tested this, but looks like it'll work just fine for the intended purpose.

I got to thinking the WordPress.org plugin directory could probably do something like `class_alias( '\WordPressdotorg\Plugin_Directory\Readme\Parser', '\WordPress\Plugin_Check\Lib\Readme\Parser' );` to override PCP :)

Probably best to do this on the PCP side though, as that'll leave the door open for a future where you need to use a custom parser for some reason.. 

I'll also have to remember to use `class_exists( ImportedClassName::class )` :)

davidperezgar (approved):

PR #967, opened 2025-06-17
Add CC0 to GPL-compatible license check

basteyy:
Foremost, thank you for the awesome plugin—it’s a pleasure to use, and I appreciate all your hard work.

I ran into an issue when creating a plugin under the `CC0 1.0 Universal` license: the license check fails. In my plugin header:

```php
/**
 * License:         CC0-1.0
 * License URI:     https://creativecommons.org/publicdomain/zero/1.0/
 */
```

And in `readme.txt`:
```text
License: CC0-1.0
```

However, this triggers the following error:
```txt
0 	0 	ERROR 	plugin_header_invalid_license 	Invalid License: CC0-1.0. 
Please update your Plugin Header with a valid GPLv2 (or later) compatible license.
```

Per the [GNU license list](https://www.gnu.org/licenses/license-list.html#CC0), CC0 1.0 Universal is indeed GPL-compatible. I’ve updated the [is_license_gpl_compatible()](https://github.com/WordPress/plugin-check/blob/trunk/includes/Traits/License_Utils.php#L85) method to include CC0 in the regex (see code below).

This small, first commit ensures that CC0-licensed works are correctly recognized as GPL-compatible.

### DIFF
--- a/includes/Traits/License_Utils.php
+++ b/includes/Traits/License_Utils.php
@@ -83,7 +83,7 @@ trait License_Utils {
 	 * @return bool true if the license is GPL compatible, otherwise false.
 	 */
 	protected function is_license_gpl_compatible( $license ) {
-		$match = preg_match( '/GPL|GNU|MIT|FreeBSD|New BSD|BSD-3-Clause|BSD 3 Clause|OpenLDAP|Expat|Apache2|MPL20|ISC/im', $license );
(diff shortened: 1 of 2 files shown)


davidperezgar:
Hello, I see that is compatible. You need to add tests for this new license.

basteyy:
> Hello, I see that is compatible. You need to add tests for this new license.

Thanks for pointing that out. I added it to the testcase (https://github.com/WordPress/plugin-check/pull/967/commits/9a1f315b4911877b69348ff75bb63065dc0b30a7)

davidperezgar (approved):

ernilambar (approved):

PR #969, opened 2025-06-18
Update ruleset

ernilambar:


ernilambar:
This change is not necessary as of now. Closing.

PR #987, opened 2025-06-26
Fix wrong regex for WP 10+

kkmuffme:
In this PR:
- Version_Utils - Adds support for "XX.X" WordPress version format. Earlier "X.X" was only supported.
- Added unit tests for the change


ernilambar:
Thanks for the PR. Can you please add some tests for the change? Thanks.

kkmuffme:
If there were any tests that tested the previous behavior I could, but I don't have the time to create completely new ones. Since you co-authored, I guess that's the part where you come in? :-)

davidperezgar (approved):

frantorres (approved):

PR #1004, opened 2025-07-16
Support strict output format

### LINKED ISSUE #748: CLI: fully structured output for easier parsing, send to a file
I'm integrating plugin-check into a testing pipeline via wp-cli, and consuming the output programmatically is a little annoying (though totally doable) at the moment. Two things that would make it easier:

- Have a fully JSON output, rather than the mix of `FILE: path.php\n<json blob>` there is now. This is relatively easy to parse, but still requires custom code to do so (and associate the path with each entry). Either a nested structure like `[{path: "path.php", items: <same json blob as now>}, ...]` or just a straight list like `[<single item as now but with path added>,...]` would work fine for this. If memory is a concern, then the second option would also work in a line-by-line format, rather than one big JSON array.
- Allow directing the results to a file, so the status messages can still show on the console (and thus in the logs of a pipeline running this) while the output is clean.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/748

- Introduce more options for `--format` 
    * strict-table
    * strict-csv
    * strict-json
- With above new formats, **FILE:** part in the output will be omitted.
- Added feature test for the changes

### DIFF
--- a/tests/behat/includes/FeatureContext.php
+++ b/tests/behat/includes/FeatureContext.php
@@ -124,6 +124,19 @@ final class FeatureContext extends WP_CLI_FeatureContext {
 	}
 
 	/**
+	 * @Then /^STDOUT should be valid JSON$/
+	 */
+	public function stdout_should_be_valid_json() {
+		$output = $this->result->stdout;
+
+		json_decode( $output );
+
+		if ( JSON_ERROR_NONE !== json_last_error() ) {
+			throw new Exception( 'STDOUT is not valid JSON: ' . json_last_error_msg() );
+		}
+	}
+
+	/**
 	 * Ensure that a requested directory exists and create it recursively as needed.
 	 *
 	 * @param string $directory Directory to ensure the existence of.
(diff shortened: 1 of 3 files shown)


swissspidy:
Why a separate flag and not something like `--format=strict-json`?

Should we perhaps just fix the existing `json` format? Or is it still useful?

tests/behat/includes/FeatureContext.php
```diff
@@ -123,6 +123,19 @@ public function when_i_run_the_wp_cli_command( $mode, $command ) {
...
+	public function stdout_should_be_valid_json() {
+		$output = $this->result->stdout;
+
+		json_decode( $output );
+
+		if ( JSON_ERROR_NONE !== json_last_error() ) {
+			throw new Exception( 'STDOUT is not valid JSON: ' . json_last_error_msg() );
+		}
+	}
+
```
swissspidy, line 138: Nice!

ernilambar:
> Why a separate flag and not something like `--format=strict-json`?

That sounds like a good idea.
 
> Should we perhaps just fix the existing `json` format? Or is it still useful?

That would be a major breaking change I believe. I dont want to break which is already working :-)

frantorres (approved):

davidperezgar (approved):

PR #1015, opened 2025-07-29
Introduce `mode` argument

### LINKED ISSUE #970: Include option for checks if it runs in an published plugin
We actually run all checks in Plugin Repo category and actually is running for new plugins in Directory. We need to add an option to pass that the plugin is already published, and it will cause some changes in checks. For example: we cannot ask for changing prefixes as it will break WordPress installations.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/970


davidperezgar (approved):

davidperezgar:
LGTM

frantorres (approved):

PR #1016, opened 2025-07-29
README.md: update instructions on how the plugin can be installed

rodrigoprimo:
As far as I can see, the plugin was already published to the WP.org repository, so the previous information was not accurate anymore. I'm not familiar with this repository, so apologies if I missed something and this change that I'm suggesting doesn't make sense for some reason that I'm missing.


davidperezgar:
Thanks Rodrigo! you're right. We didn't remind to change the description. LGTM.

davidperezgar (approved):

ernilambar (approved):
LGTM

PR #1017, opened 2025-07-29
Remove unused class imports from two sniffs

rodrigoprimo:
This PR just removes two unused class imports from two sniffs.


swissspidy (approved):

ernilambar:
This should have been handled by slevomat standard. Looks like we have not added slevomat standard for the phpcs-sniffs package. We should do it in the followup.

ernilambar (approved):

PR #1022, opened 2025-07-30
EnqueuedResourceOffloading: bail early when handling empty named parameters and a few more tests

rodrigoprimo:
### EnqueuedResourceOffloading: rename test case file

Doing this to be able to add more test case files with intentional syntax errors.

### EnqueuedResourceOffloading: bail early when handling empty named parameter

This is a tiny optimization as there is no reason to continue running the sniff code if handling an empty named parameter. Doing so means that, as far as I can see, there will always be a non-empty token between start and the end of the parameter, so `$error_prt` will always point to a token and will never be `false`. Thus, the `false === $error_prt` check can be removed.

### EnqueuedResourceOffloading: add a few more tests

- Ensures the check for offloading resources is case-insensitive.
- Ensures the sniff handles PHP 8.0+ named parameters correctly.
- Add a test for the if condition that bails early if the parameter we are looking for is not found.

The sniff was already handling all of the above correctly, but it was not covered by tests.


phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/EnqueuedResourceOffloadingUnitTest.1.inc
```diff
@@ -0,0 +1,25 @@
...
+);
+
+wp_enqueue_style(
+	'foo-style',
+	'https://BOOTSTRAPCDN.COM/awesome-style/foo/bar.css',
+	array(),
+	'1.0.0'
+);
+
+// Safeguard support for PHP 8.0+ named parameters.
```
ernilambar, line 18: Nice.

ernilambar (approved):

davidperezgar (approved):

PR #1023, opened 2025-07-30
Localhost: make some improvements to the sniff code

rodrigoprimo:
Similar to what I did for the `EnqueuedResourceOffloading` sniff (#1018), today I checked the `Localhost` sniff and noticed some opportunities for some improvements, which I'm combining in this PR. Below is a list of the changes that I'm suggesting.

### Localhost: the regex pattern should be case-insensitive

`http://LOCALHOST` is valid, and I imagine it should be forbidden as well.

### Localhost: remove unnecessary code

PHPCS will skip checking tokens if what is returned by `LocalhostSniff::process_token()` is greater than the current stack pointer:

https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/dccc53f0632a453ae8bddb4e065a85125488fca3/src/Files/File.php#L440

Returning `$stackPtr + 1` will not skip any tokens and thus is not necessary, at least as far as I can see. The returned value will be equal to the current stack pointer in the next pass of the foreach loop instead of greater than the current stack pointer.

### Localhost: update the error message to also include "*.local"

Besides forbidding "localhost"  and "127.0.0.1", the sniff also forbids "*.local". This commit changes the error message to mention "*.local" as it was not previously mentioned.

### Localhost: correctly identify error line for heredoc/nowdoc strings

The sniff was incorrectly identifying the error line for URLs inside heredoc/nowdoc. Considering the nowdoc string added as a test in this commit, the sniff was identifying the error lines as 41 and 42 instead of the correct 40 and 41.

This was happening because of an error in the logic of the `find_token_in_multiline_string()` method. As far as I could test, this method is actually not necessary, as just passing `$stackPtr` to `addError()` should be enough for it to correctly identify the error line.

PHPCS tokenizes multi-line strings using one token per line. So `$stackPtr` will correspond to the correct error line (the line that contains the localhost URL).

In a sniff like `Offloading`, it makes sense to use this method as it combines the content of multi-line strings using `TextStrings::getCompleteTextString()`, but that is not the case for `Localhost`.

Removing `find_token_in_multiline_string()` also meant that it was not necessary anymore to pass the `PREG_OFFSET_CAPTURE` flag to `preg_match_all()` as the information about the offset is not used anymore.

### Localhost: update a test to use a T_DOUBLE_QUOTED_STRING

The sniff listens for `T_DOUBLE_QUOTED_STRING`, and before, there were no tests with this type of token.

### DIFF
--- a/phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/LocalhostUnitTest.inc
+++ b/phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/LocalhostUnitTest.inc
@@ -3,7 +3,7 @@
 $var            = 'This file contains https://127.0.0.1/example url.'; // Error.
 $sample_url     = 'Sample URL is http://docker.local/example here.'; // Error.
 $custom_url     = 'Custom URL https://docker.localhost/example here.'; // Error.
-$custom_content = file_get_contents("http://127.0.0.1/api.php?url=".$url); // Error.
+$custom_content = file_get_contents("http://127.0.0.1/api.php?url=$url"); // Error.
 
 wp_register_script( 'custom-js', 'https://127.0.0.1/dist/custom.js' ); // Error.
 wp_enqueue_script( 'custom-js', 'https://127.0.0.1/dist/custom.js' ); // Error.
--- a/phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/LocalhostUnitTest.inc
+++ b/phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/LocalhostUnitTest.inc
@@ -28,3 +28,15 @@ $multiple_urls = 'Custom URL https://docker.localhost/example and https://localh
 <img src="<?php echo "http://localhost/image.png"; ?>" alt="" /><!-- Error. -->
 
 <p>URL in inline HTML is http://localhost/example2.php bad.</p><!-- Error. -->
+
+<?php
+
+$multiline_string = 'A multi-line string with URLs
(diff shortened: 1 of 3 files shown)


phpcs-sniffs/PluginCheck/Tests/CodeAnalysis/LocalhostUnitTest.inc
```diff
@@ -3,12 +3,12 @@
 $var            = 'This file contains https://127.0.0.1/example url.'; // Error.
 $sample_url     = 'Sample URL is http://docker.local/example here.'; // Error.
 $custom_url     = 'Custom URL https://docker.localhost/example here.'; // Error.
-$custom_content = file_get_contents("http://127.0.0.1/api.php?url=".$url); // Error.
+$custom_content = file_get_contents("http://127.0.0.1/api.php?url=$url"); // Error.
```
ernilambar, line 7: May be it would be good to add new test cases instead of modifying existing ones.

rodrigoprimo: What exactly is being tested in this original test that is not already being tested in the other tests to justify keeping it? I don't see anything, and that is why I opted to replace it. The fact that it is one of the few tests with double quotes indicates to me that maybe the original intention of this test was to test the sniff handling a `T_DOUBLE_QUOTED_STRING` token.

ernilambar (approved):

davidperezgar (approved):

PR #1027, opened 2025-08-04
Check for missing textdomain args

ernilambar:


ernilambar:
We are not going to use this approach. Closing PR.

PR #1028, opened 2025-08-05
Prevent deletion of custom WordPress tables during cleanup

### LINKED ISSUE #907: Erroneously drops tables
I see that during testing, a load of custom tables are created, then dropped on completion.

However, when using non-standard [user table names](https://developer.wordpress.org/reference/classes/wpdb/tables/) via `CUSTOM_USER_TABLE` and `CUSTOM_USER_META_TABLE` these are not duplicated and are dropped during the cleanup method, the following tables are dropped:

```
[
    [users] => wp_my_custom_users
    [usermeta] => wp_my_custom_user_meta
    [posts] => wp_pc_posts
    [comments] => wp_pc_comments
    [links] => wp_pc_links
    [options] => wp_pc_options
    [postmeta] => wp_pc_postmeta
    [terms] => wp_pc_terms
    [term_taxonomy] => wp_pc_term_taxonomy
    [term_relationships] => wp_pc_term_relationships
    [termmeta] => wp_pc_termmeta
    [commentmeta] => wp_pc_commentmeta
]
```

as you see, the two user tables are not `wp_pc_` prefixed.

kofimokome:
Fixes #907


includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -223,11 +229,10 @@ public function can_set_up() {
 	/**
 	 * Installs WordPress, while providing tweaks to allow for early execution of the install process.
 	 *
-	 * @since 1.3.0
```
swissspidy, line 226: Could you please undo all the unrelated changes like this?

kofimokome: okay. It's the work of my IDE 🤦

kofimokome: okay. It's the work of my IDE 🤦

kofimokome: Done.

felixarntz (changes requested):
@kofimokome Thank you for the PR! However, this is only a partial solution for the bug reported in #907.

We need to do more than prevent deletion of other tables: We also need to make sure that the table duplication still works for sites that use `CUSTOM_USER_TABLE` and `CUSTOM_USER_META_TABLE`.

* When creating the duplicate tables, we need to make sure there are duplicates for these two as well.
* When writing or reading data from these two tables, we need to make sure the duplicates are used instead of the real ones.

Since the constants are used for this, it may be more tricky than for the regular tables. But we need to look into what our options are to address this - maybe some low-level filters.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -110,6 +110,14 @@ public function clean_up() {
 		$prefix_cleanup = $this->amend_db_base_prefix();
 		$tables         = $wpdb->tables();
 
+		// Remove tables not starting with the prefix.
+		$tables = array_filter(
+			$tables,
+			function ( $table ) use ( $wpdb ) {
+				return str_starts_with( $table, $wpdb->prefix );
+			}
+		);
```
felixarntz: How about this instead? It focuses more concretely on the problem and is more efficient than iterating through all tables.
```suggestion
		// Do not remove custom tables (which by definition weren't duplicated because we cannot override constants).
		if ( isset( $tables['users'] ) && defined( 'CUSTOM_USER_TABLE' ) && $tables['users'] === CUSTOM_USER_TABLE ) {
			unset( $tables['users'] );
		}
		if ( isset( $tables['usermeta'] ) && defined( 'CUSTOM_USER_META_TABLE' ) && $tables['usermeta'] === CUSTOM_USER_META_TABLE ) {
			unset( $tables['usermeta'] );
		}
```

felixarntz (commented):
Revising my last feedback, it looks like it's indeed not possible to duplicate the `users` and `usermeta` table if they use the constants, because we can't override constants.

I think it's okay for now to focus on purely not deleting them, as that's most important.

We should however also open an issue, to prevent executing runtime checks on sites where one or both of these constants are defined, as that is too risky and can lead to problems.

felixarntz:
I opened #1062 for the other piece I mentioned.

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -138,6 +140,23 @@ public function clean_up() {
 		}
 	}
 
+	/**
+	 * Excludes the custom user and user meta tables from the list of tables to be deleted.
+	 *
+	 * @since 1.7.0
+	 *
+	 * @param array $tables          List of WordPress database tables.
```
felixarntz:
```suggestion
	 * @param array $tables List of WordPress database tables.
```

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -138,6 +140,23 @@ public function clean_up() {
...
+	 */
+	private function ignore_custom_tables( array &$tables ) {
```
felixarntz: It's not very common in WordPress to pass things by reference. Can you maybe update this to pass by value and return the updated `array`? Not strictly a blocker, but it feels a bit more WordPress-y :)

felixarntz (approved):
@kofimokome Two small points of feedback, but overall this looks good.

includes/Checker/Runtime_Environment_Setup.php
kofimokome: Okay. fixing it now

includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -138,6 +140,26 @@ public function clean_up() {
...
 	}
 
+	/**
+	 * Excludes the custom user and user meta tables from the list of tables to be deleted.
+	 *
+	 * @since 1.7.0
+	 *
+	 * @param array $tables List of WordPress database tables.
+	 *
+	 * @return array List of WordPress database tables to delete
```
felixarntz: Per WP coding standards, there's normally no blank line between `@param` and `@return`. Also a missing period at the end of the return description.
```suggestion
	 * @param array $tables List of WordPress database tables.
	 * @return array List of WordPress database tables to delete.
```

felixarntz (approved):
@kofimokome Thanks!

ernilambar (approved):

PR #1029, opened 2025-08-08
Implement slevomat rules for sniffs

ernilambar:
- Implements `slevomat/coding-standard` for `phpcs-sniffs`

### DIFF
--- a/phpcs-sniffs/PluginCheck/Tests/AbstractSniffUnitTest.php
+++ b/phpcs-sniffs/PluginCheck/Tests/AbstractSniffUnitTest.php
@@ -8,9 +8,9 @@
 namespace PluginCheckCS\PluginCheck\Tests;
 
 use PHP_CodeSniffer\Config;
-use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest as BaseAbstractSniffUnitTest;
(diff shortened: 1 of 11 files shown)


dd32:
These are all very personal coding practices that probably shouldn't be enforced for plugin developers.. IMHO..

@ernilambar Can you expand on what you're attempting to implement either in the PR description or a linked issue in future? The title/description of this PR provides no context to someone who's watching the repo.

ernilambar:
@dd32 These are the rules we have already been using in the main repo. https://github.com/WordPress/plugin-check/blob/trunk/phpcs.xml.dist I am trying to make uniform in both main repo and "phpcs-sniffs" folder.

dd32:
Gotcha..  the main repo is for sniffs against **this** plugin though, and the phpcs-sniffs is against the **tested** plugin right? It would make sense that plugin-check holds itself to an opinionated code-style, while allows developers to use their own practices?

swissspidy (approved):

swissspidy:
@dd32 There's a misunderstanding. The change here only affects the coding style in this repo, it's not about enforcing something for plugin developers.

davidperezgar (approved):

PR #1030, opened 2025-08-09
Add The Unlicense to GPL-compatible license check

john-shaffer:
The Unlicense is a GPL-Compatible free software
license, but is not recognized by the is_license_gpl_compatible check.
See https://www.gnu.org/licenses/license-list.html#Unlicense.

- Add The Unlicense to the regex in is_license_gpl_compatible() so the plugin will correctly recognize it as GPL-compatible.
- Add tests for recognition of The Unlicense

### DIFF
--- a/includes/Traits/License_Utils.php
+++ b/includes/Traits/License_Utils.php
@@ -55,6 +55,7 @@ trait License_Utils {
 		$license = str_replace( ' version ', 'v', $license );
 		$license = preg_replace( '/GPL\s*[-|\.]*\s*[v]?([0-9])(\.[0])?/i', 'GPL$1', $license, 1 );
 		$license = preg_replace( '/Apache.*?([0-9])(\.[0])?/i', 'Apache$1', $license );
+		$license = preg_replace( '/(The\s+)?Unlicense/i', 'Unlicense', $license );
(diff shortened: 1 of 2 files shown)


ernilambar:
@john-shaffer Thanks for the PR. I believe following methods also need to be updated in `License_Utils.php`
- get_normalized_license()
- is_license_valid_identifier()

You can test it by keeping `License: Unlicense` in plugin header.

davidperezgar:
@john-shaffer did you see the comments?

john-shaffer:
> @john-shaffer Thanks for the PR. I believe following methods also need to be updated in `License_Utils.php`
> 
>     * get_normalized_license()
> 
>     * is_license_valid_identifier()
> 
> 
> You can test it by keeping `License: Unlicense` in plugin header.

Good catch. I updated get_normalized_license to normalize it to "Unlicense". The checks pass for me with any combination of "The Unlicense" and "Unlicense" in readme.txt and ${plugin}.php.

I don't think anything in is_license_valid_identifier should be changed since it only checks the normalized name, which will always pass.


> @john-shaffer did you see the comments?

Yes

ernilambar (approved):
LGTM.

davidperezgar (approved):

PR #1038, opened 2025-08-26
Update composer dependencies

ernilambar:
In this PR:

* Upgrade `wp-cli-tests` to version 5.
* Use latest `php-compatibility` package
* Update code coverage setup

### DIFF
--- a/tests/behat/utils/maybe-generate-wp-cli-coverage.php
+++ b/tests/behat/utils/maybe-generate-wp-cli-coverage.php
@@ -2,56 +2,46 @@
 
 use SebastianBergmann\CodeCoverage\CodeCoverage;
 use SebastianBergmann\CodeCoverage\Driver\Selector;
-use SebastianBergmann\CodeCoverage\Driver\Xdebug;
 use SebastianBergmann\CodeCoverage\Filter;
-use SebastianBergmann\CodeCoverage\Report\PHP as PHPReport;
+use SebastianBergmann\CodeCoverage\Report\Clover;
 
 $root_folder = realpath( dirname( __DIR__, 3 ) );
-
-$feature   = getenv( 'BEHAT_FEATURE_TITLE' );
-$scenario  = getenv( 'BEHAT_SCENARIO_TITLE' );
-$step_line = (int) getenv( 'BEHAT_STEP_LINE' );
-$name      = "{$feature} - {$scenario} - {$step_line}";
-
-if ( empty( $feature ) || empty( $scenario ) ) {
-	return;
-}
 
 if ( ! class_exists( 'SebastianBergmann\CodeCoverage\Filter' ) ) {
 	require "{$root_folder}/vendor/autoload.php";
 }
 
 $filtered_items = new CallbackFilterIterator(
-	new DirectoryIterator( $root_folder ),
-	function ( $file ) {
-		if ( $file->isDir() && in_array( $file->getFilename(), [ 'php', 'src' ], true ) ) {
-			return true;
-		}
+  new DirectoryIterator( $root_folder ),
+  function ( $file ) {
+    if ( $file->isDir() && in_array( $file->getFilename(), [ 'includes' ], true ) ) {
+      return true;
+    }
 
-		if ( $file->isFile() && false !== strpos( $file->getFilename(), '-command.php' ) ) {
-			return true;
-		}
+    if ( $file->isFile() && false !== strpos( $file->getFilename(), 'plugin.php' ) ) {
+      return true;
+    }
 
-		return false;
-	}
+    return false;
+  }
 );
 
 $files = [];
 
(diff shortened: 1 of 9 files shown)


swissspidy:
Looks like `tests/behat/utils/maybe-generate-wp-cli-coverage.php` needs some updating now that newer versions of relevant dependencies are installed. May need to compare with https://github.com/wp-cli/wp-cli-tests/blob/5d933e384003af15161099174771fbf740cc2e4a/utils/generate-coverage.php

ernilambar:
@swissspidy In PHP 8.4, test is not passing. Error message: `Error: The operation was canceled.`. I am not being able to debug why it was cancelled.

swissspidy:
It was cancelled because it took too long:

> The job has exceeded the maximum execution time of 20m0s

My guess is now that too many files and directories are included for code coverage, so it takes too long.

Did you just copy the whole `generate-coverage.php` file from WP-CLI? It needs to be adapted for the plugin... I'll leave some comments inline.

tests/behat/utils/maybe-generate-wp-cli-coverage.php
```diff
@@ -2,41 +2,87 @@
...
+			}
+		}
+	} else {
+		$files[] = $item->getPathname();
+	}
+}
+
 $filter = new Filter();
-$filter->includeDirectory( "{$root_folder}/includes" );
-$filter->includeFiles( array( "{$root_folder}/plugin.php" ) );
```
swissspidy, line 16: This logic needs to be restored: we only want coverage for the `includes` folder and the `plugin.php` file.

tests/behat/utils/maybe-generate-wp-cli-coverage.php
```diff
@@ -2,41 +2,87 @@
...
+$filtered_items = new CallbackFilterIterator(
+	new DirectoryIterator( $root_folder ),
+	function ( $file ) {
+		if ( $file->isDir() && in_array( $file->getFilename(), [ 'php', 'src' ], true ) ) {
+			return true;
+		}
+
+		if ( $file->isFile() && false !== strpos( $file->getFilename(), '-command.php' ) ) {
+			return true;
+		}
```
swissspidy:
```suggestion
		if ( $file->isDir() && in_array( $file->getFilename(), [ 'includes' ], true ) ) {
			return true;
		}

		if ( $file->isFile() && false !== strpos( $file->getFilename(), 'plugin.php' ) ) {
			return true;
		}
```

tests/behat/utils/maybe-generate-wp-cli-coverage.php
```diff
@@ -2,41 +2,87 @@
...
+		$db_type         = strtolower( getenv( 'WP_CLI_TEST_DBTYPE' ) );
+		$destination     = "$root_folder/build/logs/$feature_suffix-$scenario_suffix-$step_line-$db_type.cov";
+
+		$dir = dirname( $destination );
+		if ( ! file_exists( $dir ) ) {
+			mkdir( $dir, 0777, true );
+		}
 
-		( new Clover() )->process( $coverage, $destination, $name );
+		( new PHPReport() )->process( $coverage, $destination );
```
swissspidy: I suggest reverting this and keep using `Clover`.

What WP-CLI does for the test suite is generating separate PHP reports per scenario, and then glueing them together after all tests ran. See https://github.com/wp-cli/wp-cli-tests/blob/5d933e384003af15161099174771fbf740cc2e4a/src/Context/FeatureContext.php#L285-L310

I don't think we need this for this plugin (it's not that big).

tests/behat/utils/maybe-generate-wp-cli-coverage.php
```diff
@@ -2,41 +2,87 @@
...
 		$coverage->stop();
 
-		$project_dir = getenv( 'BEHAT_PROJECT_DIR' );
-
 		$feature_suffix  = preg_replace( '/[^a-z0-9]+/', '-', strtolower( $feature ) );
 		$scenario_suffix = preg_replace( '/[^a-z0-9]+/', '-', strtolower( $scenario ) );
-		$filename        = "clover-behat/{$feature_suffix}-{$scenario_suffix}.xml";
-		$destination     = "{$project_dir}/build/logs/{$filename}";
+		$db_type         = strtolower( getenv( 'WP_CLI_TEST_DBTYPE' ) );
+		$destination     = "$root_folder/build/logs/$feature_suffix-$scenario_suffix-$step_line-$db_type.cov";
```
swissspidy: Ditto here about reverting.

ernilambar: Ok. I will revert all and update from the fresh state :-)

swissspidy (approved):

frantorres (approved):

PR #1042, opened 2025-08-26
Add textdomain check and sniff

### LINKED ISSUE #1020: Implement missing textdomain argument sniff
The `WordPress.WP.I18n.MissingArgDomain` sniff currently flags when a translation function is missing the text domain argument. However, its current severity level means it might not receive appropriate attention from plugin developers.

### Background
- When the domain argument is missing in WordPress translation functions, the function will default to using core translations
- This behavior works technically but is discouraged in plugin or theme
- In 99.9% of cases, this occurs because plugin authors are unaware they need to specify their text domain, not because they intentionally want to use core translations

### Proposal
Increase the severity level of this sniff from its current level to **7 (as a blocker in PCP)**.

### Suggested improvement provided from Plugins Team meeting

Compare plugin strings with a map of WordPress core strings to reduce false positives when checking for missing text domains.

For example:

```php
echo esc_html__( 'Submit' ): // WARNING: This string exists in WP core.
echo esc_html__( 'Example plugin specific string' ): // ERROR: This string is not in core.
```

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/1020

**In this PR:**

* New sniff has been added `PluginCheck.CodeAnalysis.I18nTextDomain`
* This sniff checks string without text domain and generates ERROR if string is not in the core translations
* Core translations are kept here. Currently WP 6.8 strings are kept. `/phpcs-sniffs/PluginCheck/Vars`
* `phpcs-sniffs/bin` added for refining raw PHP translation files obtained from GlotPress.
* Added new check using above sniff - `I18n_Text_Domain_Check`
* Errors from the check are blockers as severity is kept as 7.
* Unit tests have been added for new check


swissspidy:
> This sniff checks string without text domain and generates ERROR if string is not in the core translations

Is this really worth it? 😅  That adds quite a lot of complexity for something that's generally discouraged (reusing core translations). I'd rather be more strict at disallowing that rather than trying to make it work for the few plugins doing that.

ernilambar:
After discussion in the team, we have decided not to pursue this approach. Closing the PR.

PR #1043, opened 2025-08-26
Detect links that request five-star reviews

### LINKED ISSUE #1037: Check: Authors asking for a 5 stars review
We detect when authors ask users to leave a 5-star review. We flag this code and ask the author to change it, as stated in the review policy.

> ## Linking directly to 5 stars reviews.
> 
> Linking to a filtered view of the reviews section to only show 5-star ratings presents a misleading picture of the plugin’s overall feedback and reputation.
> 
> It hides negative or constructive reviews, which could be critical for users to make informed decisions.
> 
> ✅ What’s the proper way to do it?
> 
> If you want to link to reviews, it’s best to link to the main reviews page:
> https://wordpress.org/support/plugin/plugin-slug/reviews/
> You can certainly say something like "Check out our great reviews!" — but you should let users see all reviews, not just the favorable ones.

davidperezgar:
Fixes #1037 

I propose changing the name of the class from Code_Obfuscation_Check to Code_NotAllowed_Check in order to make it more generic and allow for more code-related cases.


ernilambar:
I dont think we should rename "Code_Obfuscation_Check". This is specifically focused on obfuscation check. Also class has `const` properties which is focused on checking code obfuscation. Instead we should have new generic check class where we could keep reviews check now and we could add more in the future.

davidperezgar:
We have agreed to use Plugin_Content_Check for a class that will detect issues in the plugin content.

davidperezgar:
At the request of @frantorres, I’ve enforced the check to detect wp.org and wordpress.org at the beginning.

frantorres (approved):

ernilambar (approved):

PR #1057, opened 2025-09-20
New Check wp_safe_redirect

### LINKED ISSUE #1056: [Check] Encourage use of wp_safe_redirect()
We want to migrate all checks from WPorg Code Analysis to PCP, and I noticed that [this check](https://github.com/WordPress/wporg-code-analysis/blob/9cc6dcfa9a16155c2879c71474e6621c02e98566/MinimalPluginStandard/ruleset.xml#L30) is not available in PCP.

This check is flagged as a warning, so I suppose we can assign it the same level in PCP.

davidperezgar:
Fixes #1056


ernilambar:
@davidperezgar Instead of 4 test plugins, I think we could use one plugin but have multiple PHP files under it. We already have lots of test plugins there.

ernilambar (approved):

frantorres (approved):

PR #1090, opened 2025-10-27
chore: Ignore vendor_prefixed and vendor-prefixed folders

### LINKED ISSUE #1083: Exclude "vendor-prefixed" along with "vendor"
Currently the `vendor` folder is excluded, along with `node_modules`.

Many plugins use `vendor-prefixed/` or `vendor_prefixed/` when using dependency-prefixing tools, and this causes plugin-check to needlessly scan those files.

For consistency of scanning, IMHO we should exclude `vendor-prefixed/` in addition to `vendor/`

kofimokome:
Fixes #1083


davidperezgar (approved):

ernilambar (approved):

koen12344:
This only partially fixes the issue, `vendor-prefixed` is the default folder name, but it is dynamic and [can be overridden](https://github.com/BrianHenryIE/strauss?tab=readme-ov-file#configuration) in composer.json:

```
    "extra": {
      "strauss": {
        "target_directory": "vendor-prefixed",
```

Ideally the plugin should pull the folder name from the composer.json file

PR #1092, opened 2025-10-27
chore: Update localhost sniff regex

### LINKED ISSUE #1019: LocalhostSniff doesn't match protocol relative URLs
The LocalhostSniff does not match protocol relative URLs, while I believe it should, as this is a valid form of URL.

Here is an example of a code that, in my opinion, should trigger an error, but currently doesn't:

```php
wp_register_script( 'custom-js', '//127.0.0.1/dist/custom.js' ); 
```

I'm unsure how easily the regex can be adapted to match this case, as the sniff also looks for localhost URLs in the middle of text strings and not only the ones that start with a localhost URL.

kofimokome:
Fixes #1019

### DIFF
--- a/phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/LocalhostSniff.php
+++ b/phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/LocalhostSniff.php
@@ -48,7 +48,7 @@ final class LocalhostSniff extends Sniff {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


ernilambar:
@kofimokome Can you please add unit test-case like mentioned in the related issue?

kofimokome:
Hi @ernilambar. I have added a unit test.

davidperezgar (approved):

ernilambar (approved):

rodrigoprimo:
Not a big deal, but maybe it would have been a good idea to also include a test where the string starts with a protocol relative URL to safeguard against regressions for this case?

davidperezgar:
It has been merged, so we can move it as issue to prepare a new code.

PR #1105, opened 2025-11-19
Add short URL sniff

### LINKED ISSUE #960: Check: Do not use shortened URLs when possible
> The use of URL shorteners (like bit.ly or tinyurl) must not be used inside plugins or readmes as the primary use of a short URL in a plugin is to disguise or hide an affiliate link.
> 
> Since links in plugins are used as actual links (that is, with a-href code), there is no practical need to shorten them. The full and complete URL should be used, so users may make an educated choice as to if they wish to click on a URL or not.
> 
> Keep in mind, you are permitted to use UTM tracking in URLs to your own domains, but again those links must be expanded so users can be made aware before they click the link.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/960

### DIFF
--- a/phpcs-rulesets/plugin-check.ruleset.xml
+++ b/phpcs-rulesets/plugin-check.ruleset.xml
@@ -35,6 +35,11 @@
 		<severity>7</severity>
 	</rule>
 
+	<!-- Check for short URLs. -->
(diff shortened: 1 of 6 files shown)


davidperezgar (approved):

ryanwelcher (approved):
Looks good to me!

PR #1133, opened 2025-12-17
Check Direct File Access

### LINKED ISSUE #603: Check: Allowing Direct File Access to plugin files
This check consists to avoid direct file access if someone directly queries a file. 

For files that only contain a PHP class the risk of something funky happening when directly accessed is pretty small. For files that contain procedural code, functions and function calls, the chance of security risks is a lot bigger.

We can avoid this by putting this code at the top of all PHP files that could potentially execute code if accessed directly :    
`if ( ! defined( 'ABSPATH' ) ) exit; // Exit if accessed directly`
 
What we do actually in the internal scanner:
1. Check that is not an uninstall file.
2. Check that is valid for direct access.
3. Look for ABPATH or WPINC, otherwise it's wrong.

Finally in results we say all files that does not contain this header.

davidperezgar:
This check is based on the code that @frantorres created for the Internal Scanner.
There are many cases based on past experience.
Fixes #603

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
@@ -11,6 +11,7 @@ use WordPress\Plugin_Check\Checker\Check_Categories;
 use WordPress\Plugin_Check\Checker\Check_Result;
 use WordPress\Plugin_Check\Checker\Checks\Abstract_File_Check;
 use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Find_Uninstall;
 use WordPress\Plugin_Check\Traits\Stable_Check;
 
 /**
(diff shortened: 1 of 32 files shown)


includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
```diff
@@ -0,0 +1,336 @@
...
+	/**
+	 * Checks if a file is the uninstall.php file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $file       The file path to check.
+	 * @param string $plugin_path The plugin path.
+	 * @return bool True if the file is uninstall.php, false otherwise.
+	 */
+	private function is_uninstall_file( $file, $plugin_path ) {
```
ernilambar: Logic for checking uninstall file seems to complex. In "Plugin_Uninstall_Check.php" it is pretty simple. What do you think?

includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
```diff
@@ -0,0 +1,336 @@
...
+
+	use Amend_Check_Result;
+	use Stable_Check;
+
+	/**
+	 * Gets the categories for the check.
+	 *
+	 * Every check must have at least one category.
+	 *
+	 * @since n.e.x.t
```
ernilambar:
```suggestion
	 * @since 1.8.0
```

includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
davidperezgar: ok, I'll make it as other check.

frantorres (approved):

ernilambar (approved):

PR #1135, opened 2025-12-20
Check Mismatched Tested up to

### LINKED ISSUE #1134: Mismatched "Tested up to" header between Plugin Header and Readme
**Problem**

The "Tested up to" header can exist in both:
- The plugin's main file header
- The plugin's readme file

Currently, there is no validation to ensure these values match when both are present.

**Effect**

If the "Tested up to" value exists in the plugin header, it could override or conflict with the correct value from the readme file. This can lead to:
- Inconsistent version information
- Confusion about which value is authoritative

**Expected Behavior**

If "Tested up to" exists in both locations, they must be exactly the same.

davidperezgar:
Fixes #1134 

This check validates that the "Tested up to" header matches between the plugin's main file header and the readme.txt file.

## Problem

The "Tested up to" header can exist in both:
- The plugin's main file header
- The plugin's readme.txt file

When both are present, they must be exactly the same. If they differ, it can lead to:
- Inconsistent version information displayed on WordPress.org
- Confusion about which WordPress version the plugin actually supports
- The plugin header value overriding the readme value

## Solution

The check has been integrated into the `Plugin_Header_Fields_Check` class and performs the following validation:

1. **Skips single-file plugins** - Single-file plugins don't have readme.txt files
2. **Checks for readme existence** - If no readme exists, the check is skipped
3. **Compares major versions** - Extracts the major version (e.g., "6.7" from "6.7.1") from both locations
4. **Reports mismatch** - If both values exist and don't match, an error is reported
5. Refactored the `Find_Readme` trait into a more generic `Readme_Utils` trait and moved the `get_readme_tested_value()` method from `Plugin_Header_Fields_Check` into it for better code reusability.

### Location
- **Check Class**: `includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php`
- **Method**: `check_tested_up_to_mismatch()`
- **Test File**: `tests/phpunit/tests/Checker/Checks/Tested_Up_To_Check_Tests.php`
- **Additional Tests**: Added to `tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php`

### Error Code
- `mismatched_tested_up_to_header`

### Severity
- **7** (Error)

## Test Coverage

The following test cases are included:

1. `test_run_with_mismatch()` - Verifies error is reported when values don't match
2. `test_run_with_match()` - Verifies no error when values match
3. `test_run_with_readme_only()` - Verifies no error when only readme has the value
4. `test_run_with_header_only()` - Verifies no error when only header has the value
6. `test_run_with_single_file_plugin()` - Verifies single-file plugins are skipped
7. `test_run_with_no_readme()` - Verifies no error when readme doesn't exist
8. Tests added to `Plugin_Header_Fields_Check_Tests.php` for integration testing

## Usage

The check runs automatically as part of the `Plugin_Header_Fields_Check` and will report errors when:
- Both readme.txt and plugin header contain "Tested up to" values
- The major versions don't match

## Error Message

When a mismatch is detected, the following error is displayed:

```
Mismatched "Tested up to": {readme_version} != {plugin_header_version}.

The "Tested up to" value in the readme file must match the "Tested up to" value in the plugin header. 
If the plugin header has a "Tested up to" value, it will override the readme value, which can cause confusion.
```

## Documentation URL

https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information


includes/Traits/Readme_Utils.php
```diff
@@ -0,0 +1,96 @@
...
+	 * Gets the "Tested up to" value from the readme file.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param string $plugin_path The plugin directory path.
+	 * @return string The "Tested up to" value from readme, or empty string if not found.
+	 */
+	protected function get_readme_tested_value( $plugin_path ) {
+		// Build list of potential readme files.
+		$potential_files = array(
```
ernilambar: We should use existing helper method to find the readme file to avoid code duplication.

davidperezgar: Ok, I have moved to Plugin_Readme_Check so I've more variables.

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -123,6 +123,9 @@ protected function check_files( Check_Result $result, array $files ) {
 
 		// Check the readme for language.
 		$this->check_language( $result, $readme_file, $parser );
+
+		// Check for mismatched "Tested up to" header between plugin header and readme.
+		$this->check_tested_up_to_mismatch( $result, $parser, $result->plugin()->main_file() );
```
ernilambar, line 128: Lets not pass main file from here. We generally fetch inside the method itself from $result as we are already have access of "$result".

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
```diff
@@ -953,6 +956,66 @@ private function check_language( Check_Result $result, string $readme_file, $par
...
+		// Get the "Tested up to" value from the readme.
+		$readme_tested = isset( $parser->tested ) ? $parser->tested : '';
+
+		if ( empty( $readme_tested ) ) {
+			return;
+		}
+
+		// Get the "Tested up to" value from the plugin header.
+		$plugin_header = get_file_data(
+			$plugin_main_file,
```
ernilambar: We already have helper for this "private function get_plugin_data()"

ernilambar: Moreover this check should be in 'Plugin_Header_Fields_Check.php" class.

tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php
```diff
@@ -78,6 +78,46 @@ public function test_run_with_valid_requires_plugins_header() {
...
+		$check_context = new Check_Context( UNIT_TESTS_PLUGIN_DIR . 'test-plugin-tested-up-to-mismatch/load.php' );
+		$check_result  = new Check_Result( $check_context );
+
+		$check->run( $check_result );
+
+		$errors = $check_result->get_errors();
+
+		// The "Tested up to" mismatch check has been moved to Plugin_Readme_Check.
+		// This test now verifies that Plugin_Header_Fields_Check does NOT report this error.
+		if ( isset( $errors['load.php'][0][0] ) ) {
```
ernilambar: We should try to avoid such "if" conditionals in the unit tests.

tests/phpunit/tests/Checker/Checks/Tested_Up_To_Check_Tests.php
```diff
@@ -0,0 +1,145 @@
...
+		$check->run( $check_result );
+
+		$errors   = $check_result->get_errors();
+		$warnings = $check_result->get_warnings();
+
+		// Should not have mismatched tested up to error when values match.
+		// Note: Other readme errors may still be present.
+		if ( ! empty( $errors ) ) {
+			foreach ( $errors as $file => $file_errors ) {
+				if ( isset( $file_errors[0][0] ) ) {
```
ernilambar: "wp_list_filter" approach is better to fetch code here I guess.

tests/phpunit/tests/Checker/Checks/Tested_Up_To_Check_Tests.php
```diff
@@ -0,0 +1,145 @@
+<?php
+/**
```
ernilambar: Instead of creating separate file for this "Tested_Up_To_Check_Tests.php", we could include it in "Plugin_Header_Fields_Check_Tests.php" to simplify.

davidperezgar: The check is in readme checks.

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
davidperezgar: I have moved to Plugin_Readme_Check_Tests.php as it's part of readme checks.

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
davidperezgar: It's easier in Readme as the readme parser is already loaded. My first approach was to load in the Header fields and you need more functions to load.

tests/phpunit/tests/Checker/Checks/Tested_Up_To_Check_Tests.php
davidperezgar: It's part of Readme checks, not Plugin Header.

includes/Checker/Checks/Plugin_Repo/Plugin_Readme_Check.php
davidperezgar: the function get_plugin_data does not support Tested Up, that's why I used

ernilambar (approved):

frantorres (approved):

PR #1137, opened 2025-12-27
Add Write File Check to detect plugin directory writes

### LINKED ISSUE #665: Check: Asks users to edit/write to plugin (use uploads folder)
In our reviews we already say:

> ## Saving data in the plugin folder and/or asking users to edit/write to plugin.
> 
> We cannot accept a plugin that forces (or tells) users to edit the plugin files in order to function, or saves data in the plugin folder.
> 
> Plugin folders are deleted when upgraded, so using them to store any data is problematic. Also bear in mind, any data saved in a plugin folder is accessible by the public. This means anyone can read it and use it without the site-owner’s permission.
> 
> It is preferable that you save your information to the database, via the Settings API, especially if it’s privileged data. If that’s not possible, because you’re uploading media files, you should use the media uploader.
> 
> If you can’t do either of those, you must save the data outside the plugins folder. We recommend using the uploads directory, creating a folder there with the slug of your plugin as name, as that will make your plugin compatible with multisite and other one-off configurations.
> 
> Please refer to the following links:
> 
> https://developer.wordpress.org/plugins/settings/
> https://developer.wordpress.org/reference/functions/media_handle_upload/
> https://developer.wordpress.org/reference/functions/wp_handle_upload/
> https://developer.wordpress.org/reference/functions/wp_upload_dir/

We warn developers to use properly folders to save their data. This check could have positives, so maybe we should use a 6 severity and error. 

We should check the functions: 'fwrite', 'fputs', 'file_put_contents', 'copy', 'rename', 'copy_dir', 'move_dir', and wp_filesystem. Finally, detect the place that they are using to save the data.

davidperezgar:
Fixes #665 

Implements a new check to detect when plugins save data in the plugin folder.

Plugin folders are deleted when upgraded, so using them to store any data is problematic. This check helps developers identify these issues and directs them to use the uploads directory or database instead.

### 1. WriteFileSniff (PHPCS Sniff)
Created `phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/WriteFileSniff.php`:
- Extends `AbstractFunctionParameterSniff` to detect file write functions
- Monitors functions: `fwrite`, `fputs`, `file_put_contents`, `touch`, `copy`, `rename`, `copy_dir`, `move_dir`, `unzip_file`
- Detects plugin directory indicators:
  - Constants: `WP_PLUGIN_DIR`, `WP_PLUGIN_URL`, `PLUGINDIR`, `WPINC`, `WP_CONTENT_DIR`, `WP_CONTENT_URL`
  - Functions: `plugins_url()`, `plugin_dir_path()`, `plugin_dir_url()`
  - Magic constants: `__FILE__`, `__DIR__`
- Allows safe paths using `wp_upload_dir()`, `wp_tempnam()`, `get_temp_dir()`
- Provides clear error messages with remediation guidance

### 2. Write_File_Check Class
Updated `includes/Checker/Checks/Plugin_Repo/Write_File_Check.php`:
- Added proper description
- Added documentation URL
- Integrates the sniff into Plugin Check

### 3. Test Coverage
**Test Plugins:**
- `test-plugin-write-file-with-errors/`: Contains 7 examples of incorrect usage
- `test-plugin-write-file-without-errors/`: Contains examples of correct usage

**PHPUnit Tests:**
- `Write_File_Check_Tests.php`: Tests the check class
- `WriteFileUnitTest.php` and `WriteFileUnitTest.inc`: Tests the sniff directly

### 4. Configuration
- Added `PluginCheck.CodeAnalysis.WriteFile` rule to `phpcs-sniffs/PluginCheck/ruleset.xml`

## Limitations
The sniff uses static analysis and can only detect file write operations where the path is directly specified in the function call. It cannot detect:
- Paths stored in variables before the function call
- Dynamically constructed paths using complex logic
- File writes through wrapper functions or classes

This limitation is acceptable as it catches the most common cases.

## Related Resources
- Original guideline text in reviews: https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
- Settings API: https://developer.wordpress.org/plugins/settings/
- wp_upload_dir(): https://developer.wordpress.org/reference/functions/wp_upload_dir/
- wp_handle_upload(): https://developer.wordpress.org/reference/functions/wp_handle_upload/

## Acknowledgments

This check is based on the `calls_write_file_warning` detection logic from the internal plugin review scanner developed by @frantorres, which has been successfully identifying these issues during manual reviews. This implementation makes that same detection available to plugin developers as an automated check.


includes/Checker/Checks/Plugin_Repo/Write_File_Check.php
```diff
@@ -0,0 +1,88 @@
...
+use WordPress\Plugin_Check\Checker\Check_Categories;
+use WordPress\Plugin_Check\Checker\Check_Result;
+use WordPress\Plugin_Check\Checker\Checks\Abstract_PHP_CodeSniffer_Check;
+use WordPress\Plugin_Check\Traits\Amend_Check_Result;
+use WordPress\Plugin_Check\Traits\Stable_Check;
+
+/**
+ * Check to detect loading files from external sites.
+ *
+ * @since n.e.x.t.
```
ernilambar:
```suggestion
 * @since 2.0.0
```

includes/Checker/Checks/Plugin_Repo/Write_File_Check.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @return array The categories for the check.
+	 */
+	public function get_categories() {
+		return array( Check_Categories::CATEGORY_PLUGIN_REPO );
+	}
+
+	/**
+	 * Returns an associative array of arguments to pass to PHPCS.
+	 *
+	 * @since 1.0.0
```
ernilambar:
```suggestion
	 * @since 2.0.0
```

phpcs-sniffs/PluginCheck/Sniffs/CodeAnalysis/WriteFileSniff.php
```diff
@@ -0,0 +1,326 @@
...
+
+/**
+ * Verifies plugins don't save data in the plugin folder.
+ *
+ * Plugin folders are deleted when upgraded, so using them to store any data is problematic.
+ * Plugins should use the uploads directory or the database instead.
+ *
+ * @link https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
+ *
+ * @since 1.1.0
```
ernilambar:
```suggestion
 * @since 2.0.0
```

ernilambar (approved):

frantorres (approved):

PR #1147, opened 2026-01-03
Improve `direct_file_access` check to ignore class-only files

### LINKED ISSUE #1146: Improve `direct_file_access` check to ignore class-only files
The `direct_file_access` check currently flags PHP files that do not include an explicit protection against direct access (e.g. `defined( 'ABSPATH' ) || exit;`).

However, this behaviour can produce **false positives** for files that:

- Contain **only class definitions**
- Do not execute any logic on load
- Are intended to be included or autoloaded, not accessed directly

### Current Behaviour

Files that only declare classes (no side effects, no executable code) are still reported by the `direct_file_access` check if they lack an `ABSPATH` guard.

Example:

```php
<?php

class My_Plugin_Service {
    public function do_something() {
        // ...
    }
}
````

This file is flagged, even though direct access would not produce output or cause unintended behavior.

### Expected Behavior

The `direct_file_access` check should **skip or ignore PHP files that only contain class declarations**, provided that:

* No code is executed at the top level
* No functions are called outside class/method scope
* No side effects (echo, include, require, hooks, etc.) are present

### Proposed Improvement

Enhance the check logic to detect whether a file contains **only structural code**, such as:

* `class`, `interface`, or `trait` declarations
* Namespaces and `use` statements
* PHPDoc blocks

If so, the file should not require a direct access guard and should not be reported.

### Benefits

* Reduces false positives in modern, OOP-based plugins
* Aligns better with PSR-style and autoloaded architectures
* Improves developer experience and signal-to-noise ratio of Plugin Check

### Additional Notes

This could potentially be implemented using AST parsing (e.g. via `nikic/php-parser`) or by enhancing the existing token-based analysis to detect top-level executable statements.

### PR DESCRIPTION
**Pull request by davidperezgar**
Fixes #1146

## Description

This PR improves the `direct_file_access` check to use AST parsing with `nikic/php-parser` to accurately detect files that only contain structural code (classes, interfaces, traits, namespaces) and skip them from requiring direct access guards. This reduces false positives for modern, OOP-based plugins that follow PSR-style and autoloaded architectures.

The implementation uses Abstract Syntax Tree (AST) parsing to detect whether a file contains only structural code without executable statements, making the check more accurate than the previous regex-based approach.

## Changes

- **Implemented AST parsing** using `nikic/php-parser` to analyze PHP files for structural code only
- **Removed deprecated imports** - Replaced `PhpParser\Node\Scalar\*` class imports with string class name comparisons to avoid deprecation warnings
- **Enhanced detection logic**:
  - Files with only class/interface/trait/enum definitions are now correctly identified as safe
  - Files with namespace declarations and use statements are handled recursively
  - Asset files (files with only assignments and return statements) are properly allowed
  - Safe if statements (with `class_exists`, `function_exists`, etc.) are now recognized as safe
  - Top-level function declarations are correctly flagged as unsafe (procedural code)
- **Refactored code complexity** - Split `is_safe_expression()` into smaller, focused methods to reduce cyclomatic complexity from 28 to below threshold
- **Added PHPStan ignore rules** for property access on dynamically typed PhpParser nodes
- **Maintained backward compatibility** - Falls back to regex-based checking if AST parsing fails

## Credits

This implementation is based on the AST parsing approach used in the internal plugin review scanner, originally developed by @frantorres. The logic for detecting safe structural code has been adapted and integrated into the public Plugin Check tool.

## Benefits

- **Reduces false positives** - Class-only files no longer trigger unnecessary warnings
- **Better developer experience** - Improves signal-to-noise ratio of Plugin Check
- **More accurate detection** - AST parsing provides precise analysis of file structure
- **Aligns with modern practices** - Better support for PSR-style and autoloaded plugin architectures
- **Maintains security** - Still correctly flags files with procedural code that need guards

## Testing Instructions

1. **Test class-only files**:
   - Create a PHP file with only a class definition (no ABSPATH guard)
   - Run Plugin Check on the plugin
   - Expected: No error should be reported for the class-only file

2. **Test interface/trait-only files**:
   - Create files with only interface or trait definitions
   - Run Plugin Check
   - Expected: No errors for these files

3. **Test procedural code**:
   - Create a PHP file with a top-level function declaration (no ABSPATH guard)
   - Run Plugin Check
   - Expected: Error should be reported requiring ABSPATH guard

4. **Test asset files**:
   - Create a file with only variable assignments and return statements (like build asset files)
   - Run Plugin Check
   - Expected: No error should be reported

5. **Test safe function calls**:
   - Create a file with if statements checking `class_exists()` or `function_exists()` with return statements
   - Run Plugin Check
   - Expected: No error should be reported

6. **Run PHPUnit tests**:
   ```bash
   npm run test-php -- --filter Direct_File_Access_Check_Tests
   ```
   - Expected: All tests should pass

## Checklist

- [x] Code follows WordPress Coding Standards
- [x] Self-reviewed the code
- [x] Added necessary comments
- [x] No new linter errors (PHPStan ignores added for expected dynamic property access)
- [x] PHPMD complexity issues resolved
- [x] Tests updated/verified to work with new implementation
- [x] Backward compatibility maintained with regex fallback

### CODE REVIEW

**Review by ernilambar — APPROVED** [review]

File: includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php [hunk 1]
```diff
@@ -184,6 +365,462 @@ private function is_valid_for_direct_access( $file ) {
...
+	 *
+	 * @since 1.9.0
+	 *
+	 * @param array $ast The parsed AST nodes.
+	 * @return bool True if the AST only contains structural code, false otherwise.
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
+	 */
+	private function is_ast_valid_for_direct_access( array $ast ) {
+		$safe_node_types = array(
+			Stmt\Nop::class,
```
marekdedic, line 392: I'm not that familiar with the AST, but probably there should be `Stmt\Declare_::class` as well - if I'm reading it correctly, this would also allow

```php
declare(strict_types=1);
```

marekdedic:
Hi, is there any way we can help move this forward?

davidperezgar:
We're a little busy in Plugins team. You could propose code to this branch if you want, if not, I'll check next week.

frantorres (approved):

includes/Checker/Checks/Plugin_Repo/Direct_File_Access_Check.php
davidperezgar: ok thanks

PR #1193, opened 2026-02-27
Fix fatal on WP 7.0 by deferring to core AI client

davidperezgar:
**Summary**
- Prevents class/interface signature conflicts between the plugin’s bundled AI client and WordPress 7.0 core AI client.
- After loading the plugin’s Composer autoloader, if core AI client is present, unregisters `WordPress\AiClient\` and `WordPress\AI_Client\` PSR-4 mappings so core classes are used.
- Adds a small helper to detect the core AI client bootstrap.

**Testing**
- Local environment

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -41,7 +41,13 @@ function wp_plugin_check_load() {
 	}
 
 	// Load the Composer autoloader.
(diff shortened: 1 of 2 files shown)


ernilambar:
@davidperezgar Tests are still failing.

davidperezgar:
Yes, but I didn't manage to reproduce the WP 7 error. Could you check it?

PR #1213, opened 2026-03-12
Fix: AI Check model preference not being applied

### LINKED ISSUE #1212: Plugin Check Namer: Selected AI model is ignored — backend always uses Anthropic regardless of chosen model
## Description

When using the **Plugin Check Namer Tool** (`/wp-admin/tools.php?page=plugin-check-namer`) 
with **WordPress 7.0** and the **trunk branch** of plugin-check, the **AI model dropdown 
selection is completely ignored**. Even when a GPT (OpenAI) model is explicitly selected, 
the backend always routes the request to the **Anthropic API**, regardless of the user's 
choice.

## Steps to Reproduce

1. Install WordPress 7.0.
2. Use the trunk branch of the plugin-check plugin.
3. Navigate to **Settings → Connectors** and connect all three providers — **Anthropic**, 
   **Google**, and **OpenAI** — by entering a valid API key for each.
4. Navigate to **Tools → Plugin Check Namer**.
5. Enter a plugin name (e.g. `WordPress Plugin Test`) and an author name (e.g. `WordPress`).
6. Open the **AI model** dropdown — it lists models from all three connected providers 
   (Claude, Gemini, GPT, etc.).
7. Select any **OpenAI GPT model** (e.g. `gpt-5.2`) from the dropdown.
8. Click **Evaluate name**.

## Expected Behaviour

The AJAX request should be sent using the selected model (`gpt-5.2` via OpenAI), and the 
OpenAI API should be called accordingly.

## Actual Behaviour

Despite selecting a GPT model, the backend ignores the selection and routes the request to 
the **Anthropic API**. This is confirmed via the browser DevTools Network tab, where the 
AJAX response returns:
```json
{
  "success": false,
  "data": {
    "message": "Bad Request (400) – Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits."
  }
}
```

The error message clearly shows Anthropic was called even though an OpenAI model was 
selected. The model selection dropdown has no effect on which provider is used.

## Screenshots / Evidence

- **Connectors screen**: All three providers (Anthropic, Google, OpenAI) are connected.
- **Plugin Check Namer**: `gpt-5.2` is selected in the AI model dropdown.
- **DevTools Network → Preview**: Response shows an Anthropic-specific error, confirming 
  the wrong provider was called.

https://github.com/user-attachments/assets/db6852e1-490d-4307-8975-e34b81ad6b0b

## Environment

- **WordPress version:** 7.0 (beta5)
- **Plugin-check branch:** trunk
- **Connected providers:** Anthropic, Google, OpenAI
- **Selected model:** `gpt-5.2` (OpenAI)
- **Browser:** Chrome

## Suggested Fix

The selected model value from the dropdown should be passed correctly in the AJAX request 
payload, and the backend handler should use it to resolve the appropriate AI provider 
instead of defaulting to Anthropic.

raftaar1191:
The model preference was not being applied due to two issues:

1. method_exists() check was failing because WP_AI_Client_Prompt_Builder
   uses magic __call() method. Changed to directly call the method with
   try-catch error handling instead.

2. normalize_model_preference() was returning a nested array
   [['provider', 'model']] instead of the expected ['provider', 'model']
   format. Removed the extra array wrapper.

This ensures that when users specify a model preference like
'openai::gpt-5.4-pro', it is correctly passed to the WordPress AI
client instead of falling back to the default model.

Fixes #1212

## Issue Video:

https://github.com/user-attachments/assets/5cc81dc1-3ab8-498e-928a-7a3b43a68bce

## Video after adding the patch

https://github.com/user-attachments/assets/6c874309-9d1e-4a94-85a1-2afd6212759c

### DIFF
--- a/includes/Traits/AI_Check_Names.php
+++ b/includes/Traits/AI_Check_Names.php
@@ -210,25 +210,22 @@ trait AI_Check_Names {
 			return $builder;
 		}
 
-		$method = null;
-		if ( method_exists( $builder, 'using_model_preference' ) ) {
-			$method = 'using_model_preference';
-		} elseif ( method_exists( $builder, 'usingModelPreference' ) ) {
-			$method = 'usingModelPreference';
-		}
-
-		if ( ! $method ) {
-			return $builder;
-		}
-
 		$preference = $this->normalize_model_preference( $model_preference );
-		$result     = call_user_func( array( $builder, $method ), $preference );
 
-		if ( is_wp_error( $result ) ) {
-			return $result;
+		try {
+			$result = $builder->using_model_preference( $preference );
+			return $result ? $result : $builder;
+		} catch ( \Exception $e ) {
+			// If method doesn't exist or fails, return WP_Error.
+			return new WP_Error(
+				'model_preference_error',
+				sprintf(
+					/* translators: %s: Exception message */
+					__( 'Failed to apply model preference: %s', 'plugin-check' ),
+					$e->getMessage()
+				)
+			);
 		}
-
-		return $result ? $result : $builder;
 	}
 
 	/**
@@ -249,7 +246,7 @@ trait AI_Check_Names {
 			if ( false !== strpos( $trimmed, $separator ) ) {
 				list( $provider, $model ) = array_map( 'trim', explode( $separator, $trimmed, 2 ) );
 				if ( '' !== $provider && '' !== $model ) {
-					return array( array( $provider, $model ) );
+					return array( $provider, $model );
 				}
 			}
 		}


davidperezgar:
Hello, You need to fix the tests before we can review.

raftaar1191:
On it @davidperezgar

raftaar1191:
Hey @davidperezgar, tests are all passing now — the PR is ready for review whenever you get a chance!

ernilambar (approved):

davidperezgar (approved):

PR #1225, opened 2026-03-17
Fix token usage not displayed in Namer result

### LINKED ISSUE #1220: Token usage not displayed in Plugin Check Namer result
The **Plugin Check Namer Tool** does not correctly display the total number of tokens used during the AI evaluation process. Instead of showing the actual token count, the UI currently displays **“undefined total”**.

This prevents users from understanding the token consumption for the AI request.

**Page where the issue occurs**

`/wp-admin/tools.php?page=plugin-check-namer`

**Current Behavior**

After running an evaluation, the result section displays:

```
Tokens used: openai:gpt-4.1 - undefined total
```

The total token count is not populated even though the request completes successfully.

**Expected Behavior**

The result section should display the **actual total number of tokens used** for the request.

Example:

```
Tokens used: openai:gpt-4.1 - 1,234 total
```

or ideally a breakdown:

```
Tokens used:
Prompt: 850
Completion: 384
Total: 1,234
```

**Steps to Reproduce**

1. Go to **WP Admin → Tools → Plugin Check Namer**.
2. Enter a plugin name.
3. Click **Evaluate name**.
4. Wait for the result to be generated.
5. Check the **Tokens used** line in the result section.

**Observed Result**

The UI displays:

```
Tokens used: openai:gpt-4.1 - undefined total
```

**Expected Result**

The UI should display the **actual token count returned by the AI provider**.

**Screenshot**

Image

The issue can be observed in the result section where token usage is reported as **undefined** instead of a numeric value.

aslamdoctor:
## Summary

Fixes #1220

- **Root cause:** `WP_AI_Client_Prompt_Builder` provides methods like `generate_text_result()` via PHP's `__call()` magic method. The code used `method_exists()` which returns `false` for magic methods, so `generate_text_result()` was never called. Instead, it always fell through to `generate_text()` which returns a plain string with no token usage data — resulting in "undefined total" in the UI.
- **Fix:** Replace `method_exists()` with `is_callable()` in `execute_ai_request()` so the rich result path is used and token usage is returned.
- **Fallback:** Compute `total_tokens` from `prompt_tokens + completion_tokens` in PHP if not directly available, and add defensive JS handling to prevent "undefined total" display.

### Screenshot
image

## Test plan

- [ ] Go to **WP Admin → Tools → Plugin Check Namer**
- [ ] Enter a plugin name and click **Evaluate name**
- [ ] Verify the **Tokens used** line displays actual numeric values (e.g. `1,234 total (850 prompt + 384 completion)`)
- [ ] Verify the similar name query token count is also displayed when available

### DIFF
--- a/assets/js/plugin-check-namer.js
+++ b/assets/js/plugin-check-namer.js
@@ -268,7 +268,17 @@
 							}
 						}
 
-						tokensText += tokenUsage.total_tokens + ' total';
+						const totalTokens =
+							tokenUsage.total_tokens ||
+							( tokenUsage.prompt_tokens &&
+							tokenUsage.completion_tokens
+								? tokenUsage.prompt_tokens +
+								  tokenUsage.completion_tokens
+								: null );
+
+						if ( totalTokens ) {
+							tokensText += totalTokens + ' total';
+						}
 
 						// Add breakdown with prompt and completion tokens.
 						if (
--- a/includes/Traits/AI_Check_Names.php
+++ b/includes/Traits/AI_Check_Names.php
@@ -163,10 +163,12 @@ trait AI_Check_Names {
 		}
 
 		// Try to generate a rich result first.
+		// Use is_callable() instead of method_exists() to detect methods
+		// provided via __call() magic (e.g. WP_AI_Client_Prompt_Builder).
 		$result = null;
-		if ( method_exists( $builder, 'generate_text_result' ) ) {
+		if ( is_callable( array( $builder, 'generate_text_result' ) ) ) {
 			$result = $builder->generate_text_result();
-		} elseif ( method_exists( $builder, 'generateTextResult' ) ) {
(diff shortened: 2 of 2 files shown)


raftaar1191:
@aslamdoctor 

here with the token I am also see the model name. I think here should be one more section call as **Model used**



Now:
image



How I guess it should be:
image

@davidperezgar what do you think?

davidperezgar:
Look nice!

aslamdoctor:
@raftaar1191 should we show model name only when the "AI Model" dropdown is set to "Automatic" or always?

raftaar1191:
@aslamdoctor I think **always** will be the best options.

@davidperezgar Any comment of your on this?

davidperezgar:
@aslamdoctor that would be another PR. 

LGTM

davidperezgar (approved):

ernilambar (approved):

PR #1230, opened 2026-03-20
Fix: non-line specific issues should output line 1 instead of 0

### LINKED ISSUE #983: Non-line specific issues should output line 1 instead of 0
Some errors are reported at line 0 - unfortunately this means that some post-processing tools, will skip those errors or report an error/not work as expected, since lines generally start at 1
(besides the fact that some IDE file protocols won't work with line=0 either)

I understand why you chose line 0 and it makes absolute sense - it's just that it causes some compatibility issues.

meravi:
## Summary

Non-line specific issues were outputting line 0, which causes 
compatibility issues with post-processing tools and IDEs since 
lines start at 1, not 0.

## Changes

- `includes/Checker/Check_Result.php` — force minimum line value 
   to 1 when storing errors/warnings in the array
- `includes/Traits/Amend_Check_Result.php` — updated default 
   `$line` parameter from 0 to 1 in all three methods
- `includes/Utilities/Results_Exporter.php` — updated normalize_entry
- `tests/behat/features/plugin-check.feature` — updated test 
   assertions to expect line 1

Fixes #983

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1230-f7c13b34b1efe97383f6ec55ceef758893c33bba.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


meravi:
Hi @swissspidy 👋 This is a small bug fix for #983. 
 Would appreciate a review when you get a chance! Thanks 🙏

davidperezgar:
@meravi Could you check the tests?

davidperezgar:
@meravi tests are still pending...

davidperezgar:
Thanks for tackling #983! I have a concern about the approach here.

The PR changes the **default value** of `line` from `0` to `1` in `Check_Result::add_message()` and throughout `Amend_Check_Result`. The problem is that `line` is used as an **internal indexing key**:

```php
$this->errors[ $file ][ $line ][ $column ][] = $data;
```

This means an error with a genuinely known line 1 and an error with an *unknown* line now collide under the same key (`[1]`), whereas before, `0` acted as an unambiguous "unknown" bucket. This isn't just a test-breaking change — it introduces real ambiguity into the data model that didn't exist before.

This is also why CI is failing: 42 failures / 31 errors across many test files (`Direct_File_Access_Check_Tests`, `File_Type_Check_Tests`, `Plugin_Readme_Check_Tests`, `Trademarks_Check_Tests`, `Plugin_Updater_Check_Tests`, etc.), all hitting hardcoded `$errors['file.php'][0][...]` accesses that no longer resolve since the bucket moved from `0` to `1`.

The original issue (#983) is purely cosmetic — not showing "line 0" to users. I'd suggest fixing this only at the **presentation layer** (CLI output / Admin UI results table), e.g. rendering `$line > 0 ? $line : 'N/A'` when displaying, instead of changing the default in the core data model. That avoids:
- Breaking the existing test suite.
- Changing the semantics of "unknown line" across the whole check engine.
- Risking side effects in any other consumer of `get_errors()`/`get_warnings()` that relies on key `0`.

Happy to help narrow down where the CLI/Admin rendering happens if that's useful.

PR #1281, opened 2026-04-27
Use latest stable WP version for wp-since generation

ernilambar:
**What**

The "Generate WP Functions Since Data" workflow was checking out `WordPress/WordPress` with no `ref`, so it used the default branch instead of a **stable** release.

**Change**

- Resolve the current stable version from the WordPress.org version-check API: `https://api.wordpress.org/core/version-check/1.7/`
- Check out the matching release tag for that version so the generated `@since` dataset reflects **latest stable** core.

### DIFF
--- a/.github/workflows/generate-wp-functions-since-data.yml
+++ b/.github/workflows/generate-wp-functions-since-data.yml
@@ -21,6 +21,10 @@ jobs:
         id: wp
         run: |
           VERSION=$(curl -fsSL 'https://api.wordpress.org/core/version-check/1.7/' | jq -r '.offers[0].version')
+          if [ -z "$VERSION" ] || [ "$VERSION" = "null" ]; then
+            echo "::error::Failed to resolve WordPress version from API"
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-wp-functions-since-data.yml
```diff
@@ -17,10 +17,18 @@ jobs:
       - name: Checkout plugin-check
         uses: actions/checkout@v6
 
+      - name: Resolve latest stable WordPress version
+        id: wp
+        run: |
+          VERSION=$(curl -fsSL 'https://api.wordpress.org/core/version-check/1.7/' | jq -r '.offers[0].version')
```
mukeshpanchal27, line 23: Let's add a check for `VERSION`, if API down then it will return empty or null for `VERSION`.

Note tested but something like:
```
if [ -z "$VERSION" ] || [ "$VERSION" = "null" ]; then
    echo "::error::Failed to resolve WordPress version from API"
    exit 1
fi
```

mukeshpanchal27 (approved):

davidperezgar (approved):

PR #1291, opened 2026-05-04
feat(plugin-repo): Add privacy policy content check

### LINKED ISSUE #1249: Privacy: Check that plugins handling personal data call wp_add_privacy_policy_content()
### Summary
The WordPress Plugin Handbook requires that every plugin which collects, uses, stores, or passes personal data to a third party should suggest privacy policy text to site administrators using `wp_add_privacy_policy_content()`. See: https://developer.wordpress.org/plugins/privacy/suggesting-text-for-the-site-privacy-policy/

In practice the vast majority of plugins that handle personal data do not implement this, leaving site administrators without the information they need to write an accurate privacy policy.

### What to check
Plugin Check should detect when a plugin appears to handle personal data — indicated by usage of functions or patterns such as:

- `wp_remote_post()` / `wp_remote_get()` sending data to external URLs
- `setcookie()` or `$_COOKIE`
- `$_POST` / `$_GET` handling personal fields (email, name etc.)
- Loading external scripts or pixels (see #1156 / #475)

...and warn if `wp_add_privacy_policy_content()` is not called anywhere in the plugin.

faisalahammad:
## Summary

Implements the feature requested in #1249.

WordPress.org guidelines require that every plugin which collects, uses, stores, or passes personal data to a third party must suggest privacy policy text to site administrators using `wp_add_privacy_policy_content()`. In practice, the vast majority of plugins that handle personal data do not implement this.

This PR adds a new `Privacy_Policy_Check` (static file check) that detects common personal-data-handling signals and warns if `wp_add_privacy_policy_content()` is absent.

## How it works

The check scans all PHP files in the plugin for signals that indicate potential personal data handling:

| Signal | Rationale |
|--------|-----------|
| `wp_remote_post()` | Transmitting data to external services |
| `wp_remote_get()` | Fetching from external URLs (analytics, tracking) |
| `setcookie()` | Setting cookies that may track users |
| `$_COOKIE` | Reading cookie data |
| `wp_set_auth_cookie()` | Authentication cookie handling |

If any signal is found **and** `wp_add_privacy_policy_content()` is not called anywhere in the plugin, a **single warning** (severity 5) is emitted on the plugin's main file with a link to the [WordPress privacy developer docs](https://developer.wordpress.org/plugins/privacy/suggesting-text-for-the-site-privacy-policy/).

Plugins with no signals are completely unaffected — the check stays silent.

## Design decisions

- **Warning, not error** — the presence of these API calls doesn't guarantee personal data is being mishandled; it's a strong enough signal to inform authors but not block submission
- **One warning per plugin** — avoids noise by stopping after the first matched signal
- **Early exit** — checks for `wp_add_privacy_policy_content()` first; if present, skips all signal scanning
- **Conservative signals** — only high-confidence indicators (remote posts/gets, cookies) are included, not noisy ones like raw `$_POST`/`$_GET`

## Files changed

- `includes/Checker/Checks/Plugin_Repo/Privacy_Policy_Check.php` — new check class
- `includes/Checker/Default_Check_Repository.php` — register as `privacy_policy`
- `docs/checks.md` — add row to the checks table
- `tests/phpunit/tests/Checker/Checks/Privacy_Policy_Check_Tests.php` — 3 test cases
- `tests/phpunit/testdata/plugins/test-plugin-privacy-policy-with-errors/` — has signal, no privacy call → triggers warning
- `tests/phpunit/testdata/plugins/test-plugin-privacy-policy-without-errors/` — has signal + `wp_add_privacy_policy_content()` → clean
- `tests/phpunit/testdata/plugins/test-plugin-privacy-policy-no-signals/` — no signals → clean

## Testing

Manually verified against **Mailchimp for WordPress** (mc4wp) — correctly fires the `missing_privacy_policy_content` warning on `mailchimp-for-wp.php`.

Verified against **Akismet** (which properly calls `wp_add_privacy_policy_content()`) — no warning produced.

Closes #1249

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1291-d025f52cf90a19e18934ca4edf0dee82b9e62b80.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -35,3 +35,4 @@
 | enqueued_styles_scope | performance | Checks whether any stylesheets are loaded on all pages, which is usually not desirable and can lead to performance issues. | [Learn more](https://developer.wordpress.org/plugins/) |
 | enqueued_scripts_scope | performance | Checks whether any scripts are loaded on all pages, which is usually not desirable and can lead to performance issues. | [Learn more](https://developer.wordpress.org/plugins/) |
 | non_blocking_scripts | performance | Checks whether scripts and styles are enqueued using a recommended loading strategy. | [Learn more](https://developer.wordpress.org/plugins/) |
+| privacy_policy | plugin_repo | Checks that plugins handling personal data call wp_add_privacy_policy_content() to suggest privacy policy text to site administrators. | [Learn more](https://developer.wordpress.org/plugins/privacy/suggesting-text-for-the-site-privacy-policy/) |
--- a/includes/Checker/Checks/Plugin_Repo/Privacy_Policy_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Privacy_Policy_Check.php
@@ -130,6 +130,8 @@ class Privacy_Policy_Check extends Abstract_File_Check {
 	 * @param array                 $files           List of absolute file paths.
 	 * @param array<string, string> $function_names Map of function names (lowercase) to human-readable labels.
 	 * @return string|false Human-readable label of the first matched function, or false if none found.
+	 *
+	 * @SuppressWarnings(PHPMD.NPathComplexity)
 	 */
 	private function find_function_in_files( array $files, array $function_names ) {
 		$lowercase_names = array();
--- a/includes/Checker/Checks/Plugin_Repo/Privacy_Policy_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Privacy_Policy_Check.php
@@ -21,6 +21,9 @@ use WordPress\Plugin_Check\Traits\Stable_Check;
  * administrators via wp_add_privacy_policy_content(). This check detects common
  * personal-data-handling patterns and warns if that function is not used.
  *
+ * Uses token-based parsing to avoid false positives from function names appearing
+ * in comments or string literals.
+ *
  * @since 1.7.0
  */
 class Privacy_Policy_Check extends Abstract_File_Check {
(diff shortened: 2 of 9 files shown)


davidperezgar:
Thanks for the PR! I'm reviewing.

davidperezgar:
Thanks for the PR! The intent makes sense, but I think this needs one more pass before merging.

The check currently scans raw PHP file contents with regexes, so it matches function names in comments and strings as if they were real code. That creates both false negatives and false positives:

- A comment like `// TODO: call wp_add_privacy_policy_content() later` will make the check return early, even if the plugin never actually calls the function.
- A comment or example string containing `wp_remote_post()` will trigger the warning, even if the plugin does not actually make a remote request.

This also affects the current “without errors” fixture, because the plugin header description contains `wp_add_privacy_policy_content()` before the real call is reached.

Could we update the detection to ignore comments and strings, for example by using `token_get_all()` and only matching actual function calls / variable usage in code tokens? It would also be good to add explicit tests for function names appearing only in comments or strings, both for `wp_add_privacy_policy_content()` and for the signal patterns.

faisalahammad:
Thanks for the review. Good catch on the regex.

I moved the detection from regex to token parsing with `token_get_all()`. Now it only looks at real PHP tokens, so function names or variables that show up inside comments or string literals are not matched anymore. This fixes both the false positives (a comment mentioning the function would mark it as present and hide the warning) and keeps real calls detected.

What changed:

- Function calls are matched on `T_STRING` followed by `(`, with a guard so method calls and definitions (`->`, `::`, `function`, `new`) are skipped. `?->` is handled too, guarded behind `defined()` so it stays PHP 7.4 safe.
- `$_COOKIE` is matched as a `T_VARIABLE` token, not text.
- Added two fixtures: one with the privacy function name only inside a comment, one only inside a string. Both also have a real signal, so the test proves the warning still fires while the comment/string mention is ignored.

All checks pass locally: PHPUnit, PHPCS (WordPress + PHPCompatibility), and PHPStan.

Pushed in d025f52.

masteradhoc:
Hey @davidperezgar did you had the chance to recheck the PR once again? :)

davidperezgar:
You need to validate the tests before we can review.

PR #1293, opened 2026-05-04
feat(privacy): add check for missing wp_privacy_personal_data_erasers registration

### LINKED ISSUE #1252: Privacy: Add check for wp_privacy_personal_data_erasers filter (GDPR personal data erasure)
## Summary

Plugin Check should detect when a plugin collects personal data but does not register an eraser callback via the `wp_privacy_personal_data_erasers` filter.

## Background

Since WordPress 4.9.6, WordPress ships a Personal Data Removal tool to help sites comply with privacy regulations such as the GDPR. Plugins that store personal data (e.g. in comment meta, post meta, or custom post types) are expected to hook into this tool by registering an eraser callback via:

add_filter( 'wp_privacy_personal_data_erasers', 'my_plugin_register_erasers' );

The eraser callback receives a user's email address and a pagination page number, removes the associated personal data, and returns a status array indicating whether items were removed, retained, and whether processing is complete.

Reference: https://developer.wordpress.org/plugins/privacy/adding-the-personal-data-eraser-to-your-plugin/

## Proposed Check

Add a detection rule that warns plugin authors when their plugin appears to handle personal data (e.g. uses `add_comment_meta`, `update_user_meta`, custom DB writes, etc.) but does not register a handler for the `wp_privacy_personal_data_erasers` filter.

faisalahammad:
## Summary

Adds a new static check (`Personal_Data_Eraser_Check`) that warns plugin authors when their plugin appears to handle personal data — via `add_user_meta`, `update_user_meta`, `add_comment_meta`, `update_comment_meta`, or direct `$wpdb` writes — but does not register a callback via the `wp_privacy_personal_data_erasers` filter.

Since WordPress 4.9.6, the Personal Data Removal tool lets site administrators honor GDPR erasure requests. Plugins storing personal data are expected to hook into this tool.

Fixes #1252

### `includes/Checker/Checks/Plugin_Repo/Personal_Data_Eraser_Check.php` _(new)_

Two-step static check: first confirm the plugin has at least one personal-data storage call, then verify whether it registers the `wp_privacy_personal_data_erasers` filter. A warning is only emitted when step 1 matches and step 2 does not — avoiding false positives for plugins that never touch personal data.

Detection uses a `token_get_all()` token scanner, porting the approach from #1292. It matches real function and `$wpdb` method calls, ignores comments and string literals, and skips files under a top-level `tests/` directory.

**Key constants:**
```php
const PERSONAL_DATA_FUNCTIONS = array( 'add_user_meta', 'update_user_meta', 'add_comment_meta', 'update_comment_meta' );
const WPDB_METHODS            = array( 'insert', 'update', 'replace' );
const ERASER_FILTER           = 'wp_privacy_personal_data_erasers';
```

The check is **experimental** and only runs with `--include-experimental`.

Warning code: `missing_personal_data_eraser`  
Docs link: https://developer.wordpress.org/plugins/privacy/adding-the-personal-data-eraser-to-your-plugin/

### `includes/Checker/Default_Check_Repository.php`

```php
// Added:
'personal_data_eraser' => new Checks\Plugin_Repo\Personal_Data_Eraser_Check(),
```

## Testing

**Test 1: Plugin stores user meta, no eraser registered → should warn**
1. Activate a plugin that calls `update_user_meta()` without hooking `wp_privacy_personal_data_erasers`
2. Run Plugin Check against it (Admin UI or WP-CLI)
3. Result: warning `missing_personal_data_eraser` is reported ✅

**Test 2: Plugin stores user meta and registers an eraser → should pass**
1. Activate a plugin that calls `update_user_meta()` and registers via `add_filter( 'wp_privacy_personal_data_erasers', ... )`
2. Run Plugin Check against it
3. Result: no `missing_personal_data_eraser` warning ✅

**Test 3: Plugin has no personal data handling → should pass**
1. Run Plugin Check against any plugin that does not call user/comment meta functions or `$wpdb` writes
2. Result: no `missing_personal_data_eraser` warning ✅

**Test 4: Direct `$wpdb` write → should warn**
1. Activate a plugin that calls `$wpdb->insert()` without registering an eraser
2. Run Plugin Check with `--include-experimental`
3. Result: warning `missing_personal_data_eraser` is reported ✅

**Test 5: Eraser registered only in a `tests/` directory → should warn**
1. Activate a plugin that calls `update_user_meta()` but only registers the eraser inside its own `tests/` folder
2. Run Plugin Check with `--include-experimental`
3. Result: warning `missing_personal_data_eraser` is reported ✅

PHPUnit test class: `tests/phpunit/tests/Checker/Checks/Personal_Data_Eraser_Check_Tests.php` covers all cases with dedicated test data plugins.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1293-84f04de4387967527e17b2a3cb4d6aeb6c44a8ca.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dknauss:
This PR still carries a bug that #1292 already fixed: `PERSONAL_DATA_PATTERN` never matches `$wpdb` writes.

```php
'/\b(?:add_user_meta|update_user_meta|add_comment_meta|update_comment_meta|\$wpdb\s*->\s*(?:insert|update|replace))\s*\(/'
```

The `\b` sits before the alternation group, so it applies to the `\$wpdb` branch too. `$` is a non-word character, so the boundary only succeeds when `$wpdb` is preceded by a word character — which never happens in real code:

| Input | Result |
| --- | --- |
| `\t$wpdb->insert( $table, $data );` | no match |
| `    $wpdb->insert( $table, $data );` | no match |
| `$wpdb->update( $t, $d, $w );` | no match |
| `$ok = $wpdb->replace( $t, $d );` | no match |
| `foo$wpdb->insert( $t, $d );` | match |

So direct DB writes — the case #1252 explicitly calls out — are never detected. The PHPUnit fixtures here don't catch it because neither test plugin exercises a `$wpdb` write. @AndriusBurba caught the identical bug in #1292 on 4 June, and it was fixed there by replacing the regex with a `token_get_all()` scanner.

A few other things #1292 resolved that apply equally here, since the two checks share a detection strategy:

- **Comments and string literals count as signals.** Matching raw file contents means `// TODO: update_user_meta( ... )` registers as personal-data handling. The same cuts the other way for `ERASER_REGISTRATION_PATTERN`: a plugin that merely mentions `wp_privacy_personal_data_erasers` in a comment will have the warning suppressed.
- **No `tests/` exclusion**, so a plugin's own test fixtures trip the check. #1292 added a path-anchored one.
- **Not gated behind `--include-experimental`**, which #1292 adopted for exactly the false-positive noise that applies here.

Given how closely the two checks mirror each other, it may be simplest to land #1292 first and port its scanner across rather than fix these four things independently.

faisalahammad:
Thanks for the review. I ported the `token_get_all()` scanner from #1292 and fixed all four points.

- `$wpdb` writes now match. The check scans real tokens, so `$wpdb->insert()`, `$wpdb->update()`, and `$wpdb->replace()` are detected. There is no regex left.
- Comments and string literals no longer count as signals. The scanner only looks at real function calls and `$wpdb` method calls.
- The check now skips files under a top-level `tests/` directory.
- The check is now experimental and only runs with `--include-experimental`.

Added four test fixtures that cover each case. The full PHPUnit suite passes (479 tests). Please re-review.

PR #1369, opened 2026-06-20
Fix false positive for files with only require_once and class definitions

### LINKED ISSUE #1361: False positive `missing_direct_file_access_protection` for class files containing `require_once`
## Description

PHP files that contain only a class definition and `require_once` statements (to load parent/dependency classes) are incorrectly flagged with `missing_direct_file_access_protection`. These files have no top-level executable code and are genuinely safe for direct access, so the error is a false positive.

## Steps to reproduce

Create a PHP file in a plugin with only a `require_once` statement and a class definition:

```php
<?php
require_once __DIR__ . '/class-parent.php';

class My_Class extends Parent_Class {
    // ...
}
```

Run `wp plugin check <slug>` — the file is flagged with `missing_direct_file_access_protection`.

## Expected behaviour

No error. The file contains no executable code and is safe for direct access.

## Actual behaviour

`missing_direct_file_access_protection` error is reported at line 0.

*Disclaimer: Claude was used to investigate this and write the issue. I verified its outputs.*

GeorgeWebDevCy:
Fixes #1361

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1369-bbf080b58df7546dfb50add159fb265e6e012f32.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


davidperezgar:
Thanks @GeorgeWebDevCy for your contribution. You need to review the tests that are failing before we can review the PR.

GeorgeWebDevCy:
Thanks for the heads up. I reviewed the failing tests and pushed a follow-up fix in 3813175092f61553c18220e97aed6172a3b92e52. The CI matrix is now passing.

.github/workflows/php-test.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         if: ${{ matrix.coverage }}
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
         with:
-          file: build/logs/*.xml
+          files: build/logs/*.xml
```
ernilambar, line 115: Why this change?

GeorgeWebDevCy: This updates the input name for the pinned Codecov action. At commit `fb8b3582c8e4def4969c97caa2f19720cb33a72f`, the action defines `files` (a comma-separated list of explicit coverage files) and passes it through as `INPUT_FILES`; it does not define a `file` input. Both occurrences use the same `build/logs/*.xml` pattern, so this makes the workflow use the action's supported input without changing which reports we intend to upload.

davidperezgar:
Thanks to Codex’s review, I found a potential false negative in the new include handling.

Any `include`/`require` is now considered safe whenever the file contains a class, interface, trait, or enum. However, including a file executes it, so a class file that requires a bootstrap file with procedural code could be accessed directly without a guard and would no longer be reported.

Could we avoid skipping arbitrary includes, or verify that the included file is also structural-only? It would also be helpful to add a regression test for a class file that requires a file containing executable code.

PR #1380, opened 2026-06-29
Add Inlined React Runtime check for React 19 incompatibilities

### LINKED ISSUE #1356: Enhancement: Add a check to warn about React 19 incompatibilities / bundled outdated React
WordPress is planning to upgrade its bundled React from 18 to 19.
- [Original announcement](https://make.wordpress.org/core/2026/05/27/react-19-upgrade-in-wordpress/). 
- [Temporary revert](https://make.wordpress.org/core/2026/06/05/react-19-upgrade-temporarily-reverted-in-gutenberg/)

Real-world compatibility testing in the Gutenberg React 19 polyfill PR ([WordPress/gutenberg#78899](https://github.com/WordPress/gutenberg/pull/78899)) showed that large, widely-installed plugins (Yoast SEO, Elementor, Fluent Forms, Sure Cart, Sure Forms, etc.) break under React 19.

**~99% of these failures share a single root cause: plugins inline the `react/jsx-runtime` library code instead of externalizing it.**

The JSX runtime is used by transpiled JSX. Syntax like:

```jsx
<Foo className="foo">hello</Foo>
```

is transpiled into:

```js
import { jsx } from 'react/jsx-runtime';
jsx( Foo, { className: 'foo' }, 'hello' );
```

and at runtime `jsx()` returns an "element" object:

```js
{
  $$typeof: Symbol.for( 'react.element' ),
  type: Foo,
  props: { className: 'foo', children: 'hello' },
}
```

These objects are then handed to React to render. **Between React 18 and 19 the element object shape changed** — only slightly, but enough that React 19 rejects objects produced by the React 18 runtime. So if a plugin inlines the React 18 `jsx-runtime`, it produces incompatible objects and feeds them to WordPress's React 19.

**The fix is to always externalize the JSX runtime.** WordPress provides the `react-jsx-runtime` script handle, and bundled plugin code should reference it through the global `window.ReactJSXRuntime`. `@wordpress/scripts` / `@wordpress/dependency-extraction-webpack-plugin` does this automatically; the problem occurs when a plugin's build inlines the runtime instead.

Plugin Check is well placed to catch this before a plugin ships.

### Proposal

Add a static-analysis check that scans a plugin's built JavaScript for an **inlined JSX runtime**, and passes plugins that correctly externalize it.

#### Primary signal — inlined JSX runtime (high confidence)

The inlined runtime bakes the element tag into the bundle as a literal `Symbol.for()` call. The string argument survives minification (it is a runtime value, not a renameable identifier), so it is a reliable fingerprint:

- **`Symbol.for("react.element")`** — present when a **pre-19** JSX runtime is inlined. React 19's runtime uses `Symbol.for("react.transitional.element")`, so finding the non-transitional string is strong evidence of an inlined React 18-era runtime being fed to React 19. **This is the headline detector.**
- Supporting fingerprints (optional, to raise confidence): the runtime's exported helpers `jsx`, `jsxs`, `jsxDEV` defined inline alongside the `Symbol.for("react.element")` tag.

#### "All clear" / suppression — correctly externalized

A properly externalized plugin will **not** contain the inlined runtime. Use these as suppression signals to keep false positives low:

- The generated `*.asset.php` dependency array lists `react-jsx-runtime` (and/or `react`, `react-dom`) — the runtime is provided by core, not bundled.
- The built JS references `window.ReactJSXRuntime` (the external global the runtime is mapped to).

If either is present and no inlined-runtime fingerprint is found, the plugin is doing it right.

#### Secondary signal — removed React APIs (lower volume)

A minority of failures came from removed/changed APIs rather than the runtime. These are distinctive tokens that also survive minification and can be reported with their own message:

| Token | Issue | Guidance |
| --- | --- | --- |
| `unmountComponentAtNode` | Removed in React 19 | Use `root.unmount()` from `createRoot` |
| `findDOMNode` | Removed in React 19 | Use refs |
| `hydrate` (`ReactDOM.hydrate`) | Removed in React 19 | Use `hydrateRoot` |
| `ReactCurrentOwner` | Internal, restructured in React 19 | Don't access React internals |

(`ReactDOM.render` / bare `.render(` should be excluded — too noisy in minified bundles.)

### Message / guidance

When the inlined runtime is detected, the message should point at the real fix, e.g.:

> This file appears to inline the React JSX runtime (`react/jsx-runtime`), which is incompatible with WordPress's React 19. Externalize it instead and reference WordPress's `react-jsx-runtime` script (`window.ReactJSXRuntime`). When building with `@wordpress/scripts`, this happens automatically.

Link to the [React 19 upgrade post](https://make.wordpress.org/core/2026/05/27/react-19-upgrade-in-wordpress/).

### Severity

**Warning, not error** — detection is heuristic (minification, bundler differences, custom builds), and while inlining the runtime is the wrong choice for WordPress, it is not strictly forbidden. (Open to making it an error given how high-confidence the primary signal is — see open questions.)

### Implementation notes

Fits the existing check architecture with no new infrastructure:

- New class under `includes/Checker/Checks/Plugin_Repo/` (or `Performance/`) extending `Abstract_File_Check`, using the `Stable_Check` and `Amend_Check_Result` traits — same shape as `File_Type_Check`, which already fingerprints bundled libraries (jQuery, Backbone, Underscore) by pattern.
- Logic in `check_files()`: `filter_files_by_extension( $files, 'js' )`, then `files_preg_match_all()` for `Symbol.for("react.element")` (allowing for whitespace/quote variations). Optionally read sibling `*.asset.php` files to apply the externalization suppression.
- Report via `add_result_warning_for_file()` with file + line and a `docs` link.
- Register the slug in `Default_Check_Repository.php`.
- Add a test under `tests/phpunit/tests/Checker/Checks/` with fixture plugins in `tests/phpunit/testdata/plugins/`: one with an inlined `Symbol.for("react.element")` runtime (should warn), and one that externalizes via `react-jsx-runtime` / `window.ReactJSXRuntime` (should pass). Mirror `File_Type_Check_Tests`.

### Open questions

1. **Regex robustness** — confirm `Symbol.for("react.element")` survives across the common build pipelines (`@wordpress/scripts`, CRA, Vite, esbuild) and quote/whitespace variants. Worth grepping a couple of the known-broken plugin builds (Yoast, Elementor) to validate before finalizing the pattern.
2. **Error vs. warning** — given the primary signal is high-confidence and the impact is "breaks on current WordPress," should this be an error rather than a warning?
3. **Category** — Plugin_Repo, Performance, or a dedicated compatibility bucket?
4. **Transitional tag** — should we also flag `react.transitional.element` anywhere, or is its presence harmless (it indicates a React 19 runtime, which is fine, though inlining even the v19 runtime is still suboptimal vs. externalizing)?

### References

- [React 19 upgrade in WordPress](https://make.wordpress.org/core/2026/05/27/react-19-upgrade-in-wordpress/)
- [WordPress/gutenberg#78899 — React render polyfills + real-world plugin testing](https://github.com/WordPress/gutenberg/pull/78899)

gunjanjaswal:
Closes #1356.

This adds a check (`inlined_react_runtime`) for plugins that bundle their own copy of the React JSX runtime. WordPress is going to React 19 at some point, and the usual reason a plugin breaks across that jump isn't some exotic API, it's that the build inlined `react/jsx-runtime` into the bundle instead of loading it from WordPress. React 19 changed the shape of the element object, so anything an old bundled runtime creates gets rejected. The plugin runs fine today and then falls over the day WordPress updates, so the point is to catch it ahead of time.

The detection is pretty crude, honestly: it greps the plugin's JS for `Symbol.for('react.element')`, which only shows up when a pre-19 runtime got bundled in. React 19 switched that marker to `react.transitional.element`, so anyone externalizing the runtime properly won't trip it. And if the runtime is externalized there's nothing to warn about, so it stays quiet when it spots `window.ReactJSXRuntime` in the file or `react-jsx-runtime` listed in the matching `*.asset.php`. On top of that it flags a few APIs that got dropped in 19 (`unmountComponentAtNode`, `findDOMNode`, `ReactCurrentOwner`). Everything's a warning, nothing errors out. It's wired into `Default_Check_Repository` with the usual docs and changelog entries.

`Inlined_React_Runtime_Check_Tests` covers both sides: the `-with-errors` fixture throws two warnings (`inlined_jsx_runtime` on index.js, `react_removed_api` on legacy.js), and the `-without-errors` one stays clean because the runtime's externalized there, once through `index.asset.php` and once through `window.ReactJSXRuntime`. I'm not married to the category or severity if you'd rather scope it differently.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1380-fc6b9a5f91714890e19847eb8c3bb1dfad732918.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

## AI Usage Disclosure

AI assistance (Claude Code) was used to help draft the check and its tests. I reviewed and understand every line and take responsibility for it.


davidperezgar:
Hello, I've got these findings from Codex. Could you check it?

Findings
[P1] Asset dependency suppresses a real inlined-runtime signal
[Inlined_React_Runtime_Check.php (line 167)](/private/tmp/plugin-check-pr1380/includes/Checker/Checks/Performance/Inlined_React_Runtime_Check.php:167) suppresses Symbol.for( 'react.element' ) whenever the sibling asset file contains react-jsx-runtime. But the marker is in the JS file being scanned; an asset dependency only proves the file declares an external dependency, not that it did not also inline a pre-19 runtime. This can miss the exact problem the check is meant to catch, especially with stale/manual asset files or mixed builds. I’d only use react-jsx-runtime in the asset file as supporting context, not as a bypass once the marker is present.

[P2] Removed API regex reports comments/strings as usage
[Inlined_React_Runtime_Check.php (line 125)](/private/tmp/plugin-check-pr1380/includes/Checker/Checks/Performance/Inlined_React_Runtime_Check.php:125) scans raw JS for unmountComponentAtNode, findDOMNode, and ReactCurrentOwner, so a comment, changelog string, translation string, or compatibility message will produce a warning. Since this is registered as a stable default check, that could create noisy false positives. Consider limiting this to property/member access patterns or parsing tokens enough to skip comments and string literals.

gunjanjaswal:
Thanks @davidperezgar, both are fair points. I pushed a fix for each.

**P1 (asset dependency bypass):** You're right that the `.asset.php` dependency only tells us the file *declares* `react-jsx-runtime`, not that it avoided inlining a pre-19 runtime. Since `Symbol.for( 'react.element' )` is already a definitive inline marker, trusting the asset file there can hide the exact case the check is meant to catch. I dropped the asset file as a bypass, so now only an in-file `window.ReactJSXRuntime` reference suppresses the warning. I also added a fixture (`asset-declared.js` with its `.asset.php`) that inlines the runtime while declaring the dependency, and it's now flagged correctly.

**P2 (comments and strings):** Also right. The raw scan would flag `unmountComponentAtNode`, `findDOMNode`, or `ReactCurrentOwner` even inside a comment, changelog line, or translation string. It now blanks out comments and string literals before scanning, keeping the length and newlines intact so the reported line and column stay accurate, and only real usages get flagged. I added a `comment-only.js` fixture that mentions all three in prose and stays clean.

Happy to adjust either if you'd prefer a different approach.

gunjanjaswal:
Bumping this gently. The Inlined React Runtime check has been green for a few weeks now and still mergeable. If anything on trunk has moved under it I'm glad to rebase; otherwise it's ready for a review pass whenever the team has bandwidth.

jsnajdr (approved):
I think this is ready to ship, it will be very useful for the React 19 rollout we are planning for Gutenberg in the WP 7.2 cycle.

gunjanjaswal:
Thanks @jsnajdr, and thanks again for strengthening the detection before this landed. Glad it lines up with the WP 7.2 React 19 rollout, that's exactly the kind of breakage this is meant to catch early. Happy to help if anything else comes up as you start testing plugins against it.

swissspidy (approved):

ernilambar:
Reviewed by AI: Opus 5.5

## Summary
This PR adds a `react_usage` static check. It reports **errors** when a JS file inlines a pre-19 `react/jsx-runtime`, `react` or `react-dom`, and **warnings** when it calls React APIs that React 19 removed. The primary signal (`"react.element"` plus a per-package fingerprint) covers the issue's headline case well. However, the package fingerprints are loose enough to produce false **errors** in a stable, default-on check, and some messages overstate what actually breaks on WordPress.

## ✅ What's good
- Two-step detection (element marker + package-specific fingerprint) is a real improvement over the issue's bare `Symbol.for("react.element")` grep. `react-is` no longer triggers it (`react-is.js` fixture).
- Matching the string literal rather than `Symbol.for(...)` catches React 17 production builds that hoist `Symbol.for` into a local (`react-17-prod.js`).
- The `*.asset.php` bypass was dropped as requested in the P1 feedback. `asset-declared.js` locks that in.
- The per-package codes (`inlined_react_jsx_runtime`, `inlined_react`, `inlined_react_dom`) give actionable output. Detecting development builds is a nice touch.
- Blanking comments and strings before the removed-API scan preserves offsets, so line and column stay correct.
- The fixtures cover mixed builds well: tree-shaken `jsxs`, and `window.ReactDOM` not suppressing an inlined `react`.
- PHPCS passes on the new files. All CI jobs are green (PHP 7.4–8.4, WP 6.3/latest/trunk, PHPStan, sniffs).

## ⚠️ Must-fix before merge
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:234`** — The `react/jsx-runtime` fingerprint `/\bjsxs?\s*[:=][^=]/` matches any `jsx` key or assignment in the file. Any bundle that contains `"react.element"` for an unrelated reason (e.g. `react-is` pulled in via `prop-types` / `hoist-non-react-statics`, which is very common) gets an **error** if it also has:
  - `Prism.languages.jsx = …`
  - `{ ecmaFeatures: { jsx: true } }`
  - a highlight.js or CodeMirror `jsx:` alias
  - any local `const jsx = …` that isn't literally `window.ReactJSXRuntime`

  I confirmed this with synthetic inputs: `Symbol.for("react.element")` together with `Prism.languages.jsx = …` or `{jsx:true}` is reported as `inlined_react_jsx_runtime`. The fingerprint needs to be tied to the runtime itself. For example, require the `jsx` assignment on an exports-like object, or co-occurrence with the runtime's `ref`/`key`/`_owner` element literal or its `"react.fragment"` export.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:244`** — The `react` fingerprint also matches **preact/compat**, which defines both `Symbol.for('react.element')` and `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED = {…}`. A Preact plugin never touches WordPress's React, yet it would get an error saying it "breaks when WordPress upgrades to React 19".
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:188-208`** — The errors claim every inlined package "breaks when WordPress upgrades to React 19". That is only true when elements cross the boundary: an inlined runtime or `react` feeding WordPress's `react-dom` / `@wordpress/components`, or the reverse. A fully self-contained React 18 bundle (runtime + `react` + `react-dom` all inlined, rendering its own root) gets three errors but keeps working. It's bloat, not breakage. Also, the issue proposed **warning** severity ("detection is heuristic"), and the PR description still says "Everything's a warning, nothing errors out". The PR now emits errors at severity 6/7. Pick one, and make it explicit:
  - (a) Error only for mixed builds, where at least one of the three packages is inlined and another is externalized; warning otherwise.
  - (b) Warning across the board.

  Either way, the message text must match what actually happens.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:302`** — The Gutenberg React 19 work ([WordPress/gutenberg#78899](https://github.com/WordPress/gutenberg/pull/78899), merged) polyfills `render`, `hydrate` and `unmountComponentAtNode` in `@wordpress/element` **and in the `react-dom` script**. A plugin that calls `ReactDOM.render/hydrate/unmountComponentAtNode` on WordPress's externalized `react-dom` does not "stop working once WordPress upgrades React". The warning is false for those three. Either reword it for them (deprecated, polyfilled, migrate to `createRoot`/`hydrateRoot`/`root.unmount()`) or give them a separate, softer message. `findDOMNode`, `unstable_renderSubtreeIntoContainer`, `renderToNodeStream` and `createFactory` are not polyfilled and can keep the current wording.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:352`**, **`:367`** — The `replacement` values `'a ref on the element'` and `'JSX or createElement()'` are English prose substituted into a translated sentence via `%2$s`, so they can never be translated. Pass only code identifiers through the placeholder (`createRoot()`, `hydrateRoot()`, …). For the two prose cases, use a dedicated translatable string, or wrap the prose in `__()`.
- **`readme.txt:109`** — The changelog entry is added under `= 2.0.0 =`, but 2.0.0 and 2.1.0 are already tagged and released (`plugin.php` is at 2.1.0). Move it to the next release section, or leave it for the release PR.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:33`** (and every `@since 2.0.0` in the file) — Same problem. This code ships in the next release, not 2.0.0. Update `@since` to the next version.

## 💡 Suggestions (optional)
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:239`** — Global suppression treats *any* `window.React` reference as proof of externalization. A bundle that inlines React and exposes it with `window.React = exports` (a common pattern) is therefore silently cleared. I confirmed this false negative. Consider matching only read positions, e.g. rejecting `window.React\s*=(?!=)`.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:229`** — Unminified webpack output emits externals as `window["ReactJSXRuntime"]` / `window["React"]`. Terser only rewrites them to dot access when minifying. Accepting `window\[\s*(['"])ReactJSXRuntime\1\s*\]` makes suppression work for development builds too.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:386`** — The comment/string blanker has two weaknesses:
  - It does not handle regex literals. `var r=/"/g;findDOMNode(e);var s="x";` swallows the `findDOMNode` call, a false negative I confirmed.
  - On large vendor bundles with long string literals, PCRE hits its recursion limit. On a 3 MB literal it returned `null` and fell back to the raw contents, which silently brings back the comment/string false positives that P2 fixed.

  Either document this as an accepted limitation, or skip blanking above a size threshold and report nothing for the removed APIs in that file.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:66`** — `CATEGORY_PERFORMANCE` is an odd home for what is mainly a compatibility check. Only the development-build note is performance-related. `CATEGORY_GENERAL` fits better, unless the team wants a compatibility bucket (issue open question 3).
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:46`** — The issue asked for a link to the [React 19 upgrade post on make.wordpress.org](https://make.wordpress.org/core/2026/05/27/react-19-upgrade-in-wordpress/). That is more relevant to WordPress plugin authors than react.dev's generic guide, especially for the `inlined_*` errors.
- **`includes/Checker/Checks/Performance/React_Usage_Check.php:93`** — Skipping the removed-API scan for any file with an inlined package also hides the plugin's own `findDOMNode` calls when only the JSX runtime was inlined. That's acceptable, but worth a line in the docblock since the comment only justifies the `react-dom` case.
- **`tests/phpunit/tests/Checker/Checks/React_Usage_Check_Tests.php`** — Add negative fixtures for the false-positive shapes above: `react-is` + `jsx:` key, preact/compat, and a `window["ReactJSXRuntime"]` dev build. There is also no fixture proving that `window.React` / `window.ReactDOM` suppress their own packages.
- **PR description** — It is stale: it still describes `inlined_react_runtime`, `Inlined_React_Runtime_Check_Tests`, `inlined_jsx_runtime`, `ReactCurrentOwner` detection, `*.asset.php` suppression and warnings only. Update it so the merge commit and changelog reflect what ships.

## Verdict
**Request Changes**: the detection idea is sound, but a stable, default-on check emitting **errors** needs tighter fingerprints (JSX-key and preact/compat false positives) and messages that match actual React 19 behavior on WordPress (self-contained bundles, polyfilled `render`/`hydrate`/`unmountComponentAtNode`), plus the version and i18n fixes.

jsnajdr:
Hi @gunjanjaswal :wave: Could you please change the PR title to:
> Add React usage check to detect React 19 incompatibilities

and edit the PR description so that it's more up to date after recent changes. I can be much shorter, it doesn't need to describe any implementation details. They change quickly, as the detection heuristic is improved.

And this is no longer true:
> Everything's a warning, nothing errors out

Detected React 19 incompatibilities are marked as errors, as they really have the potential to break the site. `plugin-check` has many more benign issues marked as errors.

I'll soon propose some code changes as another stacked PR, but I can't modify the PR GitHub info itself.

gunjanjaswal:
Done — retitled to "Add React usage check to detect React 19 incompatibilities" and trimmed the description down. Dropped the implementation walkthrough (it moves too fast to keep in sync) and corrected the severity note: incompatibilities are reported as errors, with warnings only for the removed-API calls. Looking forward to your stacked changes.

jsnajdr:
> Reviewed by AI: Opus 5.5

I'm addressing most of the feedback in the stacked https://github.com/gunjanjaswal/plugin-check/pull/2 PR.

### Must fix
- `react/jsx-runtime` detection now also requires an `_owner:` element factory, so a `jsx:` key from Prism or other libraries is not falsely detected as JSX runtime.
- `preact/compat` false positive: `react` detection now requires the same `_owner:` marker, which `preact/compat` does not produce because it builds Preact vnodes.
- will WordPress break or not? Changing the messages to "will likely break" and keeping them as errors. Bundling React is very risky, the page itself can be fine, but then WordPress can load another React UI on the same admin page, like `core-commands`. The two Reacts won't like each other. We don't need to run a 100% reliable static analysis to flag something as error. High risk is enough.
- untranslatable messages: fixed by using custom messages for the two APIs that were using the bad interpolation.
- fixed the 2.0.0 version mismatches. Removed the changelog entry (release commit will add it), updated the `@since` comments.

### Suggestions
- more precise detection of `window.React = exports`: distinguish between reading (counts as externalizing) and assignment (clear signal that bundling happens).
- blanker weaknesses: regexp was replaced with a token scanner that detects regex literals, and doesn't have the PCRE recursion limit.
- changed the category from Performance to General. Keeping detected bundling as errors.
- documented why we skip check for removed APIs after detecting a bundled package. It makes the detection less reliable, and it's an irrelevant issue for a script that bundles React.
- added fixtures for `prism-jsx.js`, `preact-compat.js` and `fiber-inspector.js`.

### Other fixes
- correctly detect line endings and file positions. Use the file's own line endings rather than `PHP_EOL`.
- detect `react.element` as a template literal, in backtick quotes. Some minifiers do that.
- detect bundled `react-is` with a new `stale_react_is warning`, another upgrade risk.

### Not done
- not implementing the `window["React"]` detection because the pattern is not used in practice. Bundlers and minifiers minify this to `window.React`.
- still reporting the detected bundled React as an error. We have plenty of other errors where the plugin continues to work "just fine". One example is `WordPress.WP.EnqueuedResources.NonEnqueuedScript` that errors on a literal `<script>` tag in the HTML output. Such a script works, of course, but it's a risky anti-pattern worth flagging.
- not linking to a make.wordpress.org post because we don't yet have one that would be 100% relevant and precise about the current situation. Can update in near future.

gunjanjaswal:
Merged your stacked #2, so everything above is now in this PR — the `_owner:` element-factory gating for `react`/`jsx-runtime`, the token scanner, the `react-is` handling, line-ending fixes, the move to General, and the "will likely break" wording kept as errors. Thanks for driving all of it.

jsnajdr:
Thanks @gunjanjaswal, there is one more little fix that addresses the "PHP Code Linting" CI failure: #3. Fixes a too high complexity of the `blank_comments_and_strings`.

ernilambar:
Can you please check if following issues are valid? 
- **`includes/Checker/Checks/General/React_Usage_Check.php:492`**, **`:497`**, **`:502`**, **`:507`** — Every removed-API pattern requires the name to be followed directly by `(`. Webpack 5 compiles a named import from an external as `(0,r.findDOMNode)(this)`, with a `)` between the name and the call. That covers `import { findDOMNode } from 'react-dom'` and the same import from `@wordpress/element`, which is exactly what `@wordpress/scripts` builds produce. I verified that `(0,r.unmountComponentAtNode)(e)`, `(0,r.findDOMNode)(this)` and the unminified `(0,react_dom__WEBPACK_IMPORTED_MODULE_0__.findDOMNode)(this)` all produce **no warning**. As written, the scan only catches hand-written, unbundled code like the `legacy.js` / `hydrate.js` fixtures. Fix: allow an optional closing paren for the four distinctive names, e.g. `/\bfindDOMNode\s*\)?\s*\(/`. Add a fixture with the `(0,x.name)(…)` shape. `ReactDOM.render`/`hydrate` can't safely take the same treatment (`(0,r.render)` is far too generic), so say in the docblock that they only match the literal `ReactDOM.` form.
- **`includes/Checker/Checks/General/React_Usage_Check.php:453`** — This was raised in the previous round and has had no reply. For `ReactDOM.render`, `ReactDOM.hydrate` and `unmountComponentAtNode`, the shared message says the call "stops working once WordPress upgrades React". WordPress/gutenberg#78899 polyfills exactly these three in `@wordpress/element` and in the `react-dom` script, so for a plugin using WordPress's externalized React DOM, that sentence is false. The class docblock (lines 29–34) already takes this nuance into account for element shapes. The removed-API messages should do the same. Either give these three a "deprecated, polyfilled by WordPress, migrate to `createRoot()`/`hydrateRoot()`/`root.unmount()`" message, or reply on the thread explaining why the current wording should stand.

gunjanjaswal:
Both are valid, thanks for the careful testing — and sorry the line-453 point sat from last round without a reply.

On the removed-API patterns: you're right that requiring the name immediately before `(` misses the `@wordpress/scripts` external-interop form. `(0,r.findDOMNode)(this)` and the unminified `(0,react_dom__WEBPACK_IMPORTED_MODULE_0__.findDOMNode)(this)` slip straight through, so in practice the scan only catches hand-written unbundled code. Allowing an optional closing paren for the four distinctive names, plus a `(0,x.name)(…)` fixture, is the right fix, and agreed that `ReactDOM.render`/`hydrate` can't take the same widening since `(0,r.render)` is far too generic — that stays literal, documented in the docblock.

On line 453: also agreed. With WordPress/gutenberg#78899 polyfilling `render`/`hydrate`/`unmountComponentAtNode` in `@wordpress/element` and the `react-dom` script, "stops working once WordPress upgrades React" is wrong for a plugin on WordPress's externalized React. Those three should get the deprecated-but-polyfilled wording pointing at `createRoot()`/`hydrateRoot()`/`root.unmount()`, matching the nuance the class docblock already carries for element shapes.

@jsnajdr has been authoring the detection through the stacked PRs and has more queued — these two fit naturally there, so we'll fold both into the next one. I'll make sure they land.

jsnajdr:
> Webpack 5 compiles a named import from an external as `(0,r.findDOMNode)(this)`

This one is real, quite a glaring omission 😨. Fixing in https://github.com/gunjanjaswal/plugin-check/pull/4. It was missed because the corpus of some 10+ plugins that I'm locally testing on always contained a "normal" call to `findDOMNode`, inside a bundled library.

> polyfills exactly these three in `@wordpress/element` and in the `react-dom` script, so for a plugin using WordPress's externalized React DOM, that sentence is false

I don't think this is valid. Yes, in a literal sense the sentence is false, because in Gutenberg we patch `react-dom` to continue supporting the removed APIs. But that's an emergency measure so that we don't break sites, not a signal that everything continues to be OK. The Plugin Check should flag this with a strong language.

ernilambar (approved):

jsnajdr:
OK, we have all the required approvals, CI is green... Anything else needed to merge?

PR #1391, opened 2026-07-11
fix(checks): validate each Requires Plugins dependency individually

### LINKED ISSUE #1385: Enhancement: Support checking and reporting multiple dependent plugins
Currently, Plugin Check primarily validates dependencies one at a time. While WordPress supports declaring multiple plugin dependencies, the results could be improved by providing clearer validation and reporting when a plugin depends on more than one plugin.

Enhancing Plugin Check to better handle multiple dependent plugins would help plugin authors verify that all declared dependencies are correctly detected and validated during a scan.

Suggested improvements:
- Validate all declared plugin dependencies instead of focusing on a single dependency.
- Display each dependency individually in the scan results.
- Clearly indicate whether each dependency is valid, missing, inactive, or invalid.
- Provide actionable messages for any dependency-related issues.

Expected Result
Plugin Check should detect and validate all declared plugin dependencies and present their status clearly in the results, making it easier for plugin authors to identify and resolve dependency-related issues.

faisalahammad:
## Summary
Plugin Check currently validates the `Requires Plugins` header as a single comma-separated blob, so a plugin declaring more than one dependency gets one generic "header is not valid" error with no indication of which slug is the problem, or whether the declared dependency actually exists in the WordPress.org plugin directory. This updates `Plugin_Header_Fields_Check` to validate each declared dependency individually and report its status on its own.
Fixes #1385

### includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
**Before:**
```php
if ( ! preg_match( '/^[a-z0-9-]+(?:,\s*[a-z0-9-]+)*$/', $plugin_header['RequiresPlugins'] ) ) {
    // one error for the entire header, doesn't say which slug is bad
}
```
**After:**
```php
private function check_requires_plugins_header( Check_Result $result, string $requires_plugins, string $label, string $plugin_main_file ) {
    $slugs = array_map( 'trim', explode( ',', $requires_plugins ) );

    foreach ( $slugs as $slug ) {
        if ( '' === $slug ) {
            continue;
        }

        if ( ! preg_match( '/^[a-z0-9]+(?:-[a-z0-9]+)*$/', $slug ) ) {
            // error naming this specific slug
            continue;
        }

        $this->check_requires_plugins_slug_status( $result, $slug, $plugin_main_file );
    }
}
```
Each malformed slug is now reported as its own error, naming the slug. Each validly formatted slug is checked against the WordPress.org plugin directory (via `https://api.wordpress.org/plugins/info/1.0/{slug}.json`, transient-cached for a day): a warning is added if the slug can't be found there (covers both nonexistent and closed/removed plugins, since the directory API returns an `error` key for both). If the directory API can't be reached, the check skips silently rather than risk a false positive.

**Why:** Plugin authors declaring multiple dependencies had no way to tell which one was wrong. ernilambar pointed out in review that checking local install/active state isn't this check's job — a `plugin_repo` category check should validate repo-readiness, i.e. whether the declared slug actually exists in the WordPress.org directory, since a parent plugin missing from the directory means the child plugin can't be published there either. The original approach (checking `is_plugin_inactive()` against the local scan environment) has been replaced entirely with this directory-existence check.

## Testing
**Test 1: Multiple invalid slugs are reported individually**
1. Set `Requires Plugins: Example Plugin, OtherPlugin` in a test plugin's header.
2. Run Plugin Check against it.
Result: two separate errors, one naming "Example Plugin" and one naming "OtherPlugin", instead of a single generic message.

**Test 2: Valid multi-dependency header passes format validation**
1. Set `Requires Plugins: woocommerce, contact-form-7`.
2. Run Plugin Check.
Result: no `plugin_header_invalid_requires_plugins` errors.

**Test 3: Dependency directory-existence warnings**
1. Set `Requires Plugins: plugin-check, hello-dolly, not-a-real-plugin`.
2. Run Plugin Check.
Result: warning naming `not-a-real-plugin` as not found in the WordPress.org plugin directory; no warning for `plugin-check` or `hello-dolly` (both published there).

Updated PHPUnit tests in `tests/phpunit/tests/Checker/Checks/Plugin_Header_Fields_Check_Tests.php` (transients are seeded directly in tests rather than mocking the HTTP call, matching the existing convention used for `Version_Utils`/`RequiresWP` in the same file/class). Full suite (482 tests) passes, phpcs and phpstan clean.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1391-0244155d18f8e7d184db5335d16321580658645c.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
+++ b/includes/Checker/Checks/Plugin_Repo/Plugin_Header_Fields_Check.php
@@ -525,8 +525,7 @@ class Plugin_Header_Fields_Check implements Static_Check {
 	 * Validates each slug declared in the "Requires Plugins" header individually.
 	 *
 	 * Reports an error per malformed slug, and a warning per validly formatted
-	 * slug that is not installed or not active in the environment the check
-	 * is running in.
+	 * slug that could not be found in the WordPress.org plugin directory.
 	 *
 	 * @since 2.1.0
 	 *
@@ -567,7 +566,7 @@ class Plugin_Header_Fields_Check implements Static_Check {
 	}
 
 	/**
-	 * Checks whether a single declared plugin dependency is installed and active.
+	 * Checks whether a single declared plugin dependency exists in the WordPress.org plugin directory.
 	 *
 	 * @since 2.1.0
 	 *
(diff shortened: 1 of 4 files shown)


davidperezgar (approved):

davidperezgar:
Thanks for the contribution. I see reasonable.

ernilambar:
I think PR is following incorrect approach. Checking whether parent plugin is installed/active or not the responsibility of PCP or the actual check it was intended. We should check of that slug is active and published in the WordPress.org directory.
Like if "parent-plugin" is not there in the directory, child plugin will not even published in the directory. This is what the check should be focused.

faisalahammad:
Thanks for the feedback, this makes sense. I agree checking local install/active state is not the right job for a `plugin_repo` check.

Pushed an update (689a1a14): the check now queries the WordPress.org plugin directory (`api.wordpress.org/plugins/info/1.0/{slug}.json`, cached in a transient for a day, same pattern the existing `Requires at least` header validation already uses in this file) and warns only when the declared slug can't be found there. That covers both nonexistent slugs and closed/removed plugins, since the directory API returns an `error` key for both cases. Local install/active check is removed entirely, nothing left checking runtime state.

If the API is unreachable the check skips silently rather than risk a false positive, same convention as the existing WP-version check.

davidperezgar:
I reviewed this with Codex and found two issues:

1. The WordPress.org API returns `404 {"error":"Plugin not found."}` for an unknown slug. Since the code treats every non-200 response as an unreachable API and returns `null`, missing plugins are skipped silently instead of generating the intended warning. Could we handle `404` as “not found” and reserve `null` for actual transport/server failures?

2. Empty dependency entries are now silently ignored. For example, `foo,`, `,foo`, or `foo,,bar` would pass validation, whereas the previous full-header regex rejected them. Could we report these as invalid `Requires Plugins` values?

faisalahammad:
Both issues raised in @davidperezgar's July 25 review are now addressed in commit 0244155d:

**Issue 1 (404 handling):** `plugin_exists_in_directory()` now separates HTTP 404 from transport failures:
- 404 → slug confirmed absent from the directory. Caches result as `'no'` (same key existing tests consume) and returns `false`, triggering the expected warning.
- WP_Error or non-200/non-404 → returns `null`. No cache, silent skip, retries next run.
- Cached `'no'` readout: `return 'no' !== $exists` — `false` for `'no'`, `true` for everything else.

**Issue 2 (empty entries):** `check_requires_plugins_header()` emits a `plugin_header_invalid_requires_plugins` error with message "...contains an empty entry." for each blank token in the comma-separated Requires Plugins header. Same error code as malformed-slug errors, distinct message. Empty-token detection runs before slug-format validation, so `Requires Plugins: Example ,,OtherPlugin` produces 3 errors: 1 empty + 2 invalid-format.

**Tests:** New fixture + `test_run_with_empty_entry_in_requires_plugins_header` asserting 3 errors. Existing `test_run_with_requires_plugins_dependency_status` comment updated to document `'no'` = HTTP-404 cache semantics. All quality gates green: phpcs clean, phpstan clean, 18/18 tests, 50 assertions.

davidperezgar:
Thanks for addressing both issues. I reviewed the updated commit, and the 404 handling and empty-entry validation now look correct to me.

davidperezgar (approved):

ernilambar (approved):

PR #1411, opened 2026-07-24
Fix silently dropped WordPress core functions in since-data generator

### LINKED ISSUE #1398: Issue related to since scanning
## Issue

`tools/generate-wp-function-since-data.php` silently drops some WordPress core functions from `includes/Vars/wp-functions-since.json`, even though they exist in the scanned source and have a valid docblock. This happens deterministically (not "flaky") for specific docblock/declaration patterns.

### 1. `@since MU (3.0.0)` docblocks are never matched

Functions ported from WordPress MU use the core-standard tag `@since MU (3.0.0)` instead of a plain version number (this is an explicit, frozen exception in the [[WP inline documentation standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)). The extraction regex:

```php
preg_match( '/@since\s+([0-9]+(?:\.[0-9]+){1,2})/i', $doc_comment, $matches )
```

requires digits immediately after `@since`, so it never matches `MU`. Since there's no fallback, `extract_since_version()` returns `''`, and the function is skipped entirely (`continue`) instead of being recorded with any version.

**Impact:** every core function still carrying `@since MU (3.0.0)` is missing from the dataset.

### 2. By-reference functions are misdetected as anonymous and skipped

For `function &foo() {}`, the name-detection loop only checks for whitespace or `T_STRING` after `T_FUNCTION`:

```php
for ( $j = $i + 1; $j < $count; $j++ ) {
    $next = $tokens[ $j ];
    if ( is_array( $next ) && T_WHITESPACE === $next[0] ) { continue; }
    if ( is_array( $next ) && T_STRING === $next[0] ) { $name_index = $j; }
    break;
}
```

The `&` token is a single-char string token, not `T_STRING`, so the loop breaks immediately with `$name_index = null`. The function is then treated as anonymous and skipped.

**Impact:** any core function declared with a by-reference return (`function &name(...)`) never appears in the dataset.

### 3. (Lower risk, worth tracking) Docblock lookup can break on PHP 8 attributes

`get_previous_doc_comment()` walks backwards from the `function` token and only tolerates whitespace, comments, and a fixed set of modifier keywords (`T_PUBLIC`, `T_STATIC`, etc.). A PHP 8 attribute (`#[...]`) placed between a docblock and a function declaration is not in the allow-list, so the backward scan stops there and the docblock is never found — same failure mode as above (function silently dropped). Not currently confirmed to hit any function in core today, but worth guarding against as core adopts more PHP 8 syntax.

### 4. Scan scope excludes root-level core files

Only `wp-includes/` and `wp-admin/` are scanned. Root files such as `wp-cron.php`, `wp-links-opml.php`, `wp-mail.php`, `wp-signup.php`, and `xmlrpc.php` also define global functions with real `@since` docblocks (e.g. `confirm_blog_signup()` in `wp-signup.php`) and are currently never scanned.

> PS: Generated with help of Claude.

ernilambar:
Fixes https://github.com/WordPress/plugin-check/issues/1398

## Summary

`tools/generate-wp-function-since-data.php` was silently dropping valid WordPress core functions

## Fixes

- **`@since MU (3.0.0)` docblocks not matched** — the regex required digits right after `@since`, so functions using the frozen `MU (x.y.z)` form (`get_active_blog_for_user`, `wpmu_activate_signup`, `domain_exists`, …) were skipped. 
- **Root-level core files not scanned** — only `wp-includes/` and `wp-admin/` were scanned, missing global functions in `wp-signup.php` (`confirm_blog_signup`), `wp-cron.php`, and `xmlrpc.php`. 
- Other points mentioned in the related Issue is invalid.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1411-13960dacd2de326c876693ee886d2eefd302b7fd.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/tools/generate-wp-function-since-data.php
+++ b/tools/generate-wp-function-since-data.php
@@ -43,19 +43,32 @@ foreach ( $scan_dirs as $scan_dir ) {
 	);
 }
 
-// Root-level core entrypoints (e.g. wp-signup.php, wp-cron.php, xmlrpc.php).
-// Restricted to core file names and scanned non-recursively so site-specific
-// files (wp-config.php, custom root scripts, wp-content) do not pollute the dataset.
-$root_files = array_merge(
-	glob( $wordpress_dir . '/wp-*.php' ) ?: array(),
-	glob( $wordpress_dir . '/xmlrpc.php' ) ?: array()
+// Root-level core entrypoints. An explicit allowlist keeps site-specific root
+// files (wp-config.php, custom wp-*.php scripts) out of the core dataset.
+$root_entrypoints = array(
+	'index.php',
+	'wp-activate.php',
+	'wp-blog-header.php',
+	'wp-comments-post.php',
+	'wp-cron.php',
+	'wp-links-opml.php',
+	'wp-load.php',
+	'wp-login.php',
+	'wp-mail.php',
+	'wp-settings.php',
+	'wp-signup.php',
+	'wp-trackback.php',
+	'xmlrpc.php',
 );
-$root_files = array_filter(
-	$root_files,
-	static function ( string $path ): bool {
-		return 'wp-config.php' !== basename( $path );
+
+$root_files = array();
+foreach ( $root_entrypoints as $entrypoint ) {
+	$path = $wordpress_dir . '/' . $entrypoint;
+	if ( is_file( $path ) ) {
+		$root_files[] = $path;
 	}
-);
+}
+
 if ( array() !== $root_files ) {
 	$scan_targets[] = new ArrayIterator(
 		array_map(
--- a/tools/generate-wp-function-since-data.php
+++ b/tools/generate-wp-function-since-data.php
@@ -43,10 +43,20 @@ foreach ( $scan_dirs as $scan_dir ) {
 	);
 }
 
-// Root-level core files (e.g. wp-signup.php, wp-cron.php, xmlrpc.php).
-// Scanned non-recursively so wp-content plugins/themes are excluded.
-$root_files = glob( $wordpress_dir . '/*.php' );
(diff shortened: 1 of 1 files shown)


davidperezgar:
Thanks to Codex’s review, I noticed that `glob( $wordpress_dir . '/wp-*.php' )` is broader than the intended core-only scope. It will also scan site-specific root files whose names begin with wp- (for example, wp-maintenance.php), potentially adding their @since-annotated functions to the generated core dataset.

Could we use an explicit allowlist of WordPress core entrypoints instead?

davidperezgar (approved):

frantorres (approved):

Issue #350, opened 2023-12-20
Database error message when Yoast plugin is checked for Performance


ernilambar:
Steps to reproduce:

- Install `Yoast` plugin
- In the Plugin Check, enable `Performance` only
- Run the check
- Sometimes AJAX is completed but some times it halts. Nevertheless lots of notices are added in the log file.

Content of `debug.log`:

    [20-Dec-2023 10:16:40 UTC] WordPress database error Table 'wordpress.wp_pc_yoast_primary_term' doesn't exist for query SELECT * FROM `wp_pc_yoast_primary_term` WHERE `post_id` = '4' AND `taxonomy` = 'category' LIMIT 1 made by do_action('wp_ajax_plugin_check_run_checks'), WP_Hook->do_action, WP_Hook->apply_filters, WordPress\Plugin_Check\Admin\Admin_AJAX->run_checks, WordPress\Plugin_Check\Checker\Abstract_Check_Runner->run, WordPress\Plugin_Check\Checker\Preparations\Demo_Posts_Creation_Preparation->prepare, wp_insert_post, do_action('save_post'), WP_Hook->do_action, WP_Hook->apply_filters, Yoast\WP\SEO\Integrations\Watchers\Primary_Term_Watcher->save_primary_terms, Yoast\WP\SEO\Builders\Primary_Term_Builder->build, Yoast\WP\SEO\Builders\Primary_Term_Builder->save_primary_term, Yoast\WP\SEO\Repositories\Primary_Term_Repository->find_by_post_id_and_taxonomy, Yoast\WP\Lib\ORM->find_one, Yoast\WP\Lib\ORM->run, Yoast\WP\Lib\ORM::execute
[...]

PR #1453, opened 2026-08-24
Duplicate plugin-owned database tables into the runtime environment

### LINKED ISSUE #234: Plugins Fail to Create Custom Tables in Database During Runtime Check
### Describe the bug

I am encountering a persistent issue with both WooCommerce and Yoast SEO plugins. These plugins are expected to create custom tables in the WordPress database during installation or activation. However, I have noticed that sometimes, the tables are not being created, leading to "table not found" errors during runtime checks.

### Steps to Reproduce

1. Setup Plugin Checker in you local system using steps outline [here](https://github.com/10up/plugin-check#contributing)
2. Install and activate WooCommerce plugin
3. Go to Admin Panel > Tools > Plugin Check
4. Open The Browsers Network tab
5. Select "WooCommerce" and click on "Check it!" button
6. Check "enqueued_scripts_size" and "enqueued_styles_scope" check output in network tab.

Screenshot 2023-07-21 at 11 02 17 AM

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

### LINKED ISSUE #1127: Yoast Indexables Table Errors During Runtime Environment Setup
The Runtime Environment Setup routine triggers database errors when Yoast SEO is active. Plugin Check creates a temporary WordPress installation with a custom prefix ( `wp_pc_` ), but Yoast is loaded inside this environment even though its custom tables do not exist. As a result, Yoast attempts to query `wp_pc_yoast_indexable`, which is never created in the sandbox, producing errors such as:

`WordPress database error Table '...wp_pc_yoast_indexable' doesn't exist for query SELECT * FROM wp_pc_yoast_indexable WHERE object_type = 'home-page' LIMIT 1`

**Problem**
The sandbox environment created by Plugin Check does not replicate custom database tables used by active plugins. Despite this, Plugin Check loads those plugins and allows them to run initialization logic that assumes their tables exist. Any plugin that relies on custom schema (Yoast is just the most common example) will produce errors during sandbox setup.

**Expected**
The temporary environment should not load active plugins in a way that allows them to perform database-dependent operations that cannot succeed inside the sandbox. Plugin Check should ensure that plugin initialization logic does not run against incomplete or non-existent database schema.

**Actual**
Plugin Check loads all active plugins normally inside the sandbox, which triggers database queries against tables that are not present.

**Request**
Ensure that the Plugin Check sandbox environment does one of the following internally:
• Prevent plugins from executing logic that depends on custom database tables, or
• Prevent plugins from running initialization routines during the temporary installation step.

The current behavior results in noisy database errors for any plugin that uses custom tables and early-running setup hooks.

**Reproduction**
1. Activate Yoast SEO.
2. Run Plugin Check.
3. Review debug.log.
4. Observe attempts to query `wp_pc_yoast_indexable` inside the sandbox environment where those tables were never created.

guzel:
## What?

Closes #234
Closes #1127

Duplicates the schema of plugin-owned database tables into the runtime environment, so that plugins active during a runtime check can query their own tables instead of erroring out.

## Why?

`Runtime_Environment_Setup::set_up()` amends the database base prefix (e.g. `wp_` → `wp_pc_`) and installs a fresh WordPress under it. That install only creates the WordPress **core** tables.

Plugins almost always register their own tables in an activation hook, which never runs in the runtime environment. But `install_wordpress()` deliberately carries the actual site's `active_plugins` over, so those same plugins *are* loaded and *do* query their tables. The result is a stream of database errors for any plugin with custom tables:

```
WordPress database error Table 'db.wp_pc_yoast_indexable' doesn't exist for query
SELECT * FROM `wp_pc_yoast_indexable` WHERE `object_type` = 'home-page' LIMIT 1
```

Reported for Yoast SEO (#1127, #350), WooCommerce (#234) and GatherPress (comment on #234).

The earlier investigation in #234 tried calling `activate_plugin()` to trigger table creation and found it silently did nothing — `activate_plugin()` returns early when the plugin is already active. Duplicating the schema sidesteps that entirely and works no matter *how* a plugin creates its tables.

## How?

- `get_core_table_names()` derives the core table list from `wp_get_db_schema()`. The `$wpdb->tables` property is deliberately **not** used: it is public, and plugins append to it — WooCommerce registers its lookup, meta and Action Scheduler tables there. Using it would classify precisely those tables as core and skip them, leaving the most commonly reported case of this bug unfixed. As a side effect `wp_get_db_schema()` also resolves `CUSTOM_USER_TABLE` / `CUSTOM_USER_META_TABLE`, so a prefixed custom user table is correctly left alone.
- `get_custom_table_names()` lists tables matching the actual site's base prefix and subtracts that core list. On Multisite a leading site-ID segment is stripped before comparison, so `wp_3_posts` is still recognized as core. Tables already carrying the runtime prefix are skipped, and `SHOW FULL TABLES` is used so views are excluded (`CREATE TABLE ... LIKE` does not accept them).
- `create_custom_tables()` runs `CREATE TABLE IF NOT EXISTS <runtime> LIKE <source>` for each. **Structure only, never data** — runtime checks still cannot read or modify the actual site's content. `IF NOT EXISTS` keeps it idempotent across repeated runs.
- `create_custom_tables()` skips a target that already exists rather than replacing it, and reports back only the tables it genuinely created. `record_custom_tables()` stores that list in an option under the site's own prefix, merging rather than replacing so a second setup without an intermediate cleanup cannot orphan the first one's tables.
- `drop_custom_tables()` in `clean_up()` drops **only the recorded names**, never names read back from the database, and the option is deleted afterwards. A table the runtime environment did not create therefore cannot be dropped — including the case of a site that owns both `wp_foo` and a real `wp_pc_foo` (cf. #907).

One detail worth flagging for review: the duplication happens **before** `wp_install()`, not after. The install itself fires `update_option`, `user_register`, `set_user_role` etc., and active plugins already respond to those by querying their own tables — so creating the tables afterwards is too late to prevent the errors.

## Testing Instructions

1. On a site with Yoast SEO (or WooCommerce) active, run a runtime check:
   ```
   wp plugin check <plugin> --categories=performance --require=./wp-content/plugins/plugin-check/cli.php
   ```
2. Before this change, the output contains `Table 'db.wp_pc_yoast_indexable' doesn't exist` errors. After it, there are none.
3. Confirm `SHOW TABLES LIKE 'wp_pc_%'` now also lists the plugin tables during the run, and that nothing is left behind afterwards.
4. Confirm the duplicated tables are empty and the real tables are untouched.

Verified against a local site with Yoast SEO, WooCommerce and a plugin owning two custom tables:

| | Before | After |
|-|-|-|
| `wp_pc_` tables created | 12 (core only) | 59 (12 core + 47 plugin-owned) |
| `doesn't exist` errors | 14 | 0 |
| Leftover `wp_pc_` tables after cleanup | 0 | 0 |
| Rows in real `wp_posts` before / after | 153 / 153 | 153 / 153 |
| Rows in `wp_pc_options` (confirms the install still runs) | 109 | 109 |
| Rows in duplicated `wp_pc_woocommerce_tax_rates` | — | 0 (schema only) |

Verified with Yoast SEO **and** WooCommerce active, since the two exercise different paths: Yoast's tables are unknown to `$wpdb`, while WooCommerce registers its own on `$wpdb->tables`.

Adversarial check: with a real `wp_pc_zztest_thing` present alongside a real `wp_zztest_thing`, a full setup/cleanup cycle leaves the former untouched with its row intact, and a repeated setup without cleanup still leaves no tables behind.

New unit tests cover duplication, data isolation, cleanup, the `$wpdb->tables` registration case, and that cleanup never drops a table it did not create.

Full suite passes on both configurations:

```
npm run test-php            OK (523 tests, 1498 assertions)
npm run test-php-multisite   OK (524 tests, 1500 assertions)
```

`composer lint` and `composer phpstan` are clean.

## AI Usage Disclosure

- [ ] This PR was created without the help of AI tools

If AI tools were used, please describe how they were used:
Claude Code was used throughout: investigating the root cause, drafting the implementation and tests, and verifying behaviour against a local reproduction with Yoast SEO and WooCommerce active. All changes have been manually reviewed and tested by a human before submission.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fwp-admin%2Ftools.php%3Fpage%3Dplugin-check%22%2C%22phpExtensionBundles%22%3A%5B%22kitchen-sink%22%5D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fplugin-check%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1453-638348b11b6a5433de6311a54e610efc9afaa79c.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -213,8 +213,8 @@ final class Runtime_Environment_Setup {
 	 * @return string[] List of table names, without any prefix.
 	 */
 	private function get_core_table_names( string $base_prefix, string $runtime_prefix ): array {
-		// The schema functions live in this file, which callers are not required to have loaded already.
-		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		// `wp_get_db_schema()` lives in this file, which callers are not required to have loaded already.
+		require_once ABSPATH . 'wp-admin/includes/schema.php';
 
 		if ( ! preg_match_all( '/CREATE TABLE (?:IF NOT EXISTS )?`?([0-9a-zA-Z$_]+)`?/i', wp_get_db_schema( 'all' ), $matches ) ) {
 			return array();
--- a/includes/Checker/Runtime_Environment_Setup.php
+++ b/includes/Checker/Runtime_Environment_Setup.php
@@ -18,6 +18,17 @@ final class Runtime_Environment_Setup {
 	use Amend_DB_Base_Prefix;
 
 	/**
+	 * Name of the option that records which plugin-owned tables the runtime environment created.
+	 *
+	 * The record is what cleanup deletes, so that a table the runtime environment did not create can never be
+	 * dropped, even if its name happens to match the runtime environment's prefix.
+	 *
+	 * @since n.e.x.t
+	 * @var string
+	 */
+	const CUSTOM_TABLES_OPTION = 'plugin_check_runtime_custom_tables';
+
+	/**
 	 * Sets up the WordPress environment for runtime checks
 	 *
 	 * @since 1.0.0
@@ -42,8 +53,25 @@ final class Runtime_Environment_Setup {
 		// Get the existing permalink structure.
 		$permalink_structure = get_option( 'permalink_structure' );
 
+		// Get the actual site's base prefix, before it is amended below.
+		$base_prefix = $wpdb->base_prefix;
+
 		// Set the new prefix.
 		$prefix_cleanup = $this->amend_db_base_prefix();
+
+		/*
+		 * Duplicate the schema of any tables owned by other plugins.
+		 *
+		 * Installing WordPress below only creates the WordPress core tables. Plugins usually register their own tables
+		 * in an activation hook, which never runs in the runtime environment, so those tables would be missing even
+		 * though the same plugins are activated here and will query them.
+		 *
+		 * This has to happen before the install, because the install itself already triggers hooks such as
+		 * 'update_option' and 'user_register' that active plugins may respond to by querying their own tables.
+		 *
+		 * See https://github.com/WordPress/plugin-check/issues/234.
+		 */
+		$created_tables = $this->create_custom_tables( $base_prefix, $wpdb->base_prefix );
 
 		// Create and populate the test database tables if they do not exist.
 		if ( $wpdb->posts !== $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $wpdb->posts ) ) ) {
@@ -80,6 +108,9 @@ final class Runtime_Environment_Setup {
 		// Restore the old prefix.
 		$prefix_cleanup();
 
+		// Record the created tables, so that cleanup only ever deletes those. Requires the actual site's prefix.
+		$this->record_custom_tables( $created_tables );
+
 		// Return early if the plugin check object cache already exists.
 		if ( defined( 'WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION' ) && WP_PLUGIN_CHECK_OBJECT_CACHE_DROPIN_VERSION ) {
 			return;
@@ -107,6 +138,9 @@ final class Runtime_Environment_Setup {
 
 		require_once ABSPATH . '/wp-admin/includes/upgrade.php';
 
+		// Read the record of created tables while the actual site's prefix is still in place.
+		$custom_tables = (array) get_option( self::CUSTOM_TABLES_OPTION, array() );
+
 		$prefix_cleanup = $this->amend_db_base_prefix();
 		$tables         = $wpdb->tables();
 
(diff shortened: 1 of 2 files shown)


includes/Checker/Runtime_Environment_Setup.php
```diff
@@ -159,6 +198,204 @@ private function ignore_custom_tables( array $tables ): array {
...
+	 * @param string $runtime_prefix The runtime environment's database table base prefix.
+	 * @return string[] List of table names, without any prefix.
+	 */
+	private function get_core_table_names( string $base_prefix, string $runtime_prefix ): array {
+		// The schema functions live in this file, which callers are not required to have loaded already.
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+
+		if ( ! preg_match_all( '/CREATE TABLE (?:IF NOT EXISTS )?`?([0-9a-zA-Z$_]+)`?/i', wp_get_db_schema( 'all' ), $matches ) ) {
+			return array();
+		}
```
swissspidy: `wp_get_db_schema` is actually defined in `src/wp-admin/includes/schema.php`

guzel: Thank you for catching this @swissspidy 
Fixed in 9bb68b1, now requires schema.php directly.

